PR #1, opened 2021-12-01
Add Delaunay functionality.

qnzhou:
Summary:
* Added Geogram Delaunay psm in `src/utils/Delaunay`
* Added a thin wrapper to hide Geogram Delaunay from the rest of the application.
* Added unit tests.


jiangzhongshi (approved):

src/utils/Delaunay.hpp
```diff
@@ -3,22 +3,21 @@
 #include <array>
 #include <vector>
 
-namespace wmtk
-{
-	using Point3D = std::array<double, 3>;
-	using Tetrahedron = std::array<size_t, 4>;
+namespace wmtk {
+using Point3D = std::array<double, 3>;
```
jiangzhongshi, line 7: would this be potentially conflicting with others? 
Maybe we should do type alias in a common header instead of `utils`?

jiangzhongshi:
I still have the type alias comment, but since this fixed CI, I cannot wait to merge :)

PR #2, opened 2021-12-03
Basic Vertex Smoothing with simple test

jiangzhongshi:
Vertex Smoothing, and a simple test.
[x] Smoothing test for multiple tetrahedra.
[ ] Test for numerical bad quality.
[ ] Test when smoothing should be rejected. (Bad quality and optimization lead to inversion?)


tests/test_operation_smooth.cpp
```diff
@@ -0,0 +1,31 @@
+#include <wmtk/TetMesh.h>
+#include <catch2/catch.hpp>
+#include "../app/TetWild.h"
```
teseoch, line 3: this should go in tetwild, see my PR :)

jiangzhongshi: I can resolve it when yours is merged

qnzhou:
I suggest to add a few more test cases, for example:
* Smooth a mesh with more than 1 tet.
* Test a case where the smooth operation should be rejected.
* What happens if the AMIPS energy is unbounded or its Hessian is (near) singular.

tests/test_operation_smooth.cpp
jiangzhongshi: Now you should be happy :)

src/Tuple.cpp
```diff
@@ -223,8 +223,7 @@ std::vector<TetMesh::Tuple> TetMesh::Tuple::get_conn_tets(const TetMesh& m) cons
 {
     std::vector<Tuple> tets;
     for (int t_id : m.m_vertex_connectivity[m_vid].m_conn_tets) {
-        tets.emplace_back();
-        tets.back().init_from_tet(m, t_id);
+        tets.emplace_back(init_from_tet(m, t_id));
```
jiangzhongshi, line 226: This is a hotfix, should be pushed asap.

jiangzhongshi:
> What happens if the AMIPS energy is unbounded or its Hessian is (near) singular.
 I have support for these in the code, but it might take some time to construct a test (find an example). So I suggest postponing these test as Todo and move on with the current PR?

src/TetMesh.cpp
```diff
@@ -119,3 +119,38 @@ bool wmtk::TetMesh::check_mesh_connectivity_validity() const
...
+        if (vc.m_is_removed) continue;
+        assert(!vc.m_conn_tets.empty());
+        auto tid = vc[0];
+        auto local_vid = m_tet_connectivity[tid].find(i);
+
+        // note: the following conversion of local_vid-eid is dependent on the specifics of
+        // m_local_edges
+        edges.emplace_back(tuple_from_edge(tid, local_vid));
+        if (local_vid == 3) edges.back() = switch_vertex(edges.back());
+        assert(edges.back().vid() == i);
```
qnzhou, line 145: Should `Tuple::update_version_number(*this)` be called for every newly created tuple?  Otherwise, this tuple may be invalid from the birth...

qnzhou: we should also add `assert(edges.back().is_version_number_valid(*this))`

teseoch:
looks good! should merge before the tuple vnumber fix?

jiangzhongshi:
I think so.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #3, opened 2021-12-03
Document TetMesh.h

jiangzhongshi:
Partial documentation for tetmesh.h

### DIFF
--- a/src/TetMesh.h
+++ b/src/TetMesh.h
@@ -40,6 +40,14 @@ public:
         int m_timestamp = 0;
 
     public:
+        /**
+         * @brief Generate a Tuple from global tetra index and __local__ edge index (from 0-5).
+         *
+         * @param m TetMesh where the current Tuple belongs.
+         * @param tid Global tetra index
+         * @param local_eid local edge index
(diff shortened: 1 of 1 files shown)


src/TetMesh.h
```diff
@@ -40,13 +40,28 @@ class TetMesh
         int m_timestamp = 0;
 
     public:
+        /**
+         * @brief Generate a Tuple from global tetra index and __local__ edge index (from 0-5).
```
jdumas, line 44: The `@brief` is not needed. You can have doxygen use the first sentence as the brief description.

src/TetMesh.h
```diff
@@ -40,22 +40,52 @@ class TetMesh
...
-
         Tuple() {}
 
+        /**
+         * Construct a new Tuple object with global vertex/tetra index and local edge/face index
+         *
+         * @param vid vertex id
+         * @param eid edge id (local)
+         * @param fid face id (local)
+         * @param tid tetra id (local)
```
yfzhou42, line 51: tid is global

PR #4, opened 2021-12-06
2d tuples generator and switch operations with unit tests

yfzhou42:


yfzhou42:
When switch face is done on tuple of a boundary triangle I return the original tuple

src/TriMesh.h
```diff
@@ -5,333 +5,434 @@
...
+
+        /**
+         * Switch operation. See (URL-TO-DOCUMENT) for explaination.
+         *
+         * @param m
+         * @return Tuple another Tuple that share the same face, edge, but different vertex.
+         */
+        const Tuple switch_vertex(const TriMesh& m)
+        {
+            assert(is_valid(m));
```
jiangzhongshi, line 75: We should put thse into cpp

src/TriMesh.cpp
```diff
@@ -1,188 +1,136 @@
...
-	// call invariants on all entities
-	if (false) // if any invariant fails
-	{
-		// undo changes
-		return false;
-	}
-
-	return true;
-}
+// #include <wmtk/TriMesh.h>
```
jiangzhongshi, line 1: Please avoid a large chunk of code comments: If you really want to remember them, use the history in git.

tests/2d_tests.cpp
```diff
@@ -0,0 +1,311 @@
...
+{
+    TriMesh m;
+    std::vector<std::array<size_t, 3>> tris = {{{0, 1, 2}}};
+    m.create_mesh(3, tris);
+    std::vector<TriMesh::TriangleConnectivity> m_tri_connectivity = m.get_tri_connectivity();
+    std::vector<TriMesh::VertexConnectivity> m_vertex_connectivity = m.get_vertex_connectivity();
+    SECTION("test generation from vertics")
+    {
+        auto vertices_tuples = m.generate_tuples_from_vertices();
+        REQUIRE(vertices_tuples.size() == 3);
```
jiangzhongshi, line 28: You can use `REQUIRE_EQ(x, 3)`

jdumas: Catch2 doesn't have a REQUIRE_EQ afaik. Are you confusing it with Google Test?

jiangzhongshi: Yes, I am constantly confused by various testing frameworks :)

src/CMakeLists.txt
```diff
@@ -7,7 +7,10 @@ set(SOURCES
 
 	TriMesh.h
 	TriMesh.cpp
-	Tuple.cpp
+
+	EdgeCollapsingConn.cpp
```
jdumas, line 13: What does `Conn` mean? Connected? Connectivity? Co-Neural-Network? Conegative?

jiangzhongshi: `Connectivity`, This is a merge patch from main.

jiangzhongshi: We can discuss the naming rule togethehr.

jdumas: Sorry I just have a tendency to pickup on abbreviations, which I think should be avoided :D

src/TriMesh.cpp
```diff
@@ -168,8 +168,8 @@ bool TriMesh::collapse_edge(const Tuple& loc0, std::vector<Tuple>& new_edges)
     size_t test_fid2 = loc1.get_fid();
 
     assert(
-        ((std::count(n12_intersect_fids.begin(), n12_intersect_fids.end(), n1_fids) &&
-          std::count(n12_intersect_fids.begin(), n12_intersect_fids.end(), n2_fids)),
+        ((std::count(n12_intersect_fids.begin(), n12_intersect_fids.end(), n1_fids.size()) &&
+          std::count(n12_intersect_fids.begin(), n12_intersect_fids.end(), n2_fids.size())),
```
teseoch, line 172: is this correct?

jiangzhongshi: This is quite unclear, `std::count` does not require boolean

teseoch: I just fixed compilation, @yfzhou42 is this correct?

yfzhou42: it should be the fids of the two faces along the same edge. Fixed

yfzhou42: asserting it is not zero. I want to say n12_intersect_fids contains test_fid1 and test_fid2

jiangzhongshi:
Please fix this line, this does not assert anything.
```https://github.com/wildmeshing/wildmeshing-toolkit/blob/main/src/TriMesh.cpp#L146```

PR #5, opened 2021-12-09
added compilation on arm for macos on CI

teseoch:


.github/workflows/continuous.yml
```diff
@@ -51,26 +98,38 @@ jobs:
...
+            -DCMAKE_BUILD_TYPE=${{ matrix.config }}
+
+      - name: Configure arm64
+        if: matrix.arch == 'arm64'
         run: |
           mkdir -p build
           cd build
           cmake .. \
             -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+            -DCMAKE_OSX_ARCHITECTURES=arm64 \
```
qnzhou, line 125: You can probably merge this step with the previous with `-DCMAKE_OSX_ARCHITECTURES=${{ matrix.arch }}`

teseoch: is the other one correct? I searched the doc for `CMAKE_OSX_ARCHITECTURES` and it doesn't specify the values

jdumas: Not clear from the doc, but the [following arch](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-ios-tvos-or-watchos) should be supported:
`armv7;armv7s;arm64;i386;x86_64`

PR #7, opened 2021-12-10
moved tetwild files into folder, cleanup cmake, disabled soring headers in format

teseoch:
moved tetwild files into folder, cleanup cmake, disabled soring headers in format

TODO:

- [x] add new target for quadriedge collapse


tests/test_operations.cpp
```diff
@@ -2,35 +2,9 @@
...
-    std::vector<std::array<size_t, 4>> tets = {{{0, 1, 3, 2}}};
-    std::vector<TetAttributes> tet_attrs(1);
-
-    tetwild.init(vertices.size(), tets);
-    tetwild.create_mesh_attributes(vertices, tet_attrs);
-
-    tetwild.split_all_edges();
-
-	REQUIRE(tetwild.check_mesh_connectivity_validity());
+    // TODO
```
qnzhou, line 9: Can edge split be tested without tetwild?

teseoch: i think it should, I just moved files around from another PR

app/CMakeLists.txt
```diff
@@ -52,119 +39,94 @@ option(SANITIZE_THREAD    "Sanitize Thread"        OFF)
...
-#				Envelope.h
-#		)
-#
-#prepend_current_path(SOURCES)
-#wildmeshing_toolkit_copy_headers(${SOURCES})
-#wildmeshing_toolkit_set_source_group(${SOURCES})
-#target_sources(${PROJECT_NAME} PRIVATE ${SOURCES})
-
-add_library(${PROJECT_NAME} #SOURCES)
+add_library(TetWild
```
jdumas, line 64: Why do you compile it as a library if it has a `main.cpp`? Just use `add_executable()` directly.

app/CMakeLists.txt
```diff
@@ -52,119 +39,94 @@ option(SANITIZE_THREAD    "Sanitize Thread"        OFF)
...
 		external/Exception.h
 		external/MshSaver.cpp
 		external/MshSaver.h
 
-		TetWild.cpp
-		TetWild.h
+		tetwild/TetWild.cpp
+		tetwild/TetWild.h
 
 		Logger.cpp
```
jdumas, line 74: Consider making a `app/shared/` library for shared utils.

app/CMakeLists.txt
```diff
@@ -52,119 +39,94 @@ option(SANITIZE_THREAD    "Sanitize Thread"        OFF)
...
-		add_sanitizers(${PROJECT_NAME}_bin)
-	endif()
-
-	if(NOT (${CMAKE_VERSION} VERSION_LESS "3.6.0"))
-		set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT ${PROJECT_NAME}_bin)
-	endif()
-endif()
\ No newline at end of file
+include(CTest)
+enable_testing()
```
jdumas, line 131: Since CMake 3.15, `include(CTest)` will call `enable_testing()`, so only 1 line is needed (but we'd need to update our min CMake version).

Also, `include(CTest)` should be set once at the root CMakeLists.txt.

app/tests/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+
+################################################################################
+# Register tests
+################################################################################
+
+foreach(source IN ITEMS ${test_sources})
+    source_group("tests" FILES "${source}")
+endforeach()
+
+set(PARSE_CATCH_TESTS_ADD_TO_CONFIGURE_DEPENDS ON)
```
jdumas, line 41: This option is only for the alternative `ParseAndAddCatchTests()` function. This has no effect on `catch_discover_tests()`.

jdumas (approved):

app/CMakeLists.txt
teseoch: for the test, i need to link the test target against tetwild

app/CMakeLists.txt
teseoch: yeah, but also root folder is ok, if you have strong feelings for folder, I will add

app/CMakeLists.txt
jdumas: An app shouldn't have unit test. If it does, it should be compiled as a proper lib, and not have any `main.cpp`.

app/CMakeLists.txt
teseoch: done!, changed the required cmake to 3.15

app/tests/CMakeLists.txt
teseoch: done!

app/CMakeLists.txt
jdumas: `app/shared/` should be a proper CMake target, rather than just compiling the sources all over again. Having it as a subfolder prevents accidental usage of non-dependencies (e.g. `Logger.cpp` including something from `tetwild/*.h`).

jdumas: (By having a proper cmake target you avoid repeating yourself when adding common files between apps).

app/CMakeLists.txt
teseoch: tetwild is an app, it should be tested, no?

teseoch: uh, sorry you mean that the main shouldn't be there?
If so, it was a mistake sorry

app/CMakeLists.txt
```diff
@@ -58,32 +58,43 @@ endif()
...
 
-		tetwild/TetWild.cpp
-		tetwild/TetWild.h
+	shared/external/Exception.h
+	shared/external/MshSaver.cpp
+	shared/external/MshSaver.h
+)
+target_include_directories(app_shared PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/shared ${CMAKE_CURRENT_SOURCE_DIR}/shared/external)
+target_link_libraries(app_shared PUBLIC WildMeshingToolkit warnings::all)
+target_compile_features(app_shared PUBLIC ${CXX14_FEATURES})
```
jdumas, line 75: You can directly use CMake meta features now: `cxx_std_14`

tests/test_operations.cpp
jiangzhongshi: Let's add this to the TODO list and move on with this PR.

jiangzhongshi (approved):

jiangzhongshi:
OK, I merged this too soon.

Something is fishy with the mac CI https://github.com/wildmeshing/wildmeshing-toolkit/runs/4490425650?check_suite_focus=true none of the tests are run @teseoch

jdumas:
No test run usually means there is a linking error (typically a dll that cannot be loaded at runtime). Since the catch test executable fails to run, CMake cannot register the list of unit tests -- ergo no unit test is run.

jiangzhongshi:
Seems that problem is fixed as of #11

teseoch:
yeah sorry, in the process of merge the two macos build, I left a typo. this is now fixed

PR #12, opened 2021-12-14
added fenvelope to tetwild

teseoch:


teseoch:
predicates work only on x86, fenvelope needs to be fixed on arm

jiangzhongshi:
Seems that x86_64 are also failing, is it related?

teseoch:
> Seems that x86_64 are also failing, is it related?

not this is the config for x86_64

jiangzhongshi:
Would be good to have a simple test entry, to showcase the usage so we know how to integrate

jiangzhongshi (approved):

teseoch:
This is the interface:
```
FastEnvelope(const std::vector<Vector3>& m_ver, const std::vector<Vector3i>& m_faces, const Scalar eps);
		FastEnvelope(){}

		void init(const std::vector<Vector3>& m_ver, const std::vector<Vector3i>& m_faces, const Scalar eps);
		void init(const std::vector<Vector3>& m_ver, const std::vector<Vector3i>& m_faces, const std::vector<Scalar> eps);
		void init(const std::vector<std::vector<std::array<Vector3, 3>>> halfspace_input,
			const std::vector<std::array<Vector3, 2>> cornerlist_input, const Scalar eps);
		static void printnumber();

		//check if tri or point is outside
		bool is_outside(const std::array<Vector3, 3> &triangle) const;
		bool is_outside(const Vector3 &point) const;
		bool is_outside(const Vector3 &point0, const Vector3 &point1) const;
		bool is_outside_no_optimazation(const std::array<Vector3, 3> &triangle)const ;
		```

PR #15, opened 2021-12-16
Fixes for tuple timestamps.

jiangzhongshi:
We proabably need to clarify the semantics of timestamp.
- Add timestamp into tuple init, note that I am using the most current (mesh global) timestamp.
- `update_version_number` assertion should allow equal, otherwise calling twice would abort.
- `is_version_number_valid` checks for "<" instead of "!=", otherwise updating a single tetrahedra would always invalidate the tuple with respect to the opposite one.


jiangzhongshi:
Here is my understanding.
A counter is associated with either each tuple, each tetrahedron, or the whole mesh.

The `mesh-ts` is always greater than `tetra-ts`'s. If the `tuple-ts` is greater than the tetra it is adjacent to (more than one). If the `tuple-ts` is larger or equal to all adjacent `tetra-ts`, it is valid.

Claim: in the execution, such a tuple is always initialized with the most current edge-adjacent tetrahedra.

jdumas:
I'd say we shouldn't have a version tag associated to the whole mesh, this will make parallel operation unpractical. We should have a version tag associated to each tet. When an operation is being performed on a region (list of tet ids passed to each pre/post condition), their respective version tag is updated (or decreased if we are rolling back). When testing if a tuple's version tag is valid, we should only test it against the tuple's tet's version tag stored in the mesh (not iterate on all the neighbors).

teseoch:
is this still active?

jiangzhongshi:
It is active draft. Haven’t push the latest discussion

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

jiangzhongshi:
Some additional context for this PR, for future reference:
There are two different notions of timestamp
- The update-to-date tag for the queue elements (including the energy, as I was discussing before). This is something that should be application-dependent, and sometimes not needed if the key is simply edge length.
- the inherent one that belong to a tet. Think about `std::vector<>::iterator` , it is invalid when the vector is resized and reallocated, but stays valid otherwise. In our case, we are mimicking this behavior by saying that, our iterator stays valid if the mesh modifications happens “far away”, and thus the navigation capability stay “as expected” under tuple semantics.

More specifically, (for quotes in 2.), define local operations as [taking a list of tet] and [replacing with a new list]. - Regardless of how we handle memory (`push_back` or `fill_in`), each tet newly created this way should have a unique identifier (hash), and the tuples from this tet inherit this identifier.
- Therefore, if we find out a tuple invalid in this sense, we know that local tets are changed after we constructed this tuple, and the `tid/vid` stored are likely incorrect (unless by luck in memory allocation).
during local operation, the newly assigned hash should be different from everything before. However, this comes with the issue of overflow, and global counter. For this, the proposal is using pair of numbers: `<tid, local_hash++>`  as the hash. - This also satisfies our semantics, because such hash-pair is also unique than before. And implementation takes care of `tid` implicitly.
- For navigation, as long as the current tuple is valid, we can use it. For example, if we have mesh regions A and B, then construct tuple t somewhere inside A, and, afterward, B is modified. Note that the actual time-of-creation of t  proceed B; but t should still be valid, as, navigating through the mesh should have no issue at all: this means that the hash can be quite local to a tet, in practice, when we switch tet, just assign the new tet-hash to the new tuple.
- Also, this should not affect how the pass of collapse is scheduled at all, and be fairly opaque to the application queue, where a different energy-up-to-date tag will be employed at will.

PR #16, opened 2021-12-22
hide version number

teseoch:
* Merged `is_valid` with `loc.is_version_number_valid`
* Removed global version from mesh
* Removed `update_version_number`, the version is created correctly in the inits
* Updated edge split with new v number semantics @Yixin-Hu check please :)
* Updated `get_vertices` @jiangzhongshi check please :)
* `find_next_empty_slot_t` creates a new tet with vnumber -1, it gets increased after, check me

### DIFF
--- a/src/EdgeSplittingConn.cpp
+++ b/src/EdgeSplittingConn.cpp
@@ -44,10 +44,13 @@ bool wmtk::TetMesh::split_edge(const Tuple& loc0, std::vector<Tuple>& new_edges)
         int j = m_tet_connectivity[t_id].find(v1_id);
         m_tet_connectivity[new_t_id] = m_tet_connectivity[t_id];
         m_tet_connectivity[new_t_id][j] = v_id;
+        m_tet_connectivity[new_t_id].timestamp = 0;
 
         //
         m_vertex_connectivity[v_id].m_conn_tets.push_back(t_id);
         m_vertex_connectivity[v_id].m_conn_tets.push_back(new_t_id);
+        m_tet_connectivity[t_id].timestamp++; // new timestamp is +1 old one
+
         //
         for (int j = 0; j < 4; j++) {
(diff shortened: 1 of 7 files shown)


teseoch:
Do we need that? I changed the vnumber above, is not enough?

src/TetMesh.h
```diff
@@ -88,15 +88,10 @@ class TetMesh
          * Check if the current tuple is already invalid (removed during editing).
          *
          * @param m TetMesh where the tuple belongs.
-         * @return if not removed
+         * @return if not removed and thet connectivity is up to date
```
jiangzhongshi, line 91:
```suggestion
         * @return if not removed and the tuple is up to date with respect to the connectivity.
```

src/EdgeSplittingConn.cpp
```diff
@@ -129,6 +137,8 @@ bool wmtk::TetMesh::split_edge(const Tuple& loc0, std::vector<Tuple>& new_edges)
     }
 
     // new_edges
+    for (size_t t_id : n12_t_ids) m_tet_connectivity[t_id].set_version_number(m_tet_connectivity[t_id].timestamp + 1);
+    for (size_t t_id : new_t_ids) m_tet_connectivity[t_id].set_version_number(0);
```
jiangzhongshi, line 141:
```suggestion
    for (size_t t_id : new_t_ids) m_tet_connectivity[t_id].set_version_number(m_tet_connectivity[t_id].timestamp + 1);
```

teseoch: do we need this?
I added this:
https://github.com/wildmeshing/wildmeshing-toolkit/blob/e9e53c70f22436940c811e886e7f2faf8d3eb2a5/src/EdgeSplittingConn.cpp#L63

jiangzhongshi (approved):

PR #17, opened 2021-12-23
Update TriMesh.cpp

jiangzhongshi:
fix the wrong use of assertion, and clarify the `count` conditions.

### DIFF
--- a/src/TriMesh.cpp
+++ b/src/TriMesh.cpp
@@ -144,9 +144,9 @@ bool TriMesh::collapse_edge(const Tuple& loc0, Tuple& new_t)
     size_t test_fid2 = loc1.get_fid();
 
     assert(
-        ((std::count(n12_intersect_fids.begin(), n12_intersect_fids.end(), test_fid1) &&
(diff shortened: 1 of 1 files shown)


src/TriMesh.cpp
```diff
@@ -144,9 +144,9 @@ bool TriMesh::collapse_edge(const Tuple& loc0, Tuple& new_t)
     size_t test_fid2 = loc1.get_fid();
 
     assert(
-        ((std::count(n12_intersect_fids.begin(), n12_intersect_fids.end(), test_fid1) &&
-          std::count(n12_intersect_fids.begin(), n12_intersect_fids.end(), test_fid2)),
-         "faces at the edge is not correct"));
+        std::count(n12_intersect_fids.begin(), n12_intersect_fids.end(), test_fid1) > 0 &&
```
jdumas, line 147: Might as well write an auxiliary function called `contains` that takes a vector or T, a T, and does a std::find to check if the element is present.

jiangzhongshi: Agree,
@yfzhou42 could you fix this too?

PR #18, opened 2021-12-23
Fixing 3d tuple API

teseoch:
changes:
* private constructors
* private initialization
* friend class
* ts is mandatory to init tuple (to prevent bugs)
* vid and tid takes mesh as input

~~Possibly controversial change:
the `init_*` start methods are not static anymore but modify the **current** tuple~~


teseoch:
should `get_conn_tets` got in tetmesh?

teseoch:
I moved the inits to the tet mesh and also these 2:
```std::array<Tuple, 4> oriented_tet_vertices(const Tuple& t) const;
 std::vector<Tuple> get_conn_tets(const Tuple& t) const;```

teseoch:
I added `check_validity(m);` inside the constructor of tuple and now `Tuple(*this, v_id, 0, 0, t_id)` is wrong as edge 0 might not point to `v_id` in edge splitting

teseoch:
this is probably fixed in #19

PR #22, opened 2021-12-24
Compact

teseoch:
Missing:
- [x] move face attrs
- [x] move edge attrs


jiangzhongshi:
There is something about `compact` that I am not super happy about: 
After consolidation, we want to invalidate all the previously created Tuples. However, there is no way to achieve that.

One design I can think of, is enriching the hash, with an address to the tetmesh (addr, tid, hash). And when consolidation happens, always move to a new address.

teseoch:
I agree, I am also not happy about that. We could keep the tet time stamps instead of resetting them

jiangzhongshi:
That would not work: if you shrink the Tets, the additional slots are not there anymore, unless you always keep the memory, which defeats the purpose of consolidate.

jdumas:
Why do we need to invalidate previous tuples when calling 'compact'? It should only be called between operations, so we shouldn't have any tuple lying around that we need to invalidate. The semantic should be that a tuple with correct hash is only valid for the duration of its operation.

jiangzhongshi:
That also makes sense. 

Semantically speaking, it is like putting a context manager (like `with open() as file:` in python)

I was thinking, for example, in a flat program.
```
Mesh m;
Tuple e = from(m);
m.some_modify_others();
assert(e.is_valid());

m.compact();
assert(!e.is_valid());
```

jdumas:
If you think of a tuple as a quadruplet of indices, it should be ok to reuse the same tuple over two copies of the same mesh. You can already do that for vertex id and tet id. Because the structure is index-based (and not pointer-based), it should remain valid for a copy of the mesh.

As for the context, think of what happens with a std::vector::iterator: after a push_back, they are invalidated. But iterators don't have à is_valid method, not do they carry a hash of the whole vector. The only reason we need a hash for our tuple is because we want to lazily skip stale items from a queue. I think worrying about tuple validity/safety outside of an active operation is not necessary, and introduces too much design complexity. Let's just stick to what we know we need.

app/tests/test_operations.cpp
```diff
@@ -49,21 +49,38 @@ TEST_CASE("edge_collapsing", "[test_operation]")
...
+    }
 
     tetwild.split_all_edges();
     REQUIRE(tetwild.check_mesh_connectivity_validity());
 
     tetwild.collapse_all_edges();
-    CHECK(tetwild.check_mesh_connectivity_validity());
+    REQUIRE(tetwild.check_mesh_connectivity_validity());
+
+    // REQUIRE(tetwild.tet_capacity() == 5617);
```
jdumas, line 71: ?

app/tetwild/TetWild.h
```diff
@@ -75,24 +77,41 @@ class TetWild : public wmtk::TetMesh
...
     std::vector<TetAttributes> m_tet_attribute;
 
-    void resize_attributes(size_t v, size_t e, size_t t, size_t tt) override
+    void resize_attributes(size_t v, size_t t) override
     {
         m_vertex_attribute.resize(v);
-        m_edge_attribute.resize(e);
-        m_face_attribute.resize(t);
-        m_tet_attribute.resize(tt);
+        m_edge_attribute.resize(6 * t);
```
teseoch, line 90: it is not better to have 6*t outside tetwild, ie in the tetmet call?

teseoch, line 90: it is not better to have 6*t outside tetwild, ie in the tetmet call?

src/wmtk/EdgeSplittingConn.cpp
```diff
@@ -89,8 +89,6 @@ bool wmtk::TetMesh::split_edge(const Tuple& loc0, std::vector<Tuple>& new_edges)
     /// checks (possibly call the resize_attributes
     resize_attributes(
         m_vertex_connectivity.size(),
-        m_tet_connectivity.size() * 6,
-        m_tet_connectivity.size() * 4,
```
teseoch, line 93: I preferred it like that, the fact that we use *4 and *6 is an internal details

app/tests/test_operations.cpp
jiangzhongshi: This is pending PR #35

PR #25, opened 2021-12-28
Improve get_edges; an iterator example, Fixes #26

### LINKED ISSUE #26: Iterator

jiangzhongshi:


src/wmtk/TetMesh.h
```diff
@@ -37,10 +37,10 @@ class TetMesh
     // Cell Tuple Navigator
     class Tuple
     {
-        size_t m_global_vid;
-        size_t m_local_eid;
-        size_t m_local_fid;
-        size_t m_global_tid;
+        size_t m_global_vid = -1;
```
jdumas, line 40: size_t is an unsigned type... initialize with std::numeric_limits<size_t>::max() instead.

src/wmtk/TetMesh.h
```diff
@@ -140,6 +140,22 @@ class TetMesh
 
         ////testing code
         void check_validity(const TetMesh& m) const;
+        friend bool operator==(const Tuple& a, const Tuple& t)
+        {
+            return (
+                std::tie(
```
jdumas, line 146: You can use std::make_tuple instead of tie.

tests/test_tuple.cpp
```diff
@@ -102,3 +102,68 @@ TEST_CASE("count_edge_on_boundary", "[test_tuple]")
     REQUIRE(edges.size() == 10);
     REQUIRE(cnt == 9);
 }
+
+TEST_CASE("tuple_iterator", "[test_tuple]")
+{
+    class IterableMesh : public TetMesh
+    {
+    public:
+        struct EdgeIterator
```
jdumas, line 111: I find it easier to write a foreach_edges function with a custom callback function rather than have to implement your own iterator type.

jiangzhongshi: In fact, this PR is meant to initiate a discussion on how we want to proceed with the interface. That's why I am prototyping in a testing section.

PR #27, opened 2021-12-30
(To be fixed) test failure in collapse

jiangzhongshi:


Yixin-Hu:
`TEST_CASE("edge_collapsing", "[test_operation]")` is not a correct way to test collapsing. the init mesh is invalid (no coordinates set), and the mesh after collapsing is empty

Yixin-Hu:
fixed `rollback_operation`. Please remove `edge_collapsing`.

teseoch:
Remoing the test is not a fix

PR #55, opened 2022-01-07
Teseo/more 2d

teseoch:


teseoch:
@yfzhou42  there is a bug in `get_vertices` it return an invalid tuple

teseoch:
the loop is
```
    const size_t n_vertices = m_vertex_connectivity.size();
    std::vector<Tuple> all_vertices_tuples;
    all_vertices_tuples.resize(n_vertices);
    for (size_t i = 0; i < n_vertices; i++) {
```
never checks if is removed

PR #60, opened 2022-01-08
Add mshio support

qnzhou:
Summary:
* fix https://github.com/wildmeshing/wildmeshing-toolkit/issues/37
* The io support is independent from the specific tet classes and the data structures it uses.  Most of data are transferred via user provided call backs.

Usage:
```c++
#include <wmtk/utils/io.h>

wmtk::TetWild mesh; // using tetwild as an example.

// Callbacks to access tet vertices and tet indices.
auto get_tetwild_vertex = [&](size_t vid) -> Vector3f {
    return mesh.m_vertex_attribute[vid].m_posf; 
};
auto get_tetwild_tet = [&](size_t tid) -> std::array<size_t, 4> {
    const auto t_tuple = mesh.tuple_from_tet(tid);
    const auto v_tuples = mesh.oriented_tet_vertices(t_tuple);
    return {{v_tuples[0].vid(mesh), v_tuples[1].vid(mesh), v_tuple[2].vid(mesh), v_tuples[3].vid(mesh)}};
};

wmtk::MshData msh;

// To visualize tet and tet attributes.
msh.add_tet_vertices(num_vertices, get_tetwild_vertex);
msh.add_tets(num_tets, get_tetwild_tet);
msh.save("output.msh");
```

### DIFF
--- a/cmake/recipes/mshio.cmake
+++ b/cmake/recipes/mshio.cmake
@@ -1,0 +1,18 @@
+include_guard()
+
+if (NOT mshio::mshio)
+    FetchContent_Declare(
+        MshIO
+        GIT_REPOSITORY https://github.com/qnzhou/MshIO.git
+        GIT_TAG        main
+        GIT_SHALLOW TRUE
+    )
+
+    FetchContent_GetProperties(MshIO)
+    if (NOT mshio_POPULATED)
+        FetchContent_Populate(MshIO)
+        option(MSHIO_EXT_NANOSPLINE "Enable nanospline extension" On)
+        add_subdirectory(${mshio_SOURCE_DIR} ${mshio_BINARY_DIR})
+    endif()
+
+endif()
--- a/cmake/recipes/mshio.cmake
+++ b/cmake/recipes/mshio.cmake
@@ -1,18 +1,15 @@
-include_guard()
-
-if (NOT mshio::mshio)
-    FetchContent_Declare(
-        MshIO
-        GIT_REPOSITORY https://github.com/qnzhou/MshIO.git
-        GIT_TAG        main
-        GIT_SHALLOW TRUE
-    )
-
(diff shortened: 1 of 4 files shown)


cmake/recipes/mshio.cmake
```diff
@@ -0,0 +1,18 @@
+include_guard()
+
+if (NOT mshio::mshio)
```
jdumas, line 3: Why not use the usual pattern?
```cmake
if(TARGET mshio::mshio)
    return()
endif()
```

cmake/recipes/mshio.cmake
```diff
@@ -0,0 +1,18 @@
...
+    FetchContent_Declare(
+        MshIO
+        GIT_REPOSITORY https://github.com/qnzhou/MshIO.git
+        GIT_TAG        main
+        GIT_SHALLOW TRUE
+    )
+
+    FetchContent_GetProperties(MshIO)
```
jdumas, line 11: Can be simplified to
```cmake
FetchContent_MakeAvailable(mshio)
```

cmake/recipes/mshio.cmake
```diff
@@ -1,18 +1,15 @@
...
-    FetchContent_GetProperties(MshIO)
-    if (NOT mshio_POPULATED)
-        FetchContent_Populate(MshIO)
-        option(MSHIO_EXT_NANOSPLINE "Enable nanospline extension" On)
-        add_subdirectory(${mshio_SOURCE_DIR} ${mshio_BINARY_DIR})
-    endif()
+FetchContent_Declare(
+    mshio
+    GIT_REPOSITORY https://github.com/qnzhou/MshIO.git
+    GIT_TAG        main
```
jdumas, line 10: use a specific sha1/tag rather than main?

jdumas (approved):

PR #94, opened 2022-01-14
fix shallow clone

jiangzhongshi:


cmake/recipes/libigl.cmake
```diff
@@ -33,9 +33,8 @@ option(LIBIGL_WITH_XML               "Use XML"                      OFF)
 include(FetchContent)
 FetchContent_Declare(
     libigl
-    GIT_REPOSITORY https://github.com/libigl/libigl.git
-    GIT_TAG v2.3.0
-    GIT_SHALLOW TRUE
+    URL https://github.com/libigl/libigl/archive/refs/tags/v2.3.0.zip
+    URL_HASH MD5=6f17257bf1c898872091e5095d477cbf
```
jdumas, line 37: why the switch to archives instead of git tags?

jiangzhongshi: Originally, this is a fix PR because git shallow was specified everywhere, and it caused a problem (for mshio), when there isn't a tag.

While changing that, I just went through the versions of each dep and changed to zips. I don't really have a preference.

jdumas: Zips are not really faster in my experience, just more annoying to update (since you need to provide the MD5 hash).

PR #113, opened 2022-01-17
changed critical logs to trace

teseoch:


app/EdgeOperations2d/AdaptiveRemeshing.cpp
```diff
@@ -110,18 +110,18 @@ bool EdgeOperations2d::collapse_remeshing(double L)
 
         TriMesh::Tuple new_vert;
         // c++;
-        // spdlog::critical("collapse attempt {}", c);
+        // spdlog::trace("collapse attempt {}", c);
         if (!TriMesh::collapse_edge(loc, new_vert)) continue;
         s++;
-        spdlog::critical("collapse success {}", s);
+        spdlog::trace("collapse success {}", s);
```
jdumas, line 116: These should be `wmtk::logger()`, not `spdlog::...`

teseoch: true, @yfzhou42 can you fix it?

PR #118, opened 2022-01-18
[bug reproduce] MshIO with not found file trigger assertion in debug, or hang indefinitely in release.

jiangzhongshi:


jiangzhongshi:
Thanks!

PR #138, opened 2022-01-24
Use Lagrange to resolve nonmanifoldness

jiangzhongshi:


jdumas:
You can't just "borrow" the function and release it under MIT license, it has to be Apache 2. For that you need to at least keep the license header, and put a proper [NOTICE.txt](https://github.com/adobe/lagrange/blob/aeebd56eaaf9d3ca448174501781c8bedaa05b63/NOTICE.txt#L1) in the repo root. Honestly I'd suggest just pulling Lagrange as a dependency and calling the function, as it will also do the remaining cleanup steps:
```c++
    out_mesh = resolve_vertex_nonmanifoldness(*out_mesh);
    out_mesh = remove_topologically_degenerate_triangles(*out_mesh);
    out_mesh = remove_duplicate_facets(*out_mesh);
    out_mesh = remove_isolated_vertices(*out_mesh);
```

Do you need to keep track of which vertex was modified exactly? Or only the freeze all boundary vertices after the resolve? One easy way to check the former might be to build a hash table position -> vertex index. If there are more than 1 vertex per bucket, then we have de-duplicated a vertex there and it needs to be frozen.

danielepanozzo:
Both solutions are fine now, let's go for whatever is faster to integrate @jdumas

danielepanozzo:
We need to keep track of all the edges and vertices duplicated, as these will have to be preserved in the coarsening.

PR #140, opened 2022-01-25
Parallel triangle insertion

jiangzhongshi:


jdumas:
Let's try to avoid checking huge meshes in the source code please.

PR #154, opened 2022-01-27
optimized triangle insertion

JcDai:


jiangzhongshi:
There is a random assertion failure (in `setup_attributes`) [here](https://github.com/wildmeshing/wildmeshing-toolkit/runs/5196154430?check_suite_focus=true). 
Cannot reproduce now, note to be careful when running larger scale experiments.

PR #167, opened 2022-02-09
qslim

yfzhou42:


app/EdgeOperations2d/ShortestEdgeCollapse.cpp
```diff
@@ -54,6 +54,7 @@ std::vector<TriMesh::Tuple> Edge2d::EdgeOperations2d::new_edges_after(
 
 bool Edge2d::EdgeOperations2d::collapse_shortest(int target_operation_count)
 {
+    size_t initial_size = get_vertices().size();
```
jiangzhongshi, line 57: Why is this here? This result is not used in your code.

app/QSLIM/app/main.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+int main(int argc, char** argv)
+{
+    // input
+    // target_verts
+    // output
+    // ep
+
+    const std::string root(WMT_DATA_DIR);
+    const std::string path = argv[1];
```
jiangzhongshi, line 49: prefer use cli11 instead of these confusing argv[x].

app/EdgeOperations2d/ShortestEdgeCollapse.cpp
yfzhou42: This is the old code. The whole /EdgeOPerations2d folder can be deleted soon

src/wmtk/ExecutionScheduler.hpp
```diff
@@ -271,6 +272,10 @@ struct ExecutePass
 
                 if (stop.load(std::memory_order_acquire)) return;
                 if (cnt_update > stopping_criterion_checking_frequency) {
+                    wmtk::logger().info(
+                        "I am in stopping criteria cnt : {} freqency is :{}",
+                        cnt_success,
+                        stopping_criterion_checking_frequency);
```
jiangzhongshi, line 278:
```suggestion
```

PR #199, opened 2022-05-06
Added missing include

zfergus:
I was getting an error when compiling:
```
error: ‘size_t’ was not declared in this scope; did you mean ‘std::size_t’?
```
when using `g++ (GCC) 11.2.0`.

Fixed this by including `cstddef`.

### DIFF
--- a/src/wmtk/utils/Delaunay.hpp
+++ b/src/wmtk/utils/Delaunay.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cstddef>
 #include <array>
 #include <vector>
 


jiangzhongshi:
Huh, this is interesting, so an outside PR wouldn't trigger auto-ci, thus cannot be merged....

jdumas:
> Huh, this is interesting, so an outside PR wouldn't trigger auto-ci, thus cannot be merged....

That's because current setup only builds on `push` events, not on `pull_requests` event. See #199 for a fix.

PR #258, opened 2023-02-16
Operators and Recording on TriMesh

mtao:
Notable changes:
* Each Operation (and its before/after checks) are derived from TriMeshOperator and their full execution is handled by the class
* TriMesh topology (vertices and triangles) are now Attributes, which moves the burden of undoing operations from the operator to the Attribute class

Initial implementation was thanks to @zlyfunction and adapted by @mtao. To pursue the history in more detail, changes come from mtao/wildmeshing-toolkit:mtao/TriMesh_event_logging (around 3e3d7c0), which comes from zlyfunction/wildmeshing-toolkit:main .


mtao:
BTW my understanding is that @yunfanzhou @JcDai might be interested in looking over the changes. I'm currently taking a pass of deprecating redundant functionality and changing existing applications in main over to this new format

src/wmtk/utils/OperationLogger.h
```diff
@@ -0,0 +1,102 @@
+#pragma once
+
+#include <oneapi/tbb/concurrent_vector.h>
+#include <oneapi/tbb/mutex.h>
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wcast-qual"
```
jdumas, line 7: You might want to use the `EnableWarnings.hpp` and `DisableWarnings.hpp` for this instead.

cmake/recipes/libigl.cmake
```diff
@@ -20,8 +20,11 @@ message(STATUS "Third-party: creating target 'igl::core'")
 include(FetchContent)
 FetchContent_Declare(
     libigl
-    GIT_REPOSITORY https://github.com/libigl/libigl.git
-    GIT_TAG 3ea7f9480967fcf6bf02ce9b993c0ea6d2fc45f6
+    #GIT_REPOSITORY https://github.com/libigl/libigl.git
+    #GIT_TAG 3ea7f9480967fcf6bf02ce9b993c0ea6d2fc45f6
+    GIT_REPOSITORY https://github.com/mtao/libigl.git
```
mtao, line 25: This change needs to be undone before the final submission - this branch is a working set of updates to libigl to suppress some unnecessary warnings on gcc 11.3

PR #264, opened 2023-04-19
yunfan/paired_operations

yunfanzhou:


src/wmtk/utils/json_sink.cpp
```diff
@@ -44,9 +45,18 @@ std::shared_ptr<spdlog::logger> make_json_file_logger(
...
+    std::vector<std::shared_ptr<spdlog::sinks::sink>> sinks;
+    sinks.emplace_back(file_sink);
+    if(also_output_stdout) {
+        sinks.emplace_back(console_sink);
+    }
     // dunno why make_shared didn't work for me today
     std::shared_ptr<spdlog::logger> js_logger;
-    js_logger.reset(new spdlog::logger(name, {console_sink, file_sink}));
+
+    spdlog::init_thread_pool(8192,1);
```
jdumas, line 56: That's the default spdlog thread pool. I don't think we need to call it explicitly?

PR #273, opened 2023-05-12
Joined collapse pair

mtao:
implementing collapse that works on pairs
this involves constructing a link condition that can understand pairs, which resulted in a refactor


app/adaptive_tessellation/Collapse.h
```diff
@@ -12,23 +12,37 @@
...
+        size_t v1; // first vertex index of the edge being collapsed
+        size_t v2; // second vertex index of hte edge being collapsed
+
         double length3d;
+
+        size_t v_top; // remaining vertex of the input triangle being collapsed
+        size_t v_bot; // remaining vertex of the other triangle being collapsed
+
+        // pair of vids that are an existing edge -> mirror edge, curveid
+        std::map<std::array<size_t, 2>, SeamData> mirrored_edges;
```
jdumas, line 40: You may want to use an `std::unordered_map<>` rather than a `std::map` here. You can take a look at the `OrderedPairHash` object in `lagrange/utils/hash.h` for a hash operator that works on pairs.

PR #282, opened 2023-05-17
Collecting logger data

daniel-zint:
Adding the class `LoggerDataCollector` which collects all the data of a mesh that should be logged.

### DIFF
--- a/app/adaptive_tessellation/LoggerDataCollector.h
+++ b/app/adaptive_tessellation/LoggerDataCollector.h
@@ -17,7 +17,7 @@ class LoggerDataCollector
         double std_dev_ = -std::numeric_limits<double>::max();
 
     public:
-        void compute_statistics(std::vector<double>& data)
+        StatisticsObj(std::vector<double> data)
         {
             std::sort(data.begin(), data.end());
 
@@ -36,20 +36,25 @@ class LoggerDataCollector
             std_dev_ = std::sqrt(sq_sum / data.size());
         }
 
-        double min() const { return min_; }
-        double max() const { return max_; }
-        double mean() const { return mean_; }
-        double median() const { return median_; }
-        double std_dev() const { return std_dev_; }
+        nlohmann::json to_json() const
+        {
+            // const nlohmann::json& js,
+            return {
+                {"min", min_},
+                {"max", max_},
+                {"mean", mean_},
+                {"median", median_},
+                {"std_dev", std_dev_}};
+        }
     };
 
     size_t peak_memory_ = -1; // peak memory in bytes
     size_t num_faces_ = -1;
     size_t num_vertices_ = -1;
-    StatisticsObj edge_length_;
-    StatisticsObj energy_;
-    StatisticsObj area_;
-    // StatisticsObj min_angle_;
+    std::vector<double> triangle_energies_;
+    std::vector<double> triangle_areas_;
+    // std::vector<double> triangle_min_angles_;
+    std::vector<double> edge_lengths_;
 
     mutable igl::Timer timer_;
     bool timer_started_ = false;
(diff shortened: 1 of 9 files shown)


app/adaptive_tessellation/LoggerDataCollector.h
```diff
@@ -0,0 +1,189 @@
...
+#include "wmtk/utils/getRSS.h"
+
+namespace adaptive_tessellation {
+
+class LoggerDataCollector
+{
+    class StatisticsObj
+    {
+        double min_ = std::numeric_limits<double>::max();
+        double max_ = -std::numeric_limits<double>::max();
```
mtao, line 14: use lowest rather than -max, much cleaner

app/adaptive_tessellation/LoggerDataCollector.h
```diff
@@ -0,0 +1,189 @@
...
+
+#include <spdlog/stopwatch.h>
+#include "AdaptiveTessellation.h"
+#include "wmtk/utils/getRSS.h"
+
+namespace adaptive_tessellation {
+
+class LoggerDataCollector
+{
+    class StatisticsObj
```
mtao, line 11: we should deprecate defining classes in classes, but probably not worth doing for now (this statistics object seems to be an internal detail anyway so it doesn't need to be defined either)

app/adaptive_tessellation/LoggerDataCollector.h
```diff
@@ -0,0 +1,189 @@
...
+
+            std::vector<double> diff(data.size());
+            std::transform(data.begin(), data.end(), diff.begin(), [this](double x) {
+                return x - mean_;
+            });
+            const double sq_sum = std::inner_product(diff.begin(), diff.end(), diff.begin(), 0.0);
+            std_dev_ = std::sqrt(sq_sum / data.size());
+        }
+
+        nlohmann::json to_json() const
```
mtao, line 39: put in cpp

app/adaptive_tessellation/LoggerDataCollector.h
```diff
@@ -0,0 +1,189 @@
...
+            // const nlohmann::json& js,
+            return {
+                {"min", min_},
+                {"max", max_},
+                {"mean", mean_},
+                {"median", median_},
+                {"std_dev", std_dev_}};
+        }
+
+        operator nlohmann::json() const { return to_json(); }
```
mtao, line 50: put in cpp

app/adaptive_tessellation/LoggerDataCollector.h
```diff
@@ -0,0 +1,189 @@
...
+#include "AdaptiveTessellation.h"
+#include "wmtk/utils/getRSS.h"
+
+namespace adaptive_tessellation {
+
+class LoggerDataCollector
+{
+    class StatisticsObj
+    {
+        double min_ = std::numeric_limits<double>::max();
```
mtao, line 13: not a big deal as we will have to refactor everything in the future after a discussion on style but i far prefer having prefix annotations for member variables than suffix annotations - for non-trivial types 
matrix_.size() is harder to realize as a member than _matrix.size() (or my favorite is m_matrix.size(), which is distinguished from statics called s_matrix

app/adaptive_tessellation/LoggerDataCollector.h
```diff
@@ -0,0 +1,189 @@
...
+
+public:
+    /**
+     * @brief Collect logging data from the mesh.
+     *
+     * Computes the following things: peak memory, number of faces, number of vertices, edge length
+     * statistics, energy statistic, area statistic. All statistics contain min, max, mean, median,
+     * and standard deviation.
+     */
+    void evaluate_mesh(const AdaptiveTessellation& mesh)
```
mtao, line 73: put in cpp

app/adaptive_tessellation/LoggerDataCollector.h
```diff
@@ -0,0 +1,189 @@
...
+                triangle_energies_.emplace_back(triangle_energy);
+            }
+        }
+        // edge stuff
+        {
+            const auto edges = mesh.get_edges();
+
+            edge_lengths_.reserve(edges.size());
+
+            for (const Tuple e : edges) {
```
mtao, line 112: ref

app/adaptive_tessellation/LoggerDataCollector.h
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    void log_json_verbose(const AdaptiveTessellation& mesh, const std::string& log_name)
+    {
+        log_json(mesh, log_name, true);
+    }
+
+private:
+    nlohmann::json general_info_to_json() const
+    {
```
mtao, line 165: all to cpp

app/adaptive_tessellation/Parameters.h
```diff
@@ -105,6 +105,7 @@ struct Parameters
         const nlohmann::json& js,
         bool flush = false) // flush should force file output immediately, but will be slow for
                             // per-operation things
+        const
```
mtao, line 108: wasn't your code but i wish this was in a cpp :(

src/wmtk/utils/getRSS.cpp
```diff
@@ -31,13 +33,15 @@
...
+#include "getRSS.h"
+
 
 /**
  * Returns the peak (maximum so far) resident set size (physical
  * memory use) measured in bytes, or zero if the value cannot be
  * determined on this OS.
  */
-size_t getPeakRSS()
+size_t wmtk::getPeakRSS()
```
mtao, line 44: one sorta ugly but sorta nice thing is that in cpp you can just say `using namespace wmtk` and it'll auto-figure out that you're defining the declaration `wmtk::getPeakRSS`

mtao (approved):
looks good, minor quibbles that don't really matter

i did make a few comments about putting json stuff in  cpp files - nlohmann::json is actually a pretty heavy object (it's a header library that has some fairly intricate template specification going on which affects compilation times)

src/wmtk/utils/getRSS.cpp
jdumas: For this I prefer to wrap everything in a `namespace wmtk { .... }` surrounding the entire .cpp ....

app/adaptive_tessellation/LoggerDataCollector.h
daniel-zint: I'll keep them the way they are until we discussed naming conventions.

app/adaptive_tessellation/Parameters.h
daniel-zint: I also moved that to a cpp file. It might make sense to move other stuff to .cpp too but I'll keep it like that for now.

src/wmtk/utils/getRSS.cpp
daniel-zint: I used the `namespace wmtk {...}` version. Just writing `using namespace wmtk` seems like it does not work for MSVC.

PR #284, opened 2023-05-19
yunfan/quadrics_for_operations

yunfanzhou:


app/adaptive_tessellation/AdaptiveTessellation.cpp
```diff
@@ -104,6 +105,14 @@ void AdaptiveTessellation::mesh_preprocessing(
         position_image_path,
         normal_image_path,
         height_image_path);
+
+    m_quadric_integral =
+        wmtk::QuadricIntegral(displaced, wmtk::QuadricIntegral::QuadricType::Plane);
```
jdumas, line 110: Make sure to default to `Triangle` quadric, and also add an option to switch to `Point` quadrics.

app/adaptive_tessellation/AdaptiveTessellation.cpp
```diff
@@ -869,77 +907,97 @@ double AdaptiveTessellation::get_area_accuracy_error_per_face_triangle_matrix(
...
+{
+    double ret = 0.0;
+    wmtk::Quadric<double> q;
+    for (const Tuple& tri : get_one_ring_tris_for_vertex(v)) {
+        q += get_face_attrs(tri).accuracy_measure.quadric;
+    }
+    auto v_pos = vertex_attrs[v.vid(*this)].pos;
+    Eigen::Matrix<double, 3, 1> v_world_pos =
+        mesh_parameters.m_displacement->get(v_pos(0), v_pos(1));
+    ret = v_world_pos.transpose() * q.A() * v_world_pos - 2.0 * v_world_pos.dot(q.b()) + q.c;
```
jdumas, line 996: You can simply calculate `ret = q(v_world_pos);` instead (for smoothing I do explicitly because of autodiff).

app/adaptive_tessellation/Split.cpp
```diff
@@ -522,34 +571,40 @@ void AdaptiveTessellation::split_all_edges()
...
+            double unscaled_total_error = 0.;
+
             if (m.mesh_parameters.m_edge_length_type == EDGE_LEN_TYPE::AREA_ACCURACY) {
-                std::tie(total_error, error1, error2) = m.get_area_accuracy_error_for_split(tup);
-            } else
+                unscaled_total_error = m.get_cached_area_accuracy_error_for_split(tup);
+                total_error = unscaled_total_error * m.get_length2d(tup);
+            } else if (m.mesh_parameters.m_edge_length_type == EDGE_LEN_TYPE::TRI_QUADRICS) {
+                unscaled_total_error = m.get_quadrics_area_accuracy_error_for_split(tup);
+                total_error = unscaled_total_error * m.get_length2d(tup);
```
jdumas, line 596: I don't think we need to weight by edge length now, since the quadrics error are per-vertex and not per-face anymore.

app/adaptive_tessellation/Split.cpp
```diff
@@ -15,6 +15,35 @@ auto split_renew = [](auto& m, auto op, auto& tris) {
     return optup;
 };
 
+double AdaptiveTessellation::get_quadrics_area_accuracy_error_for_split(
```
jdumas, line 18: Where are we computing the error _after_ the split here? Do we even care? Because if it's only for the "priority", I would simply use the sum of the quadrics for the two vertices involved in the split, rather than all 2 vertices around an edge.

app/adaptive_tessellation/Split.cpp
yunfanzhou: True

app/adaptive_tessellation/AdaptiveTessellation.cpp
yunfanzhou: changing

app/adaptive_tessellation/Split.cpp
jdumas: I don't think this is needed anymore. Previously we added this weighing to favor splitting the longest edge of a given face (since all three edge would receive the same error). But the quadrics already account for that by being evaluated at the vertex position.

app/adaptive_tessellation/Split.cpp
yunfanzhou: The update happens here ([(https://github.com/wildmeshing/wildmeshing-toolkit/blob/b078116ca65f239e06a640aa3b5d760e7512e6ad/app/adaptive_tessellation/Split.cpp#L207-L224)])

PR #294, opened 2023-05-26
Generic Attribute Serialization

mtao:
A generic mechanism for serializing attributes. The default interface for now will be a member function / trait for any attribute type to a STL type and then have an automatic STL -> HDF serialization scheme using HDF5 composite types (vector of structs). 

Experimenting with how to enable good defaults yet maintaining customizability (this is is explicitly to contrast with HighFive, which effectively requires isolation along compilation units to allow customizability. I'll still end up using HighFive at the lowest level though).

Struct of vectors would be nicer, especially for introducing inter-op with Paraview. I'm actively thinking about if there's a nice way to do this but thus far my best ideas lead to callback/macro hell with lots of performance limitations and challenges for validating the completeness of serializations. I suspect I'll just write a VoS <-> SoV converter as a utility tool/lib to simplify export to Paraview.


mtao:
this was originally for wmtk v1. we're not on that anymore

PR #304, opened 2023-06-13
yunfan/new_amips_double

yunfanzhou:


mtao:
what happened to this updated implementation? is this going to be automatically incorporated into what you're doing in [PR350](https://github.com/wildmeshing/wildmeshing-toolkit/pull/350)

PR #305, opened 2023-06-14
Fix scaling issue in displaced image.

jdumas:
I think this change was lost during the post-deadline merge, but without it the results from `TextureIntegral` will have the wrong scale.


mtao (approved):
lgtm :+1:  , i swear we merged this into yunfan's branch at some point before we merged her branch back into main :upside_down_face:

PR #331, opened 2023-07-06
Remove PrimitiveType::Invalid

mtao:
This "invalid" primitivetype is becoming a source of compilation warnings and from what i can tell it's only used as a default initialized value for classes that should not be default initialized and to create asserts.  Is it ok to remove it? I think it came from the SimplicialComplex.hpp effort?


daniel-zint:
As long as these classes then cannot be default initialized that is perfectly fine for me.

daniel-zint (approved):
lgtm

PR #332, opened 2023-07-11
Hackathon/dev - add tests for components

daniel-zint:
I added some tests for the wmtk components. I am not sure if it is desired that those tests should be in an independent executable.


mtao (changes requested):
i have a few questions about why certain things are the way they are, but generally looks good. lmk about these few points / resolve them and we should be good?

CMakeLists.txt
```diff
@@ -77,6 +77,9 @@ add_library(wildmeshing_components)
 add_library(wmtk::components ALIAS wildmeshing_components)
 add_subdirectory(components/wmtk_components)
 
+# Group source files for IDEs
+file(GLOB_RECURSE COMPONENTS_FILES "components/*.cpp" "components/*.h" "components/*.hpp")
```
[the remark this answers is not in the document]
daniel-zint: That is purely for showing the folder structure in the IDE and is not used in the compilation. I thought we could make an exception here.
I could change the name from `COMPONENTS_FILES` to `COMPONENTS_FILES_FOR_SOURCE_GROUP` to make it more obvious.

src/wmtk/io/MeshReader.hpp
```diff
@@ -17,7 +17,7 @@ class MeshReader
     void read(Mesh& mesh);
 
 private:
-    const std::filesystem::path& m_filename;
+    std::filesystem::path m_filename;
```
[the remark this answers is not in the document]
daniel-zint: It does not need to be mutable. The important part is that it must not be a reference, otherwise the following code breaks at runtime:
```
MeshReader reader("test.hdf5");
reader.read(mesh1);
```
I just removed `const` because I once learned that members should never be `const`. I can add it back if you want.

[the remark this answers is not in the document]
mtao: :shrug: i agree that const members can be a huge annoyance / why people would want do disallow them, but if the semantic is that a MeshReader is defined on a single file (and one MeshReader should only ever read that one file) then i'd rather this be const.

components/wmtk_components/input/CMakeLists.txt
mtao, line 1: oh, why aren't these hpp?

components/wmtk_components/output/CMakeLists.txt
mtao, line 1: hpp?

src/wmtk/TriMeshOperation.cpp
```diff
@@ -13,22 +13,20 @@ TriMesh::TriMeshOperationState::PerFaceData TriMesh::TriMeshOperationState::get_
...
 
     return face_data;
 }
 
 // constructor
 TriMesh::TriMeshOperationState::TriMeshOperationState(TriMesh& m)
-    : flag_accessors({{m.get_flag_accessor(PrimitiveType::Vertex),
-                       m.get_flag_accessor(PrimitiveType::Edge),
-                       m.get_flag_accessor(PrimitiveType::Face)}})
+    : flag_accessors{m.get_flag_accessor(PrimitiveType::Vertex), m.get_flag_accessor(PrimitiveType::Edge), m.get_flag_accessor(PrimitiveType::Face)}
```
mtao, line 29: @daniel-zint you were asking if perhaps your formatter is doing something different from eveyrone else's - this is a case where my formatter definitely preferred the previous code. lets double check the clang-format and whatnot soon

src/wmtk/attribute/MeshAttributes.cpp
```diff
@@ -100,6 +100,12 @@ void MeshAttributes<T>::set(const AttributeHandle& handle, std::vector<T> val)
     attr.set(std::move(val));
 }
 
+template <typename T>
```
mtao, line 103: this size should be always equal to m_size, perhaps this naming is better but we should discuss some standard for these names

src/wmtk/io/MeshReader.cpp
```diff
@@ -71,7 +71,9 @@ void MeshReader::set_attribute(
 
     accessor.set_attribute(v);
 
-    mesh.set_capacities_from_flags();
+    if (name == "flags") {
```
mtao, line 74: perhaps we need a `set_capacitity_from_flag(PrimitiveType)` here. This current implementation will reload ALL capacities every time

mtao (commented):
lgtm up to some minor cleanup things :)

src/wmtk/TriMeshOperation.cpp
daniel-zint: We should double check that, yes, but in this specific case, I needed to change the parentheses/braces. Otherwise my compiler complained that a deleted function was called. I changed `({{` to `{`.

src/wmtk/io/MeshReader.cpp
daniel-zint: Yes but the overhead is really minor. Eventually, we won't have more than four capacities.

mtao (approved):
lgtm

PR #337, opened 2023-07-31
Checkpoints

mtao:
Checkpoints allow for us to look at historical views of a mesh during operation execution.
This is possible because we only flush our updates to a mesh at the end of an operation and keep track of updates through a stack of diffs over the original mesh - and each "checkpoint" is therefore a reference to a particular point in the stack.


mtao:
this is now a generalization of https://github.com/wildmeshing/wildmeshing-toolkit/pull/579 . but it's not clear how pertinent this is anymore

PR #339, opened 2023-08-02
Refactoring tests and operations

daniel-zint:
Changing names and visibility of stuff + adding comments. So far, no changes in the actual behavior.


src/wmtk/TriMeshOperationExecutor.cpp
```diff
@@ -128,94 +121,54 @@ void TriMesh::TriMeshOperationExecutor::glue_ear_to_face(
...
             return; // TODO: throw exception, non-manifold
         }
 
         // change VF for V_new,V_C
         long& vf_new = vf_accessor.scalar_attribute(new_vid);
-        long& vf_c = vf_accessor.scalar_attribute(FaceDatas[face_index].V_C_id);
-        const long f_ear_l = FaceDatas[face_index].ears[0].fid;
-        const long f_ear_r = FaceDatas[face_index].ears[1].fid;
+        long& vf_c = vf_accessor.scalar_attribute(m_incident_face_datas[face_index].opposite_vid);
+        const long f_ear_l = m_incident_face_datas[face_index].ears[0].fid;
```
mtao, line 143: i'd rather we did `auto& incident_face_data = m_incident_face_datas[face_index];` to reduce array indexing. it's a habit from teh old codebase we shoud really avoid

src/wmtk/TriMesh.cpp
```diff
@@ -18,6 +19,60 @@ TriMesh::TriMesh(TriMesh&& o) = default;
...
+    // get faces in open_star(B)
+    auto star_A_f = SimplicialComplex::open_star(PrimitiveType::Vertex, t, *this)
+                        .get_simplices(PrimitiveType::Face);
+    auto star_B_f = SimplicialComplex::open_star(
+                        PrimitiveType::Vertex,
+                        switch_tuple(t, PrimitiveType::Vertex),
+                        *this)
+                        .get_simplices(PrimitiveType::Face);
+    std::vector<long> faces_to_change_fv;
+    for (const Simplex& simplex : star_B_f) {
```
mtao, line 48: this code is older than the PR, but it breaks the semantics of a Simplex pretty harshly (isn't the point of a Simplex to lock down the interpretation of a Tuple to a single type? this is dynamically casting all simplices to an arbitrary face in its co-boundary.

this sort of logic makes me sorta wish SimplicialComplex held a vector of vector of Tuple rather than a single vector of Simplex objects :shrug:

src/wmtk/TriMeshOperationExecutor.cpp
```diff
@@ -1,41 +1,33 @@
...
-    Tuple t2_edge = m_mesh.switch_tuple(t, PV);
-    t2_edge = m_mesh.switch_tuple(t2_edge, PE);
-
-    face_data.ears[0] = EarGlobalIDs{
+    IncidentFaceData face_data;
+    face_data.fid = m_mesh.id(t, PF);
+    const Tuple t1_edge = m_mesh.switch_tuple(t, PE);
+
+    face_data.opposite_vid = m_mesh.id(m_mesh.switch_tuple(t1_edge, PV), PV);
+    const Tuple t2_edge = m_mesh.switch_tuple(m_mesh.switch_tuple(t, PV), PE);
```
mtao, line 18: it'd be good to have a comment for what's going on this older code - i'm not sure what t1 / t2 are / why switching PV vs PE is right

tests/test_2d_operation_construction.cpp
```diff
@@ -64,48 +68,83 @@ DEBUG_TriMesh test_collapse(const DEBUG_TriMesh& mesh, const Tuple& e, bool shou
...
         m = single_triangle();
-        auto e = m.tuple_from_id(PrimitiveType::Edge, 0);
+        auto e = m.tuple_from_id(PE, 0);
 
-        REQUIRE(m.id(e, PrimitiveType::Edge) == 0);
-        REQUIRE(m.id(e, PrimitiveType::Vertex) == 0);
-        REQUIRE(m.id(m.switch_tuple(e, PrimitiveType::Vertex), PrimitiveType::Vertex) == 1);
+        REQUIRE(m.id(e, PE) == 0);
+        REQUIRE(m.id(e, PV) == 0);
+        REQUIRE(m.id(m.switch_tuple(e, PV), PV) == 1);
```
mtao, line 87: i'm thinking maybe we should just create inline functions in TriMesh for switch_vertex / switch_edge / switch_face (and same for ID). they're just convenience tools anyway :shrug:

tests/test_simplicial_complex.cpp
```diff
@@ -233,38 +223,35 @@ TEST_CASE("open_star", "[SC][star]")
...
-        m.id(
-            m.switch_tuple(m.switch_tuple(t, PrimitiveType::Edge), PrimitiveType::Vertex),
-            PrimitiveType::Vertex) == 2);
+    REQUIRE(m.id(t, PV) == 0);
+    REQUIRE(m.id(m.switch_tuple(t, PV), PV) == 1);
+    REQUIRE(m.id(m.switch_tuple(m.switch_tuple(t, PE), PV), PV) == 2);
 
 
-    SimplicialComplex sc_v = SimplicialComplex::closed_star(Simplex(PrimitiveType::Vertex, t), m);
+    SimplicialComplex sc_v = SimplicialComplex::closed_star(Simplex(PV, t), m);
```
mtao, line 249: would it make sense to have a 
```
static Simplex Simplex::vertex(const Tuple& t) { return Simplex(PrimitiveType::Vertex, t); }
```
to simplify these incredibly frequent circumstances?

mtao (approved):
generally lgtm, most comments are on other historical things that bothered me that we can change further

src/wmtk/TriMesh.cpp
daniel-zint: I think there is some misunderstanding. the `.get_simplices(...)` does not cast simplices into faces. It just extracts all faces that are inside a simplicial complex. I think the behavior is exactly what you want it to be.

tests/test_2d_operation_construction.cpp
daniel-zint: I'll do that after finishing this PR

tests/test_simplicial_complex.cpp
daniel-zint: Sounds reasonable. I'll also do that after finishing this PR.

daniel-zint: Whatever, I'll include all the stuff in this PR right away..

tests/test_2d_operation_construction.cpp
daniel-zint: done

tests/test_simplicial_complex.cpp
daniel-zint: done

daniel-zint:
I also added `next_edge` and `prev_edge` to TriMesh and with that, I close that PR.

PR #341, opened 2023-08-04
Explicit Capacities Serialization

mtao:
Previously capacties were re-computed during de-serialization. This conserves some memory but in retrospect seems to be an odd middle point between garbage collecting/compactifying the mesh and preserving the entire state.

This PR makes it completely the user's choice whether to garbage collect the mesh or not by adding explicit serialization of capacities to the mesh.

Also made accessors/attributes use reserved_size instead of size because size is ambiguous without flag data


daniel-zint:
looks all good to me

PR #345, opened 2023-08-10
Hackathon/test example trimesh utils

zlyfunction:
add util functions in test_example_meshes.cpp
- trimesh_simply_connected_components
- trimesh_n_bd_loops
- trimesh_euler_char
- trimesh_genus
add two test tri_mesh


tests/test_example_meshes.cpp
```diff
@@ -16,6 +17,121 @@ struct MeshDebugInfo
...
+}
+
+
+int trimesh_simply_connected_components(const DEBUG_TriMesh& m, std::vector<int> &component_ids)
+{
+    component_ids.resize(m.capacity(PrimitiveType::Face), -1);
+    auto all_face_tuples = m.get_all(PrimitiveType::Face);
+    int component_id = 0;
+    // BFS
+    for (size_t i = 0; i < all_face_tuples.size(); ++i) {
```
mtao, line 35: why not use a range for loop?

tests/test_example_meshes.cpp
```diff
@@ -16,6 +17,121 @@ struct MeshDebugInfo
...
+    return component_id;
+}
+
+int trimesh_simply_connected_components(const DEBUG_TriMesh& m)
+{
+    std::vector<int> component_ids;
+    return trimesh_simply_connected_components(m, component_ids);
+}
+
+int trimesh_n_bd_loops(const DEBUG_TriMesh& m)
```
mtao, line 76: please use full words for function names (and typically for variable names as well).
we're not using fortran 77 :)

tests/test_example_meshes.cpp
```diff
@@ -16,6 +17,121 @@ struct MeshDebugInfo
...
+{
+    std::vector<int> bd_loops(m.capacity(PrimitiveType::Edge), -1);
+    auto all_edge_tuples = m.get_all(PrimitiveType::Edge);
+    int bd_loop_id = 0;
+    for (size_t i = 0; i < all_edge_tuples.size(); ++i) 
+    {
+        if (!m.is_boundary(all_edge_tuples[i])) {
```
mtao, line 83: why not range for loop?
also request `all_edge_tuples[i]` multiple times, just create a ref (or just const ref in a range for loop)

mtao: oh you even declare this as cur_edge as a value a few lines down

tests/test_example_meshes.cpp
```diff
@@ -16,6 +17,121 @@ struct MeshDebugInfo
...
+        }
+        long eid = m.id(all_edge_tuples[i], PrimitiveType::Edge);
+        if (bd_loops[eid] != -1) {
+            continue;
+        } // visited
+
+        Tuple cur_edge = all_edge_tuples[i];
+        long cur_eid = m.id(cur_edge, PrimitiveType::Edge);
+        // find one boundary loop
+        while (bd_loops[m.id(all_edge_tuples[i], PrimitiveType::Edge)] == -1 || cur_eid != m.id(all_edge_tuples[i], PrimitiveType::Edge))
```
mtao, line 94: declare variables when values are used twice.... `m.id(all_edge_tuples[i], PrimitiveType::Edge)` doesn't need to appear more than once

tests/test_example_meshes.cpp
```diff
@@ -16,6 +17,121 @@ struct MeshDebugInfo
...
+            cur_eid = m.id(cur_edge, PrimitiveType::Edge);
+        }
+        
+        bd_loop_id++;
+    }
+
+    return bd_loop_id;
+}
+
+int trimesh_euler_char(const DEBUG_TriMesh& m)
```
mtao, line 116: euler_characteristic...

tests/tools/TriMesh_examples.hpp
```diff
@@ -88,4 +88,11 @@ TriMesh hex_plus_two();
 //   \ / \ / \ /
 //    7---8---9
 TriMesh edge_region();
+
+//  0---2  3---4
+//  | /    | \ |
+//  1      5---6
+TriMesh three_triangles_with_two_components();
+
```
mtao, line 96: what's the topology for nine with a hole? would be nice to be consistent about having figures for everything

mtao (approved):
minor changes to naming/style required but generally good  to go

tests/test_example_meshes.cpp
```diff
@@ -16,6 +17,121 @@ struct MeshDebugInfo
...
+            {
+                if (!m.is_boundary(f_tuple))
+                {   
+                    auto adj_face_tuple = m.switch_tuple(f_tuple, PrimitiveType::Face);
+                    long adj_fid = m.id(adj_face_tuple, PrimitiveType::Face);
+                    if (component_ids[adj_fid] == -1) {
+                        q.push(adj_fid);
+                    }  
+                }
+                f_tuple = m.switch_tuple(m.switch_tuple(f_tuple, PrimitiveType::Vertex), PrimitiveType::Edge);
```
daniel-zint, line 60: use `m.switch_vertex` and `m.switch_edge` for better readability

tests/test_example_meshes.cpp
```diff
@@ -16,6 +17,121 @@ struct MeshDebugInfo
...
+            } 
+        }
+        
+        component_id++;
```
daniel-zint, line 64: `++component_id`

tests/test_example_meshes.cpp
daniel-zint: Exception: you have a lot of variables and using acronyms increases readability. In that case, add explanations to the variable definitions. But this is only rarely the case and usually already a sign that the code needs to be refactored.

tests/test_example_meshes.cpp
```diff
@@ -16,6 +17,121 @@ struct MeshDebugInfo
...
+    int n_bd_loops = trimesh_n_bd_loops(m);
+    return simplex_counts[0] - simplex_counts[1] + simplex_counts[2] + n_bd_loops;
+}
+
+int trimesh_genus(const DEBUG_TriMesh& m)
+{
+    if (trimesh_simply_connected_components(m) != 1)
+    {
+        wmtk::logger().error("not a connected surface\n");
+        throw std::runtime_error("GenusComputeError");
```
daniel-zint, line 128: A minor thing but logging and throwing an error seems to me a bit overkill. I feel like throwing is enough. Anyways, the messages of the two should be the same I think.

tests/test_example_meshes.cpp
```diff
@@ -16,6 +17,121 @@ struct MeshDebugInfo
...
+    }
+    return (2 - trimesh_euler_char(m)) / 2;
+}
+
+
+
 void run_debug_trimesh(const DEBUG_TriMesh& m, const MeshDebugInfo& info)
```
daniel-zint, line 135: give it a more explanatory name or add a comment

tests/test_example_meshes.cpp
```diff
@@ -80,9 +195,32 @@ TEST_CASE("test_debug_trimeshes_one_ear")
...
-    info.simplex_counts = std::array<long, 3>{{3, 3, 3}};
+    info.simplex_counts = std::array<long, 3>{{4, 5, 2}};
     run_debug_trimesh(m, info);
 }
+
+TEST_CASE("test_debug_trimeshes_two_components")
+{
+    DEBUG_TriMesh m;
+    m = three_triangles_with_two_components();
+    MeshDebugInfo info;
```
daniel-zint, line 210: `info` is not used in that test

tests/tools/TriMesh_examples.hpp
daniel-zint: \+

daniel-zint (commented):
some more comments, minor things though

tests/test_example_meshes.cpp
mtao: sure, exception to the exception: only for short functions :)

tests/test_example_meshes.cpp
```diff
@@ -16,6 +17,121 @@ struct MeshDebugInfo
...
+                    }  
+                }
+                f_tuple = m.switch_tuple(m.switch_tuple(f_tuple, PrimitiveType::Vertex), PrimitiveType::Edge);
+            } 
+        }
+        
+        component_id++;
+    }
+
+    return component_id;
```
mtao, line 67: actually, why are we returning the number of components AND the component id maps. just return the component ids (and avoid this return type parameter mess). size() suffices

tests/test_example_meshes.cpp
mtao: i agree but only if we have a top level try/catch that forwards exceptions to the logger.
there are cases where you want to send someone an output log (maybe a log from a file) and that won't contain the exceptions)..

on the other hand, the runtime error message is not practical (you're declaring it like you're specifying the runtime error is a class name, even though the aforementioned log message spits out more information. might as well make these two statements the same

mtao:
BTW the computation of topological invariants might be helpful outside fof unit etsts so it might be nice to leave them as utils sort of functions. It might also be nice to write a more efficient implementation of simplex counts at some point reading off flags, but this is a fairly low priority.

tests/test_example_meshes.cpp
daniel-zint: True, I didn't consider that. Then I guess having the logger and the error simultaneously with the same message might be the best option.

tests/test_example_meshes.cpp
zlyfunction: components maps is a per face attributes, to get the return size the right way is to map.max()+1, which cost extra time?

zlyfunction:
updated based on comments

tests/tools/TriMesh_examples.hpp
mtao: Uh, could we have a simpler picture? I think Daniel had a new test with a figure that could be cut down for this purpose

mtao (commented):
Responses to a few updates

tests/test_example_meshes.cpp
mtao: alternative is to have a global catch that rethrows? i'm not sure how this works with debuggers though

PR #346, opened 2023-08-16
Hackathon/increment cell hash

daniel-zint:
Add function `increment_cell_hash` to `Operation` class, so that operations can increment the hash. This is required in vertex relocation. Split and collapse operations do not require this as they perform the hash update already in the mesh operations.


mtao (approved):
I think update being an increment is an implementation detail so I'm not sure we want to make the interface clarify that, but otherwise this looks good

PR #347, opened 2023-08-17
Hackathon/sync mesh interface

zlyfunction:
Interface for Sync_mesh


src/wmtk/Mesh.hpp
```diff
@@ -24,6 +24,31 @@ class Mesh
...
+
+    // Interface for SyncMesh
+
+    // TODO: Need a wrapper for MeshAttributeHandle<Tuple>?
+
+    // size of dim(Mesh), store base_tuple for each simplex
+    std::vector<MeshAttributeHandle<long>> base_tuple_handle;
+
+    // if parent_mesh is not null, then this mesh is a child_mesh of parent_mesh
+    std::shared_ptr<Mesh> parent_mesh;
```
mtao, line 36: Should just be a raw ptr or maybe a weak ptr

src/wmtk/Mesh.hpp
```diff
@@ -24,6 +24,31 @@ class Mesh
     friend class ParaviewWriter;
     friend class MeshReader;
 
+
+    // Interface for SyncMesh
+
+    // TODO: Need a wrapper for MeshAttributeHandle<Tuple>?
+
+    // size of dim(Mesh), store base_tuple for each simplex
+    std::vector<MeshAttributeHandle<long>> base_tuple_handle;
```
mtao, line 33: We need to decide on a name for these tuples, I'm not happy with base, perhaps sticking it in a container would help isolate a name like handle as this is a attribute for a collection of handles?

src/wmtk/Mesh.hpp
```diff
@@ -24,6 +24,31 @@ class Mesh
...
+    // i.e. Dim(map_to_child[i]) == Dim(child_meshes[i])
+    std::vector<MeshAttributeHandle<long>> map_to_child_handles;
+    // store the map to the base_tuple of the parent_mesh
+    MeshAttributeHandle<long> map_to_parent_handle;
+    
+    // register child_meshes and the map from child_meshes to this mesh, child_mesh_simplex 
+    void register_child_mesh(std::shared_ptr<Mesh> child_mesh, const std::vector<long>&child_mesh_simplex_map);
+    
+    // helper function to check if this mesh is a valid child_mesh of parent_mesh
+    // i.e. the connectivity of this mesh is a subset of this in parent_mesh
```
mtao, line 49: This could be a free function?

mtao (commented):
Some comments from phone

src/wmtk/Mesh.hpp
teseoch: I don't like the pointer to parent

src/wmtk/utils/tuple_sync.hpp
```diff
@@ -0,0 +1,4 @@
+#include <Mesh.hpp>
+
+// Map t_source from M_source to M_target
+Tuple map_tuple_between_meshes(const Mesh& M_source, const Mesh& M_target, const Tuple& t_source);
```
mtao, line 4: namespace
i'd prefer `source_mesh` and `target_mesh` rather than mix a single character acronym

mtao (commented):
some comments on the actual interfaces

src/wmtk/Mesh.hpp
danielepanozzo: Since the submeshes do not need to reference the parent (it is the parent calling functions in them), what about moving all functions related to submeshes in the parent mesh, and replacing this pointer with a bool indicating if it is a submesh or not? Or maybe even the bool is not necessary.

src/wmtk/Mesh.hpp
```diff
@@ -24,6 +24,31 @@ class Mesh
...
+    std::vector<std::shared_ptr<Mesh>> child_meshes;
+
+    // size of child_meshes.size(), store the map to the base_tuple of the child_meshes (on the top level simplex of each child_mesh)
+    // i.e. Dim(map_to_child[i]) == Dim(child_meshes[i])
+    std::vector<MeshAttributeHandle<long>> map_to_child_handles;
+    // store the map to the base_tuple of the parent_mesh
+    MeshAttributeHandle<long> map_to_parent_handle;
+    
+    // register child_meshes and the map from child_meshes to this mesh, child_mesh_simplex 
+    void register_child_mesh(std::shared_ptr<Mesh> child_mesh, const std::vector<long>&child_mesh_simplex_map);
```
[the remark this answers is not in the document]
danielepanozzo: I also vote for mapping tuple to tuple, it is wasting space but it naturally encodes orientation.

src/wmtk/MultiMeshManager.cpp
```diff
@@ -0,0 +1,145 @@
...
+        if (source_mesh_base_tuple.is_null_tuple())
+        {
+            return Tuple(); // return null tuple
+        }
+        
+        if (source_mesh_ptype == PrimitiveType::Face && target_mesh_ptype == PrimitiveType::Face)
+        {
+            Tuple cur_tuple = source_tuple;
+            std::vector<char> record_switch_operations;
+            while (cur_tuple != source_mesh_base_tuple)
```
mtao, line 91: This should eventually be something like `get_navigation_between_tuples_in_single_simplex(PrimitiveType, const Tuple&, const Tuple&) -> std::vector<PrimitiveType>`. But for now perhaps a function that takes in a `Mesh` + `PrimitiveType` should suffice so you can use mesh navigation to find the sequence.

In the end though, there should be a O(1) solution for computing the navigation (we can precompute all sequences).

src/wmtk/MultiMeshManager.cpp
```diff
@@ -0,0 +1,145 @@
...
+                    break;
+                }
+                cur_tuple = source_mesh.switch_tuple(cur_tuple, PrimitiveType::Edge);
+                record_switch_operations.push_back('e');
+            }
+
+            Tuple ret_tuple = target_mesh_base_tuple;
+            for (int i = record_switch_operations.size() - 1; i >= 0; --i)
+            {
+                if (record_switch_operations[i] == 'v')
```
mtao, line 106: with `std::vector<PrimitiveType> record_switch_operations` thisfunction doesn't require branching.

src/wmtk/MultiMeshManager.hpp
```diff
@@ -0,0 +1,66 @@
...
+    MultiMeshManager(const MultiMeshManager& o);
+    MultiMeshManager(MultiMeshManager&& o);
+    MultiMeshManager& operator=(const MultiMeshManager& o);
+    MultiMeshManager& operator=(MultiMeshManager&& o);
+
+    //=========================================================
+    // Storage of MultiMesh
+    //=========================================================
+
+    bool is_parent_mesh() const { return m_is_parent_mesh; }
```
mtao, line 27: if MeshAttributeHandle is aware of whether it's in a `null` state then this bit is redundant.

src/wmtk/MultiMeshManager.hpp
```diff
@@ -0,0 +1,66 @@
...
+    long child_id() const { return m_child_id; }
```
mtao, line 28: put these implementations in the cpp

src/wmtk/MultiMeshManager.hpp
```diff
@@ -0,0 +1,66 @@
...
+    // helper function to read/write the map from source_tuple to target_tuple to the attribute
+    static void write_tuple_map_attribute(MeshAttributeHandle<long> map_handle, Mesh& source_mesh, const Tuple& source_tuple, const Tuple& target_tuple);
```
mtao, line 30: this doesn't have to be static, and there should be an assert in this function that 
`&source_mesh.multi_mesh_manager == this`

src/wmtk/MultiMeshManager.hpp
```diff
@@ -0,0 +1,66 @@
...
+    static std::tuple<Tuple, Tuple> read_tuple_map_attribute(MeshAttributeHandle<long> map_handle, const Mesh& source_mesh, const Tuple& source_tuple);
```
mtao, line 31: same as above

src/wmtk/MultiMeshManager.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+    // Child Meshes
+    std::vector<std::shared_ptr<Mesh>> child_meshes;
```
mtao, line 34: one thing that w emight want to do is use `enable_shared_from_this` 
https://en.cppreference.com/w/cpp/memory/enable_shared_from_this

one side effect is that we could just pass in meshes into functions and check to make sure that they come from a shared_ptr internally to assert / check for failure

src/wmtk/TriMesh.hpp
```diff
@@ -16,6 +16,7 @@ class TriMesh : public Mesh
     TriMesh& operator=(const TriMesh& o);
     TriMesh& operator=(TriMesh&& o);
 
+    PrimitiveType top_simplex_type() const override { return PrimitiveType::Face; }
```
mtao, line 19: i have a preference for shoving all of these implementations in to cpp files. makes headers cleaner, but not everyone thinks this way

mtao (commented):
some comments on first half. going to look at the executor stuff soon

src/wmtk/MultiMeshManager.cpp
```diff
@@ -0,0 +1,145 @@
...
+        return true;
+    }
+
+    std::vector<Tuple> MultiMeshManager::map_edge_tuple_to_all_children(const Mesh& parent_mesh, const Tuple& edge_tuple)
+    {
+        std::vector<Tuple> ret;
+        for (auto child_mesh_ptr : parent_mesh.multi_mesh_manager.child_meshes)
+        {
+            long child_id = child_mesh_ptr->multi_mesh_manager.child_id();
+            auto map_to_child_handle = parent_mesh.multi_mesh_manager.map_to_child_handles[child_id];
```
mtao, line 139: the map_tuple_between_meshes probably almost always works with an accessor instead of a handle passed in because i don't think it'll ever be called from multiple threads. so it'd probably be easiest if you can request a map's accessor from the multi map manager

src/wmtk/MultiMeshManager.cpp
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    bool MultiMeshManager::is_child_mesh_valid(const Mesh& parent_mesh, const Mesh& child_mesh)
+    {
+        // TODO: implement this
+        
+        return true;
+    }
+
+    std::vector<Tuple> MultiMeshManager::map_edge_tuple_to_all_children(const Mesh& parent_mesh, const Tuple& edge_tuple)
```
mtao, line 133: i don't understand this function's implementation. It return sa collection of tuples from all of the children - so if you have two child meshes hten you just get a bunch of tuples from either mesh without any identification wher ehtey came from? Perhaps you want to specify a specific child mesh as an argument here?

src/wmtk/TriMeshOperationExecutor.cpp
```diff
@@ -367,6 +367,101 @@ void TriMesh::TriMeshOperationExecutor::replace_incident_face(
...
 Tuple TriMesh::TriMeshOperationExecutor::split_edge()
+{
+    if (!m_mesh.multi_mesh_manager.is_parent_mesh()) {
+        return split_edge_single_mesh();
+    }
+    else
+    {
+        bool is_boundary_on_parent = m_mesh.is_boundary(m_operating_tuple);
+        
+        std::vector<Tuple> vec_t_child;
```
mtao, line 378: vector so plural variable name? and just specify what things are. i have no idea what a vec_t is
`child_tuples`?

src/wmtk/MultiMeshManager.cpp
mtao: oh. i misunderstood what this function does - it assumes that `child_meshes[i]->child_id() == i` which i guess is true right now and returns every copy of a vector if it exists.

i think that constructing this vector like this is not necessary - but i will comment on it in your Executor implementations of operations because the structure isn't necessary

src/wmtk/TriMeshOperationExecutor.cpp
```diff
@@ -367,6 +367,101 @@ void TriMesh::TriMeshOperationExecutor::replace_incident_face(
...
+        std::vector<Tuple> vec_t_opp_child;
+        vec_t_child = MultiMeshManager::map_edge_tuple_to_all_children(m_mesh, m_operating_tuple);
+        if (!is_boundary_on_parent)
```
mtao, line 381: i'd much prefer a "gather every version of this face" rather than this boundary searching logic - it doesn't generalize to other cases AND it makes the implementation much more complex. You should just be doing a for loop over a vector containing the various veresion of an edge rather than passing around two variables and copy pasting the logic for both cases everywhere. that's just asking for bugs

src/wmtk/TriMeshOperationExecutor.cpp
```diff
@@ -367,6 +367,101 @@ void TriMesh::TriMeshOperationExecutor::replace_incident_face(
...
+        {
+            Tuple t_opp = m_mesh.switch_edge(m_operating_tuple);
+            vec_t_opp_child = MultiMeshManager::map_edge_tuple_to_all_children(m_mesh, t_opp);
```
mtao, line 384: there's no need to compute every version of this edge right now. just do it in the child_mesh_ptr loop and find a single one at a time

src/wmtk/TriMeshOperationExecutor.cpp
```diff
@@ -367,6 +367,101 @@ void TriMesh::TriMeshOperationExecutor::replace_incident_face(
...
+        }
+        else
+        {
+            vec_t_opp_child = std::vector<Tuple>(vec_t_child.size(), Tuple()); // fill with null Tuples
+        }
+
+        // do split on parent_mesh
+        Tuple ret_tuple = split_edge_single_mesh();
+
+        for (auto child_mesh_ptr : m_mesh.multi_mesh_manager.child_meshes)
```
mtao, line 394: i was imagining something closer to the following

```
auto all_copies_of_operating_tuple = m_mesh.get_tuple_for_all_simplices(m_operating_tuple, PrimitiveType::Face);
for(const auto& child_mesh_ptr: m_mesh.child_mesh_ptrs()) {
  for(const Tuple& t: all_copies_of_operating_tuple) {
    Tuple child_tuple = get_child_tuple(m_mesh, *child_mesh_ptr, t);
   if(child_tuple.is_not_null()) {
    auto child_executor = child_mesh->split_edge(child_tuple);
    std::array<Tuple,2> new_tuples = executor.get_new_face_tuples(some_identifier_for_face);
    std::array<Tuple,2> new_child_tuples = child_executor.get_new_face_tuples(some_other_identifier_for_face);
   // map new_tuples[i] to new_child_tuples[i];
    ... hash stuff
   }
  }
}
```

src/wmtk/TriMeshOperationExecutor.cpp
```diff
@@ -367,6 +367,101 @@ void TriMesh::TriMeshOperationExecutor::replace_incident_face(
...
+            {
+                // this child_mesh is a TriMesh
+                TriMesh child_tri_mesh = *std::static_pointer_cast<TriMesh>(child_mesh_ptr);
+                if (!t_child.is_null_tuple())
+                {
+                    TriMesh::TriMeshOperationExecutor executor_child(child_tri_mesh, t_child);
+                    executor_child.split_edge();
+                    child_cell_ids_to_update_hash = executor_child.cell_ids_to_update_hash;
+                    
+                    child_new_cell_ids.push_back(std::make_pair(executor_child.m_incident_face_datas[0].split_f1, executor_child.m_incident_face_datas[0].split_f0));
```
mtao, line 413: why is it f1 then f0? seems like an arbitrary convention choice that needs to be discussed

src/wmtk/TriMeshOperationExecutor.cpp
```diff
@@ -367,6 +367,101 @@ void TriMesh::TriMeshOperationExecutor::replace_incident_face(
...
+                // update_hash on new cells
+                for (long i = 0; i < long(m_incident_face_datas.size()); i++)
+                {
+                    const auto& split_fs = child_new_cell_ids[i];
+                    long split_f1 = split_fs.first;
+                    long split_f0 = split_fs.second;
+                    if (split_f1 == -1) continue;
+                    
+                    const auto& incident_face_data = m_incident_face_datas[i];
+                    long split_f0_parent = incident_face_data.split_f0;
```
mtao, line 441: pick between whether you are going to call something child, parent, or skip the annotation (line 436 the child stuff skips being called child, but here you say parent. earlier code says parent and child. but sometimes you just say child and not parent? differnt parts of this function therefore have different variable name conventions

src/wmtk/TriMeshOperationExecutor.cpp
```diff
@@ -402,7 +497,88 @@ Tuple TriMesh::TriMeshOperationExecutor::split_edge()
     // return m_mesh.with_different_cid(m_operating_tuple, m_incident_face_datas[0].split_f0);
 }
 
+void TriMesh::TriMeshOperationExecutor::update_hash_in_map(TriMesh& child_mesh, const std::vector<long>& child_cell_ids_to_update_hash)
+{
+    long child_id = child_mesh.multi_mesh_manager.child_id();
+    auto child_hash_accessor = child_mesh.get_cell_hash_accessor();
+    for (auto child_cell_id : child_cell_ids_to_update_hash)
+    {
+        auto [t_child_old, t_parent_old] = MultiMeshManager::read_tuple_map_attribute(child_mesh.multi_mesh_manager.map_to_parent_handle, child_mesh, child_mesh.tuple_from_id(child_mesh.top_simplex_type(), child_cell_id));
```
mtao, line 506: the point of the manager is to enable some easy functions in the Mesh class that simplify the notation... add functions to Mesh that do these sorts of reads rather than just call these massive manager names and have so many arguments

src/wmtk/TriMeshOperationExecutor.hpp
```diff
@@ -90,9 +90,11 @@ class TriMesh::TriMeshOperationExecutor
         const long new_eid);
 
     void connect_ears();
-    Tuple split_edge();
 
+    Tuple split_edge();
     Tuple collapse_edge();
+    Tuple split_edge_single_mesh();
```
mtao, line 96: i think these should be called topology updates. and I'd prefer the multimesh logic be handled in the Mesh class rather than in the Executor. I think some pseudocode i wrote earlier in this reivew already assumed that.

src/wmtk/Tuple.cpp
```diff
@@ -42,4 +42,14 @@ bool Tuple::same_ids(const wmtk::Tuple& t) const
            std::tie(t.m_local_vid, t.m_local_eid, t.m_local_fid, t.m_global_cid);
 }
 
+bool Tuple::is_null_tuple() const
+{
+    return m_local_vid == -1 && m_local_eid == -1 && m_local_fid == -1 && m_global_cid == -1 && m_hash == -1;
+}
+
+Tuple Tuple::tuple_update_hash(long new_hash) const
```
mtao, line 50: Tuple having a function specifying tuple is unecessary `with_updated_hash(long new_hash)`

src/wmtk/Tuple.cpp
```diff
@@ -42,4 +42,14 @@ bool Tuple::same_ids(const wmtk::Tuple& t) const
            std::tie(t.m_local_vid, t.m_local_eid, t.m_local_fid, t.m_global_cid);
 }
 
+bool Tuple::is_null_tuple() const
```
mtao, line 45: `is_null`

mtao (commented):
more comments

PR #349, opened 2023-08-22
Hackathon/isotropic remeshing performance

daniel-zint:
Some minor performance tweaks with major impact. 

(More performance improvement can be achieved by refactoring the `SimplicialComplex`)


src/wmtk/Mesh.hpp
```diff
@@ -112,6 +114,31 @@ class Mesh
...
+    void update_cell_hashes_slow(const std::vector<Tuple>& cells);
+
+    /**
+     * @brief return the same tuple but with updated hash
+     *
+     * @param tuple tuple with potentially outdated hash
+     * @param hash_accessor hash accessor
+     * @return tuple with updated hash
+     */
+    Tuple resurrect_tuple(const Tuple& tuple, Accessor<long>& hash_accessor) const;
```
mtao, line 136: this will need to be a constaccessor. just copypaste two versions for now and make an issue (assigned to me) to fix accessor inheritance so accessors can be implicitly converted to constaccessors

src/wmtk/Mesh.hpp
```diff
@@ -158,6 +185,10 @@ class Mesh
     }
     Tuple switch_edge(const Tuple& tuple) const { return switch_tuple(tuple, PrimitiveType::Edge); }
     Tuple switch_face(const Tuple& tuple) const { return switch_tuple(tuple, PrimitiveType::Face); }
+    Tuple switch_tetrahedron(const Tuple& tuple) const
```
mtao, line 188: i'd like to remove all implementations of functions from the class definitions - my preference is that if you're going to force inlining then just put it at the end of the header (or in a `Mesh_impl.hpp` that is included at the bottom of the header)

mtao: this is a bit more visible in this case because your formatter decided to make this function take 3 lines instead of the previous singel line ones - but i'd like to remove those as well TBH

src/wmtk/Mesh.hpp
mtao: oh and mark the non-const accessor one as deprecated just so we know to delete it ASAP?

src/wmtk/PointMesh.cpp
```diff
@@ -49,16 +49,10 @@ void PointMesh::initialize(long count)
...
-    return true;
-}
-
-bool PointMesh::is_outdated(const Tuple& tuple) const
-{
-    const long fid = id(tuple, PrimitiveType::Face);
-    ConstAccessor<long> ha = get_cell_hash_accessor();
-    return ha.scalar_attribute(fid) == tuple.m_hash;
+    const long cid = tuple.m_global_cid;
+    return tuple.m_hash == get_cell_hash(cid, hash_accessor);
```
mtao, line 55: to be honest Mesh could have a `Mesh::is_hash_valid(const Tuple&) -> bool` because all implementations just read off the global cid

src/wmtk/TetMesh.cpp
```diff
@@ -175,9 +179,11 @@ Tuple TetMesh::face_tuple_from_id(long id) const
 
     if (lvid < 0 || leid < 0 || lfid < 0) throw std::runtime_error("face_tuple_from_id failed");
 
-    Tuple f_tuple = Tuple(lvid, leid, lfid, t, get_cell_hash_slow(t));
+    ConstAccessor<long> hash_accessor = get_const_cell_hash_accessor();
```
mtao, line 182: in the long term this construction / passing of accessors could become more prevalent (even `face_tuple_from_id` might want to haev a hash accessor

src/wmtk/TriMesh.cpp
```diff
@@ -121,13 +129,16 @@ Tuple TriMesh::switch_tuple(const Tuple& tuple, PrimitiveType type) const
         }
         assert(lvid_new != -1);
         assert(leid_new != -1);
+
+        ConstAccessor<long> hash_accessor = get_const_cell_hash_accessor();
```
mtao, line 133: sadly this will probably turn into `switch_tuple_slow`. Actually, perhaps it would make sense to have a small wrapper around TriMesh that holds a hashaccessor (and perhaps some of the other common ones) just to make this passing a bit easier

src/wmtk/TriMesh.cpp
```diff
@@ -234,9 +245,14 @@ Tuple TriMesh::vertex_tuple_from_id(long id) const
...
-            assert(is_ccw(v_tuple));
-            assert(is_valid(v_tuple));
+            Tuple v_tuple = Tuple(
+                i,
+                leid,
+                -1,
+                f,
+                get_cell_hash_slow(
+                    f)); // TODO replace by function that takes hash accessor as parameter
+            assert(is_ccw(v_tuple)); // is_ccw also checks for validity
```
mtao, line 255: why did the valid check get removed?

src/wmtk/TriMesh.cpp
```diff
@@ -275,22 +293,25 @@ Tuple TriMesh::face_tuple_from_id(long id) const
...
 }
 
-bool TriMesh::is_valid(const Tuple& tuple) const
+bool TriMesh::is_valid(const Tuple& tuple, ConstAccessor<long>& hash_accessor) const
 {
     int offset = tuple.m_local_vid * 3 + tuple.m_local_eid;
-    return tuple.m_local_vid >= 0 && tuple.m_local_eid >= 0 && tuple.m_global_cid >= 0 &&
-           autogen::auto_2d_table_ccw[offset][0] >= 0;
-}
+    const bool is_connectivity_valid = tuple.m_local_vid >= 0 && tuple.m_local_eid >= 0 &&
```
mtao, line 303: your formmatter isn't consistent in how it separates out these multi-line things (netiher was the code before tbh)
'id rather see either all statements in a mult-line form or all in a single line

src/wmtk/TriMeshOperationExecutor.cpp
```diff
@@ -52,14 +52,15 @@ TriMesh::TriMeshOperationExecutor::get_incident_face_data(Tuple t)
 // constructor
 TriMesh::TriMeshOperationExecutor::TriMeshOperationExecutor(
     TriMesh& m,
-    const Tuple& operating_tuple)
+    const Tuple& operating_tuple,
+    Accessor<long>& hash_acc)
```
mtao, line 56: maybe keep a constructor that creates the hash accessor as well? this will mess with leyi's multimesh stuff a lot

src/wmtk/TriMeshOperationExecutor.hpp
```diff
@@ -7,7 +7,7 @@ namespace wmtk {
 class TriMesh::TriMeshOperationExecutor
 {
 public:
-    TriMeshOperationExecutor(TriMesh& m, const Tuple& operating_tuple);
+    TriMeshOperationExecutor(TriMesh& m, const Tuple& operating_tuple, Accessor<long>& hash_acc);
```
mtao, line 10: keep the other constructor just for now (you can annotate with deprecated if you want)

src/wmtk/operations/Operation.cpp
```diff
@@ -50,10 +51,12 @@ bool Operation::after() const
 
 void Operation::update_cell_hash(const std::vector<Tuple>& cells)
 {
-    auto acc = m_mesh.get_cell_hash_accessor();
-    for (const Tuple& t : cells) {
-        ++acc.scalar_attribute(t);
-    }
+    m_mesh.update_cell_hashes(cells, m_hash_accessor);
```
mtao, line 54: lol fix the operation function as well to say hashes

src/wmtk/operations/Operation.cpp
```diff
@@ -50,10 +51,12 @@ bool Operation::after() const
...
+}
+
+Tuple Operation::resurrect_tuple(const Tuple& tuple)
```
mtao, line 57: we need to be extra careful that teh difference betwen `update_cell_hash` and `resurrect_tuple` (or whatever other name we come up with) have very obviously distinct names to indicate their purposes

mtao: also, why can't this be const?

tests/test_2d_operation_construction.cpp
```diff
@@ -62,8 +63,9 @@ DEBUG_TriMesh test_collapse(const DEBUG_TriMesh& mesh, const Tuple& e, bool shou
     REQUIRE(m.is_connectivity_valid());
     REQUIRE(should_succeed == result);
     if (should_succeed) {
-        TriMesh m2 = mesh;
-        m2.collapse_edge(e);
+        DEBUG_TriMesh m2 = mesh;
+        Accessor<long> hash_accessor = m2.get_cell_hash_accessor();
```
mtao, line 67: these cases make me think taht collapse_edge_slow and split_edge_slow might be helpful just to remove a bit of this boilerplate

tests/test_2d_operations.cpp
```diff
@@ -16,7 +16,9 @@ using namespace wmtk::tests;
 
 using TM = TriMesh;
 using MapResult = typename Eigen::Matrix<long, Eigen::Dynamic, 1>::MapType;
-using TMOE = decltype(std::declval<DEBUG_TriMesh>().get_tmoe(wmtk::Tuple()));
+using TMOE = decltype(std::declval<DEBUG_TriMesh>().get_tmoe(
```
mtao, line 19: i know i wrote this ugliness, but it's getting even more ugly :sob: . but this is the right thing haha

src/wmtk/SimplicialComplex.cpp
```diff
@@ -417,6 +417,35 @@ std::vector<Simplex> SimplicialComplex::vertex_one_ring(const Mesh& m, Tuple t)
     return std::vector<Simplex>(vs.begin(), vs.end());
 }
 
+std::vector<Simplex> SimplicialComplex::vertex_one_ring(const TriMesh& m, Tuple t)
```
mtao, line 420: does it make sense to specifiy this is the set of vertices that are one ring of vertices? i always forget these definitions

mtao (approved):
lgtm; mostly minor things. the accessor argument thing is, as expected, rather ugly and increases code bloat. we should identify which functions need accessor-arg versions (and the accessor-arg versions should perhaps all be protected/private?)

src/wmtk/TriMeshOperationExecutor.cpp
daniel-zint: I wanted to do that but it's not possible because the accessor is stored by reference. Or do I miss something obvious how it can be done?

src/wmtk/TriMeshOperationExecutor.hpp
daniel-zint: as mentioned in the previous comment, it is not possible because the accessor is a reference

tests/test_2d_operations.cpp
daniel-zint: it took me quite some time to figure it out 😅

src/wmtk/TriMesh.cpp
daniel-zint: Because `is_ccw` also contains a valid check. It was done twice.

src/wmtk/Mesh.hpp
```diff
@@ -67,8 +67,8 @@ class Mesh
     // operated on, and return a single tuple that refers to the new topology.
     // This returned tuple has specific meaning for each derived Mesh class
 
-    virtual Tuple split_edge(const Tuple& t) = 0;
-    virtual Tuple collapse_edge(const Tuple& t) = 0;
+    virtual Tuple split_edge(const Tuple& t, Accessor<long>& hash_accessor) = 0;
```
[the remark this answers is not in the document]
daniel-zint: I agree. I already thought I can merge that PR but then decided to clean up the accessor stuff first. That was a mistake. A new PR would have been better.

src/wmtk/Mesh.cpp
```diff
@@ -149,10 +155,42 @@ Accessor<long> Mesh::get_cell_hash_accessor()
...
+    }
+}
+
+void Mesh::update_cell_hashes_slow(const std::vector<Tuple>& cells)
+{
+    Accessor<long> hash_accessor = get_cell_hash_accessor();
+    update_cell_hashes(cells, hash_accessor);
+}
+
+Tuple Mesh::resurrect_tuple(const Tuple& tuple, Accessor<long>& hash_accessor) const
```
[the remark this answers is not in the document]
daniel-zint: I will add a comment to explain it. The name is the best I could come up with. I am open for alternatives.

src/wmtk/Accessor.cpp
```diff
@@ -113,11 +113,13 @@ auto Accessor<T, IsConst>::cacheable_scalar_attribute(const long index) -> TT
 template <typename T, bool IsConst>
 auto Accessor<T, IsConst>::vector_attribute(const Tuple& t) const -> ConstMapResult
 {
+    // assert(m_mesh.is_valid(t)); <-- done in const function
     return const_vector_attribute(t);
 }
 template <typename T, bool IsConst>
 T Accessor<T, IsConst>::scalar_attribute(const Tuple& t) const
 {
+    // assert(m_mesh.is_valid(t)); <-- done in const function
```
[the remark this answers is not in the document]
daniel-zint: But the index function does not even take a tuple so there is nothing to check for. Or did I misunderstand you?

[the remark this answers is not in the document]
mtao: the subsequent line is something like `long i = index(t);` - doesn't that take in a tuple?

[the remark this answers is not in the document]
daniel-zint: True, I moved the assert to `AccessorBase::index(t)` now.

src/wmtk/Mesh.cpp
```diff
@@ -149,10 +155,42 @@ Accessor<long> Mesh::get_cell_hash_accessor()
     return create_accessor(m_cell_hash_handle);
 }
 
+void Mesh::update_cell_hashes(const std::vector<Tuple>& cells, Accessor<long>& hash_accessor)
```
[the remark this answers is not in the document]
daniel-zint: I added `update_cell_hash`

src/wmtk/Mesh.hpp
daniel-zint: I moved the implementations to the end of the header 👍

src/wmtk/PointMesh.cpp
daniel-zint: Done. I feel like we really need a _Mesh_impl.hpp_ soon because we have plenty of tiny functions and if we do not use link time optimization that might create quite an overhead.

src/wmtk/TetMesh.cpp
daniel-zint: yes, I guess we might end up having that everywhere

src/wmtk/TriMesh.cpp
daniel-zint: We could make something like `SingleThreadMeshWrapper`.

src/wmtk/TriMesh.cpp
daniel-zint: Looks like we need to update the clang format file then. I honestly don't care about the formatting here as long as it is consistent across all machines.

src/wmtk/operations/Operation.cpp
daniel-zint: I forgot the `const`. It's there now.

PR #350, opened 2023-08-22
energy

yunfanzhou:


src/wmtk/energy/Energy.cpp
```diff
@@ -0,0 +1,9 @@
+#include "Energy.hpp"
+
+Energy::Energy(const Mesh& mesh)
+    : m_mesh(mesh)
+    , m_position_handle(m_mesh.get_attribute_handle<double>("position", PrimitiveType::Vertex)){};
```
mtao, line 5: i think the generic energy interface shouldn't be getting this handle - each energy should request these handles themselves with their constructors? but getting the mesh is fine

mtao: but the pattern for construction of these individual energies looks fine :+1:

src/wmtk/energy/AMIPS.hpp
```diff
@@ -2,6 +2,8 @@
 
 class AMIPS_2D : public wmtk::Energy
 {
+    using DScalar = DScalar2<double, Eigen::Vector2d, Eigen::Matrix2d>;
+
 public:
     double energy_eval(const Tuple& tuple) const override{};
```
mtao, line 8: code won't compile for a bit, but i'll let you figure that out :)

src/wmtk/energy/AMIPS.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    auto Fdet = F.determinant();
+    if (std::abs(Fdet) < std::numeric_limits<double>::denorm_min()) {
+        return std::numeric_limits<double>::infinity();
+    }
+    return (F.transpose() * F).trace() / Fdet;
+}
+
+static DScalar AMIPS_2D::energy_eval_autodiff(
+    const Eigen::Vector2d uv1,
```
mtao, line 67: this and hte previous should be a single templated implementation that takes in different scalar types

mtao (commented):
some comments

src/wmtk/energy/DifferentiableEnergy.hpp
```diff
@@ -1,8 +1,8 @@
-#include <wmtk/utils/autodiff.h>
+#include <wmtk/energy/utils/autodiff.h>
 #include "Energy.hpp"
 class DifferentiableEnergy : public Energy
```
mtao, line 3: all of your headers need namespaces 
```
namespace wmtk {
 class DifferentiableEnergy: public Energy {
...
 };

}
```

src/wmtk/energy/ValenceEnergy.cpp
```diff
@@ -0,0 +1,54 @@
...
+ValenceEnergy::ValenceEnergy(const TriMesh& mesh)
+    : Energy(mesh)
+{}
+
+double ValenceEnergy::energy_eval(const Tuple& tuple) const
+{
+    // assume tuple is not a boundary edge
+    const Tuple current_v = tuple;
+    const Tuple other_v = m_mesh.switch_vertex(tuple);
+    long val0 = static_cast<long>(SimplicialComplex::vertex_one_ring(mesh(), current_v).size());
```
mtao, line 15: this valence definition is very specific to collapse. The ValenceEnergy should just return the valence - this calculation for predicting the before/after only makes sense as an Invariant

PR #351, opened 2023-08-22
Hackathon/operation specialization

daniel-zint:


mtao (approved):
LGTM, this is what i was expecting.

this direction makes me worry about finalizing that style guide i think you were working on - it's now somewhat ambiguous what Operation means so it'd be great if we could convince people to always say `tri_mesh::Operation`.
This also means that we will have a rather ugly `wmtk::operations::tri_mesh::Operation` naming - but this is just an unfortunate thing that we'll have to live with

PR #352, opened 2023-08-29
Invariants

mtao:
Implementation of invariants to replace manually implemented before/after in operations.
Currently the invariant collections are stored in operation settings which is suboptimal, but the current thought is to eventually make them emerge as part of the factory pattern (which will also eventually become responsible for part of the configuration parsing as well)


components/wmtk_components/isotropic_remeshing/internal/IsotropicRemeshing.cpp
```diff
@@ -19,34 +19,37 @@ IsotropicRemeshing::IsotropicRemeshing(TriMesh& mesh, const double length, const
...
+        split_settings.split_settings.split_boundary_edges = !m_lock_boundary;
+        split_settings.initialize_invariants(m_mesh);
 
         m_scheduler.add_operation_type<tri_mesh::EdgeSplitAtMidpoint>("split", split_settings);
     }
     // collapse
     {
-        OperationSettings<tri_mesh::EdgeCollapseToMidpoint> op_settings{
-            m_position_handle,
-            m_length_min * m_length_min,
```
mtao, line 33: @daniel-zint I changed this from max to min, and this results in a failure in the unit tests, but i still believe the code to be wrong due to the choice of variable names. This is because the variable written to is `hackathon/dev` is `max_squared_length` whereas the variable used to assign is `min * min`.
https://github.com/wildmeshing/wildmeshing-toolkit/blob/b27115ad8dde64f3651af39ff8bc53902d00a2ce/src/wmtk/operations/tri_mesh/EdgeCollapseToMidpoint.hpp#L17 .
For this sort of reason I've removed as much of the bracket initialization as possible in the code - and removed default initialization of these configuration things

daniel-zint:
There is a change in perspective. The max edge length in isotropic remeshing, is the min edge length for which a split should be performed. I know this is confusing but it's the only thing that makes sense if you look at the classes outside of the context of remeshing. 

> Am 02.09.2023 - 02:19 schrieb Michael Tao:
>
>
> @mtao commented on this pull request.
> 
> In components/wmtk_components/isotropic_remeshing/internal/IsotropicRemeshing.cpp:
> 
> > m_scheduler.add_operation_type<tri_mesh::EdgeSplitAtMidpoint>("split", split_settings); } // collapse { - OperationSettings<tri_mesh::EdgeCollapseToMidpoint> op_settings{ - m_position_handle, - m_length_min * m_length_min, 
> 
> @daniel-zint I changed this from max to min, and this results in a failure in the unit tests, but i still believe the code to be wrong due to the choice of variable names. This is because the variable written to is hackathon/dev is max_squared_length whereas the variable used to assign is min * min.
> https://github.com/wildmeshing/wildmeshing-toolkit/blob/b27115ad8dde64f3651af39ff8bc53902d00a2ce/src/wmtk/operations/tri_mesh/EdgeCollapseToMidpoint.hpp#L17 .
> For this sort of reason I've removed as much of the bracket initialization as possible in the code - and removed default initialization of these configuration things
> 
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you were mentioned.
> Message ID: ***@***.***>
> 
> 

-------------------------------------------------------------------------------------------------
FreeMail powered by mail.de - MEHR SICHERHEIT, SERIOSITÄT UND KOMFORT

mtao:
Ok I'll add a comment on this

PR #355, opened 2023-09-06
Hackathon/hdf5 pearl

daniel-zint:
Just a comment in the readme that Pearl needs to be installed before calling cmake.


mtao:
perl

PR #357, opened 2023-09-10
fix IDE groups

daniel-zint:


daniel-zint:
The groups must be defined in the same file as the library. Therefore, the group for the wmtk files must be in the top level CMakeLists.txt.

mtao (approved):
sure lol

PR #362, opened 2023-09-11
throw an error if perl is not installed

daniel-zint:


mtao (approved):
lgtm - though it's sort of annoying to test i'm pretty sure this'll be good? :)

PR #369, opened 2023-09-13
Hackathon/simplex collection

daniel-zint:
This code is supposed to replace the current `SimplicialComplex`.


src/wmtk/Mesh.cpp
```diff
@@ -113,6 +113,11 @@ bool Mesh::simplex_is_equal(const Simplex& s0, const Simplex& s1) const
     return (s0.primitive_type() == s1.primitive_type()) && (id(s0) == id(s1));
 }
 
+bool Mesh::simplex_is_equal(const simplex::Simplex& s0, const simplex::Simplex& s1) const
```
mtao, line 116: perhaps `are_equal` (you have two items, so use are). I'm in general not a huge fan of the simplex class leaking into Mesh (perhaps a light utility class like `SimplexComparison{ const Mesh& m_mesh; };` would be good)?

src/wmtk/Mesh.cpp
```diff
@@ -124,6 +129,17 @@ bool Mesh::simplex_is_less(const Simplex& s0, const Simplex& s1) const
     return id(s0) < id(s1);
 }
 
+bool Mesh::simplex_is_less(const simplex::Simplex& s0, const simplex::Simplex& s1) const
```
mtao, line 132: is the previous implementation broken? `simplex::Simplex` and `Simplex` should be typedefs of one another and we should be using standard lexicographical ordering here right? Both should be something like
```
bool less([a,b], [c,d]) {
  if(a == c) {
  return b < d;
  }
return b < d;
}

File: src/wmtk/simplex/Simplex.hpp [hunk 3]
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include "../Primitive.hpp"
```
mtao, line 3: for parent folders use absolute paths `<wmtk/Primitive.hpp>`

src/wmtk/simplex/OpenStar.cpp
```diff
@@ -0,0 +1,44 @@
...
+OpenStar::OpenStar(const Mesh& mesh, const Simplex& simplex, const bool sort)
+    : SimplexCollection(mesh)
+{
+    if (!dynamic_cast<const TriMesh*>(&mesh)) {
+        throw "only testet for TriMesh";
+    }
+
+    add(simplex);
+
+    CofaceCells coface_cells = dynamic_cast<const TriMesh*>(&mesh)
```
mtao, line 23: read the `top_simplex_type` instead of these dynamic cast hacks

mtao (changes requested):
some minor comments but there are some major changes I'd change to the API:

* Currently any instance of a type derived from `SimplexCollection` (say `OpenStar`)t can be modified by calling `add` or `sort_and_clean`. These are not properties of an `OpenStar`. I do not think that `OpenStar`, `SimplexBoundary`, `CoFaceCells` etc are objects - they are operators taht act on a simplex. My _strong_ preference is to have functions with prototypes 
```
SimplexCollection open_star(const Mesh& m, const Simplex& simplex);
```
If you want to use some classes to implement them internally (such as with these iterator-like things that we had discussed previously) then those belong in  a `wmtk/simplex/detail` folder.
If you want to go down the optimization route we discussed, where some operations are implemented as coroutines/iterators)  I would also recommend you support 3 concepts of `SimplexCollection` type objects:
1. A `IterableSimplexCollection` which has a begin/end interface
2. A `SimplexCollection` which inherits from `IterableSimplexCollection but stores a std::vector<Simplex>` (or even a `std::vector<std::vector<Tuple>>`
3. A `SimplexCollectionIterator` which inherits `IterableSimplexCollection` 

* Rather than leak all of the simplex code to the Mesh class we should make a friend class `SimplicialComplex` that JUST stores a reference to a `Mesh`. I originally thought the SimplicialComplex class would be a class that could manipulate `SimplexCollection` objects and the `SimplexCollection` objects wouldn't need to hold thier own copies of the `Mesh`. If each `SimplexCollection` stores a mesh then we will need to do things like validate tuples and make sure each pair of `SimplexCollection`s that are invoked with one another utilize the same `Mesh` internally

* currently tehre are two classes called Simplex which can both be caught by ADL. Kill one of them and typedef the other.
* Your functors for less/equal should implement functionality on their own. Perhaps they hold a `SimplicialComplex` inside.

src/wmtk/Mesh.cpp
mtao: i think a new SimplicialComplex class should mediate between Mesh internals and Simplicial things - and SimplexComparison could be a friend of that?

src/wmtk/simplex/SimplexCollection.cpp
```diff
@@ -0,0 +1,92 @@
...
+{
+    std::sort(m_simplices.begin(), m_simplices.end(), m_simplex_is_less);
+    const auto last = std::unique(m_simplices.begin(), m_simplices.end(), m_simplex_is_equal);
+    m_simplices.erase(last, m_simplices.end());
+}
+
+bool SimplexCollection::contains(const Simplex& simplex) const
+{
+    // TODO this is O(n) but can and should be done in O(log n)
+    for (const Simplex& s : m_simplices) {
```
mtao, line 44: make a small pr for an enhancement using std::binary_search

src/wmtk/simplex/SimplexCollection.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include "internal/SimplexEqualFunctor.hpp"
+#include "internal/SimplexLessFunctor.hpp"
+
+namespace wmtk::simplex {
+class SimplexCollection
+{
+public:
+    SimplexCollection(const Mesh& mesh)
+        : m_mesh{mesh}
+        , m_simplex_is_less(mesh)
```
mtao, line 15: TBH this member just stores a single ref - might as well make a private member that creates the function rather than store the member (it's like storing 3 pointers to the same mesh?)

src/wmtk/simplex/SimplexCollection.hpp
```diff
@@ -0,0 +1,75 @@
...
+     * There is no sorting or any check if the vertex already exists
+     */
+    void add(const Simplex& simplex);
+
+    void add(const SimplexCollection& simplex_collection);
+
+    /**
+     * @brief Sort simplex vector and remove duplicates.
+     */
+    void sort_and_clean();
```
mtao, line 40: i think this should be private - each of the operations should by default default make sure it's cleaned/sorted?

src/wmtk/simplex/SimplexCollection.cpp
```diff
@@ -0,0 +1,92 @@
...
+    return simplices;
+}
+
+void SimplexCollection::add(const Simplex& simplex)
+{
+    m_simplices.push_back(simplex);
+}
+
+void SimplexCollection::add(const SimplexCollection& simplex_collection)
+{
```
mtao, line 29: make this guarnatee sorted/clean by using std::merge

src/wmtk/simplex/SimplexCollection.cpp
```diff
@@ -0,0 +1,92 @@
...
+            simplices.emplace_back(s);
+        }
+    }
+
+    return simplices;
+}
+
+void SimplexCollection::add(const Simplex& simplex)
+{
+    m_simplices.push_back(simplex);
```
mtao, line 25: make this insert assuming everything is sorted already (perhaps in another PR?)

src/wmtk/simplex/closed_star.hpp
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include "SimplexCollection.hpp"
+
+namespace wmtk::simplex {
+SimplexCollection
+closed_star(const Mesh& mesh, const Simplex& simplex, const bool sort_and_clean = true);
```
mtao, line 7: i'd rather just make sure that a SimplexCollection is always unique/sorted than give users this option. If you allow for users to return not-sorted-or-cleaned then some member functions of SimplexCollection do not work (in particular the union/intersection code).

src/wmtk/simplex/closed_star.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+    collection.add(simplex);
+
+    const SimplexCollection coface_cells_collection =
+        mesh.top_simplex_type() == PrimitiveType::Face
+            ? coface_cells(static_cast<const TriMesh&>(mesh), simplex, false)
+            : coface_cells(static_cast<const TetMesh&>(mesh), simplex, false);
+
+    for (const Simplex& coface_cell : coface_cells_collection.simplex_vector()) {
+        collection.add(coface_cell);
```
mtao, line 23: if we go with my thought of making add a sorted insertion (which results in O(n) insertion) it's probably better to create a `std::vector<Simplex>` to accumulate new simplices, sort it, and then merge.

src/wmtk/simplex/closed_star.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+SimplexCollection closed_star(const Mesh& mesh, const Simplex& simplex, const bool sort_and_clean)
+{
+    SimplexCollection collection(mesh);
+
+    collection.add(simplex);
+
+    const SimplexCollection coface_cells_collection =
+        mesh.top_simplex_type() == PrimitiveType::Face
+            ? coface_cells(static_cast<const TriMesh&>(mesh), simplex, false)
```
mtao, line 19: perhaps this should be a switch/case thing. soon we'll have edge/point meshes too

src/wmtk/simplex/internal/SimplexLessFunctor.hpp
```diff
@@ -0,0 +1,20 @@
...
+{
+    const Mesh& m;
+
+    SimplexLessFunctor(const Mesh& mm)
+        : m{mm}
+    {}
+
+    bool operator()(const Simplex& s0, const Simplex& s1) const
+    {
+        return m.simplex_is_less(s0, s1);
```
mtao, line 17: heh, i was teh one who asked for the change, but it's sorta ugly that this makes sense to be simplex_is_less and the other one is simplices_are_equal . In the end I think that the equal/less belong in a `struct SimplicialComplex { const Mesh& m_mesh};` datastructure. Then you'd just have `less_than` and `equal` without having to clarify simplex (+ we could remove Simplex from the Mesh class)

src/wmtk/simplex/SimplexCollection.hpp
```diff
@@ -0,0 +1,75 @@
...
+    SimplexCollection(const Mesh& mesh)
+        : m_mesh{mesh}
+        , m_simplex_is_less(mesh)
+        , m_simplex_is_equal(mesh)
+    {}
+
+    /**
+     * @brief Return const reference to the simplex vector.
+     */
+    const std::vector<Simplex>& simplex_vector() const { return m_simplices; }
```
mtao, line 22: i think this should be `as_vector()`. You definitely don't need to specify that SimplexCollection can return data as a vector.

src/wmtk/simplex/SimplexCollection.hpp
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * @brief Return vector of all simplices of the requested type.
+     */
+    std::vector<Simplex> simplex_vector(const PrimitiveType& ptype) const;
```
mtao, line 26: this could be `get_by_order` or something?

src/wmtk/simplex/open_star.cpp
```diff
@@ -0,0 +1,41 @@
...
+namespace wmtk::simplex {
+
+SimplexCollection open_star(const Mesh& mesh, const Simplex& simplex, const bool sort_and_clean)
+{
+    SimplexCollection collection(mesh);
+
+    collection.add(simplex);
+
+    const SimplexCollection coface_cells_collection =
+        mesh.top_simplex_type() == PrimitiveType::Face
```
mtao, line 18: switch/case

mtao (approved):
there's a few things i think should be changed, but lets just itemize them in a new PR?

src/wmtk/simplex/SimplexCollection.hpp
daniel-zint: not sure about the name but we can definitely change it

src/wmtk/simplex/SimplexCollection.hpp
daniel-zint: I'd like this to be public for the same reason as before. Sometimes it's better to sort things manually. An example is if you need a simplex with its boundary:
```
Simplex s;
SimplexCollection bd = simplex_boundary(mesh, s, false);
bd.add(s);
bd.sort_and_clean();
```

src/wmtk/simplex/closed_star.cpp
daniel-zint: sure, I created an issue for that

src/wmtk/simplex/closed_star.cpp
daniel-zint: Once more, I am not sure if we want to keep things sorted at all times. We need to discuss it.

src/wmtk/simplex/internal/SimplexLessFunctor.hpp
daniel-zint: You mentioned that before and I have not fully understood your point here so far. For simplex comparison, we need functors in some way. I don't see how this can be done within one class for both, `less_than` and `equal`. I guess I am missing some knowledge here.

daniel-zint (commented):
Comments to the comments. The main discussion is, if we want to keep `SimplexCollection` sorted at all times. I think it might have a significant performance impact, if we can delay sorting in certain scenarios.

src/wmtk/simplex/SimplexCollection.hpp
mtao: as_vector is the convention from the ranges library, which is a bit more clear imo?

PR #379, opened 2023-09-22
Mtao/update release name

mtao:
adding a comment to indicate that the old version of the library can be accessed via a tag


mtao:
forgot that i have bypass permission so i'd already updated hackathon/dev

PR #386, opened 2023-09-27
Edge mesh input

Zhouyuan-Chen:
the first edge_mesh_input pr


Zhouyuan-Chen:
I wrote a simple .obj EdgeMeshReader. can only support edges and vertices input.

Zhouyuan-Chen:
I debugged the .obj file reading code. I wrote a sample .obj file under data/ called edge_test.obj please check if the format I read is correct.

src/wmtk/io/EdgeMeshReader.cpp
```diff
@@ -0,0 +1,86 @@
...
+void EdgeMeshReader::read(Eigen::Matrix<long, -1, -1>& E, Eigen::MatrixXd& V)
+{
+    std::vector<std::pair<long, long>> edges;
+    std::vector<std::vector<double>> vertices;
+    char* line_buffer = nullptr;
+    std::fstream f(m_filename);
+    if (!f) {
+        throw std::runtime_error("can't open file!");
+    }
+    if (m_type == OBJ) {
```
daniel-zint, line 18: use `switch` here instead of `if`

src/wmtk/io/EdgeMeshReader.cpp
```diff
@@ -0,0 +1,86 @@
...
+        std::string buffer;
```
daniel-zint, line 19: There is a lot of code. You can push that into a local function to keep `read()` clean.

src/wmtk/io/EdgeMeshReader.cpp
```diff
@@ -0,0 +1,86 @@
...
+EdgeMeshReader::EdgeMeshReader(const std::string& filename, const file_type type)
+    : m_filename(filename)
+    , m_type(type)
+{}
+void EdgeMeshReader::read(Eigen::Matrix<long, -1, -1>& E, Eigen::MatrixXd& V)
+{
+    std::vector<std::pair<long, long>> edges;
+    std::vector<std::vector<double>> vertices;
+    char* line_buffer = nullptr;
+    std::fstream f(m_filename);
```
daniel-zint, line 14: use `std::ifstream`

src/wmtk/io/EdgeMeshReader.cpp
```diff
@@ -0,0 +1,86 @@
...
+    if (!f) {
```
daniel-zint, line 15: `f.is_open()` might do the same thing (I am not sure) but it for sure makes your intend more clear.

src/wmtk/io/EdgeMeshReader.cpp
```diff
@@ -0,0 +1,86 @@
...
+        throw std::runtime_error("can't open file!");
+    }
+    if (m_type == OBJ) {
+        std::string buffer;
+        while (!f.eof()) {
+            getline(f, buffer);
+            // here is for debugging
+            // spdlog::info("lineinfo: {}", buffer[0]);
+            if (buffer[0] != 'v' && buffer[0] != 'l') continue;
```
daniel-zint, line 24: No one line if-statements. Put brackets around `continue;`

src/wmtk/io/EdgeMeshReader.cpp
```diff
@@ -0,0 +1,86 @@
...
+            // read obj
+            line_buffer = new char[buffer.size()];
+            strcpy(line_buffer, buffer.c_str());
```
daniel-zint, line 27: You can use a `std::string` just like an array. There is no need to copy the string into a char array.
For example, you can also just write `buffer[0] == 'v'`.

src/wmtk/io/EdgeMeshReader.cpp
```diff
@@ -0,0 +1,86 @@
...
+            if (line_buffer[0] == 'v') {
+                std::vector<double> positions;
+                char* token = strtok(line_buffer, " ");
```
daniel-zint, line 30: Use a `std::istringstream` instead.

daniel-zint (commented):
Only reviewing the `EdgeMeshReader`. While the concept seems to be fine, it is quite C heavy. Please try to use C++ `std` library functions instead.

Zhouyuan-Chen:
this part code should be done if we only need one kind of file format.
wait to integrate the edgemesh part code into input.cpp.

PR #397, opened 2023-09-28
Embedding component

Zhouyuan-Chen:
open a new PR, directly to the wildmeshing toolkit.


Zhouyuan-Chen:
the user settings have not been added yet.

Zhouyuan-Chen:
added user settings

Zhouyuan-Chen:
I integrate the edge_input and the embedding code. still have little problems with the ".obj" files reading. Also, I fixed a bug in EdgeMeshReader.

Zhouyuan-Chen:
this part code is almost done. wait to integrate the EdgeMesh code.
For now, the output is TriMesh with tags. 
it should be handy to be used in the extract component.
I will update the extract component soon.

PR #407, opened 2023-09-30
406 coface cells refactor

mtao:
Adding a guarantee / checks that our coface-cells function matches new requitements from issue 406.
Also renamed to top_level_cofaces to identify taht it generates the top level cofaces of an input simplex (which is hopefully a bit more clear than coface_cells).
Added comparison operators to PrimtiiveType and Tuple to allow for sorting, as well as a specialized TopLevelCell sorter which si primarily of use just for making sure we haev unique top level simplices. Using the cell id in the Tuple lets us skip using the mesh to identify unique simplices in this particular function.

(sorry for slight wipe of some git history - a bit too tired to cherry pick properly / being lazy :-1: )


src/wmtk/SimplicialComplex.cpp
```diff
@@ -2,7 +2,7 @@
 
 #include "simplex/SimplexCollection.hpp"
 #include "simplex/closed_star.hpp"
-#include "simplex/coface_cells.hpp"
+#include "simplex/top_level_cofaces.hpp"
```
daniel-zint, line 5: Why did you replace `coface_cells` by `top_level_cofaces`?

src/wmtk/Tuple.cpp
```diff
@@ -36,15 +36,22 @@ bool Tuple::operator==(const wmtk::Tuple& t) const
     return std::tie(m_local_vid, m_local_eid, m_local_fid, m_global_cid, m_hash) ==
            std::tie(t.m_local_vid, t.m_local_eid, t.m_local_fid, t.m_global_cid, t.m_hash);
 }
+
+bool Tuple::operator<(const wmtk::Tuple& t) const
```
daniel-zint, line 40: it was about time that we get this one

tests/test_simplex_collection.cpp
```diff
@@ -218,93 +230,112 @@ TEST_CASE("simplex_coface_cells", "[simplex_collection][2D]")
...
     }
 }
 
-TEST_CASE("simplex_coface_cells_iterable", "[simplex_collection][2D]")
+TEST_CASE("simplex_top_level_cofaces_iterable", "[simplex_collection][2D]")
 {
     tests::DEBUG_TriMesh m = tests::hex_plus_two();
 
-    std::unique_ptr<Simplex> ptr_simplex;
+    Simplex simplex(PrimitiveType::Vertex, {});
```
daniel-zint, line 309: You can use `Simplex::vertex` here.

daniel-zint (approved):
Should be fine. I just don't understand the renaming.

src/wmtk/SimplicialComplex.cpp
mtao: i didn't find the name very descriptive/direct. right now the name is "the cells that are cofaces". instead the name now implies "the cofaces that are at the top level". the specialization of which cofaces we are looking at seems more apt than the specialization of which cells we are looking at (i've had trouble remembering this definition at least?)

daniel-zint: It's fine for me. I understand both.

PR #410, opened 2023-09-30
`PolygonMesh` implementation

rjc8237:
Adding a potential interface for a PolyMesh class (derived from Mesh) based on TriMesh.

Major design decisions are explained in the initial comment in `PolyMesh.h`.

### DIFF
--- a/src/wmtk/CMakeLists.txt
+++ b/src/wmtk/CMakeLists.txt
@@ -5,10 +5,14 @@ set(SRC_FILES
     Mesh.hpp
     PointMesh.cpp
     PointMesh.hpp
+    EdgeMesh.hpp
+    EdgeMesh.cpp
     TriMesh.cpp
     TriMesh.hpp
     TetMesh.cpp
     TetMesh.hpp
+    EdgeMeshOperationExecutor.hpp
+    EdgeMeshOperationExecutor.cpp
     TriMeshOperationExecutor.hpp
     TriMeshOperationExecutor.cpp
     TetMeshOperationExecutor.hpp
--- a/src/wmtk/EdgeMesh.cpp
+++ b/src/wmtk/EdgeMesh.cpp
@@ -1,0 +1,262 @@
+#include "EdgeMesh.hpp"
+
+
+#include <wmtk/utils/edgemesh_topology_initialization.h>
+#include <wmtk/EdgeMeshOperationExecutor.hpp>
+#include <wmtk/SimplicialComplex.hpp>
+#include <wmtk/utils/Logger.hpp>
+namespace wmtk {
+EdgeMesh::EdgeMesh()
+    : Mesh(1)
+    , m_ve_handle(register_attribute<long>("m_ve", PrimitiveType::Vertex, 1))
+    , m_ev_handle(register_attribute<long>("m_ev", PrimitiveType::Edge, 2))
+    , m_ee_handle(register_attribute<long>("m_ee", PrimitiveType::Edge, 2))
+{}
+EdgeMesh::EdgeMesh(const EdgeMesh& o) = default;
+EdgeMesh::EdgeMesh(EdgeMesh&& o) = default;
+EdgeMesh& EdgeMesh::operator=(const EdgeMesh& o) = default;
+EdgeMesh& EdgeMesh::operator=(EdgeMesh&& o) = default;
+
+Tuple EdgeMesh::split_edge(const Tuple& t, Accessor<long>& hash_accessor)
+{
+    EdgeMesh::EdgeMeshOperationExecutor executor(*this, t, hash_accessor);
+    return executor.split_edge();
+}
+
+Tuple EdgeMesh::collapse_edge(const Tuple& t, Accessor<long>& hash_accessor)
+{
+    EdgeMesh::EdgeMeshOperationExecutor executor(*this, t, hash_accessor);
+    return executor.collapse_edge();
+}
+
+long EdgeMesh::id(const Tuple& tuple, PrimitiveType type) const
+{
+    switch (type) {
+    case PrimitiveType::Vertex: {
+        ConstAccessor<long> ev_accessor = create_const_accessor<long>(m_ev_handle);
+        auto ev = ev_accessor.vector_attribute(tuple);
+        return ev(tuple.m_local_vid);
+    }
+    case PrimitiveType::Edge: {
+        return tuple.m_global_cid;
+    }
+    case PrimitiveType::Face:
+    case PrimitiveType::Tetrahedron:
+    default: throw std::runtime_error("Tuple id: Invalid primitive type");
+    }
+}
+
+bool EdgeMesh::is_boundary(const Tuple& tuple) const
+{
+    return is_boundary_vertex(tuple);
+}
+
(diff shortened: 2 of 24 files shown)


deniszorin:
The generic polygon operations (split/join vertex/face)  are not included so far  - but more complex compound operations collapse/split are included so I am not sure how these would be implemented.  I think we should follow the writeup exactly as described unless there is a very good reason not to  (in which case we need to rewrite it and verify pre/post for new formulations  of compound operations). 

This data structure seems unnecessarily heavy and  so far I  do not understand why - we need to discuss more.   Instead of  5 arrays next, prev,, tip, fh, vh (assuming opp is implicit) this has 10.  I am not entirely confident that implicit opp is a good idea (I recall we had to change it in the conformal code - initially it was, but having an explicit opp simplified something).   I am guessing 0 and 1 refer to orientation -- this goes against what we've discussed perhaps we did not understand each other (and I still do not see anything written down explaining the need to store the orientations).   Orientation bits for tuple representation are needed, but can be generated on the fly.  I still do not see how storing these reduces code complexity, or is needed for navigation,  while having multiple redundant arrays stored both increases memory and hurts performance. 

Suppose a tuple is t = (h,or) where or = 0 or 1  denote   reverse(t)  = (h, not(or))  
    attribute::MeshAttributeHandle<long> m_hn0_handle;  // next
     attribute::MeshAttributeHandle<long> m_hn1_handle;  // unnecessary  next(t)  = prev(reverse(t))
     attribute::MeshAttributeHandle<long> m_hp0_handle; // prev
     attribute::MeshAttributeHandle<long> m_hp1_handle; // unnecessary  prev(t) = next(reverse(t))

     attribute::MeshAttributeHandle<long> m_hv0_handle;  // tip 
     attribute::MeshAttributeHandle<long> m_hv1_handle;  //unnecessary  tip(t) = tip(opp(reverse(t))
     attribute::MeshAttributeHandle<long> m_hf0_handle;   //  face
     attribute::MeshAttributeHandle<long> m_hf1_handle;   // unnecessary face(t) = face(reverse(t))

     attribute::MeshAttributeHandle<long> m_fh_handle;
     attribute::MeshAttributeHandle<long> m_vh_handle;

rjc8237:
I agree about the generic polygon operations. I only included the traversal operations and the operations that are inherited from the base class in the current interface.

There are 10 arrays instead of 5 as the proposed interface stores halfedgre attributes as pairs of edge attributes, so they are all doubled. 0 and 1 refers to the first and second halfedges of the edge, not the orientation. There are 8 longs per edge (or 4 longs per halfedge). I don't believe any of these can be compressed further for general polygons.

deniszorin:
> I agree about the generic polygon operations. I only included the traversal operations and the operations that are inherited from the base class in the current interface.
> 
> There are 10 arrays instead of 5 as the proposed interface stores halfedgre attributes as pairs of edge attributes, so they are all doubled. 0 and 1 refers to the first and second halfedges of the edge, not the orientation. There are 8 longs per edge (or 4 longs per halfedge). I don't believe any of these can be compressed further for general polygons.

Ok - as we discussed on slack we are in agreement, modulo the details of how per-half edge storage is implemented  with edge attributes. 

I'd prefer  to define the interface for these in the tuple form the start to avoid any surprises.  I do not expect any as far as function arguments/returns are concerned -  the map from tuple to halfedge is unique,  and if we return a tuple instead of a half edge, we can always return the one with the same orientation as the half edge.  Still no harm in having a more complete API defined from the start, and make the conventions explicit.

rjc8237:
I added the polygon mesh modification operations. I also made the halfedge arrays |E|x2 edge attributes, and I added a boolean face attribute for holes.

I have two questions about the interface and data management:
1)  What should the handles for vertices and faces be? Just longs? How should they be generated and deleted?
2) For halfedge operations such as splice, should the orientation bit be ignored and a default ccw orientation assumed, or should consistent orientation be enforced as a precondition? This is particularly relevant for operations like split face, where the new edge would be made between the vertices at the ccw tip of the two input halfedges and *not* the vertices indicated by the local vid index.

src/wmtk/PolyMesh.hpp
```diff
@@ -0,0 +1,264 @@
...
+     * @brief Primitive for changing connectivity without changing the number of edges.
+     * 
+     * Requires t and s to have consistent orientation
+     */
+    Tuple splice(const Tuple& t, const Tuple& s);
+
+    /**
+     * @brief Primitive for adding a bubble component with a single edge.
+     */
+    Tuple make_edge();
```
mtao, line 110: i think you're mimicing the split_edge / collapse_edge operation stuff here - and to be honest I do not think that it's a great idea.

The original intent from way back when was that the class implements the operation and returns a reference `Tuple` to help perform navigation for other tasks. The issue that arose is that each operation's return Tuple had a different semantic and so every consumer of these return Tuple had to keep track of which operation the return data came from and how to do navigation with respect to that tuple.

IMO these operations will/should provide per-operation classes to help identify information about what happened (which tuple was deleted? what simplices got modified? etc)

Furthermore these operations really shouldn't be called by anyone other than some Operation object - and once they only really get invoked by an Operation object then the implementation might as well be shoved to those objects instead of being left within the mesh struct (in theory these functions are all protected if you keep htem in the Mesh).

src/wmtk/PolyMesh.hpp
```diff
@@ -0,0 +1,264 @@
...
+protected:
+    long id(const Tuple& tuple, PrimitiveType type) const override;
+
+private:
+    /**
+     * @brief Generate a new vertex handle
+     * 
+     * TODO: Check if should be implemented using request_simplex_indices
+     */
+    long new_vertex();
```
mtao, line 207: in principle we never want to expose interfaces that use indices - even "internally"

mtao (commented):
some comments for now

src/wmtk/PolyMesh.hpp
```diff
@@ -0,0 +1,264 @@
...
+    Tuple make_hole(const Tuple& t);
+
+    /**
+     * @brief Mark a face as not a hole in the surface.
+     */
+    Tuple fill_hole(const Tuple& t);
+
+    // TODO These only make sense for triangle meshes, but we do want them in the interface
+    // since they can be implemented using splice and bubble operations
+    Tuple split_edge(const Tuple& t, Accessor<long>& hash_accessor) override;
```
mtao, line 160: these will be implemented as an Operation struct that holds several other Operation structs internally.

src/wmtk/PolyMesh.hpp
rjc8237: So should all mesh operations (splice, make_edge, split_face, joint_vertex, etc.) be implemented instead as Operations akin to those in `operations/tri_mesh`?

mtao: Being more akin to the `opeartoins/tri_mesh` is my personal preference as it leaves the PolygonMesh structure itself fairly simple, but we can discuss this with others. Right now in TriMesh/TetMesh we have these TriMeshOperationExecutor objects which really are the Operation implementations but in a weird place? If you prefer constructing the execution structures and returning the data from those objects that's fine as well - but having all of this data in an Operation object would have been easier imo. These Executor classes were in part developed because the Operation infrastructure wasn't there yet and should potentially be refactored.

PR #414, opened 2023-10-02
autogen headers can now have comments + do have some

mtao:
just commenting the autogen to help keep track of what the various variables do


mtao:
primarily just added comments so i force pushed

PR #426, opened 2023-10-05
 Component `extract_subset`

### LINKED ISSUE #403: Component `extract_subset`
Extract a subset from a mesh that is represented by a tag attribute (of type `long`)

Possible options:
```
struct ExtractSubsetOptions
{
    std::string type;
    std::string input;
    std::string output;
    std::string tag_name;
    long tag_value;
    long embedding_cell_dimension;
    long subset_cell_dimension;
};
```

ahighmoon:
Create a component: to extract a subset/submesh of a given mesh according to a given tag


daniel-zint:
this PR should be a draft

components/wmtk_components/extract_subset/internal/extract_subset_2d.cpp
```diff
@@ -62,23 +62,34 @@ wmtk::TriMesh extract_subset_2d(
...
-//         for (size_t k2 = 0; k2 < 3; ++k2) {
-//             // faces_in(k, k2) = triangles(tag[k], k2);
-//             vertices_in_bool[triangles(tag[k], k2)] = true;
-//         }
-//     }
+    int nb_vertex = vertices.size();
+    int nb_vertex_in = 0;
+    int nb_tri_in = tag.size();
+    std::vector<bool> vertices_in_bool(nb_vertex);
+    for (int k = 0; k < nb_vertex; ++k) {vertices_in_bool[k] = false;}
```
daniel-zint, line 73: It looks like you are not using an auto formatter. We have a .clang-format file in our repository. Please install an auto formatter and use this format file. I recommend also to activate "format on save". That way it is sure that you never forget formatting before you commit.

components/wmtk_components/extract_subset/internal/extract_subset_2d.cpp
```diff
@@ -62,23 +62,34 @@ wmtk::TriMesh extract_subset_2d(
 
 
 // Note: the above is a draft version of the algo, implemented in bad, drafted data structure
-// The following is new code to be finished, commented out to compile
+// The following is new code to be finished
 wmtk::TriMesh extract_subset_2d(std::vector<wmtk::Tuple> vertices, std::vector<wmtk::Tuple> triangles, std::vector<size_t> tag){
```
daniel-zint, line 66: I don't really understand the parameters here. Why `vertices` and `triangles`? I think you should have a `Mesh` as input in combination with an `Attribute` that is used as tag. 

Minor comment:
The `return` value should also just be a `Mesh`. This code can be written in a generic enough way so that it works for extracting triangles or just a bunch of edges. However, we can do the generalization as soon as the basic concept is there, so don't worry about that for now.

components/wmtk_components/extract_subset/extract_subset.cpp
```diff
@@ -0,0 +1,22 @@
+#include "extract_subset.hpp"
+
+
+namespace wmtk{
+namespace components {
+wmtk::TriMesh extract_subset(long dimension, const wmtk::TriMesh& m, std::vector<size_t> tag){
```
daniel-zint, line 6: Change the order of parameters: `mesh, tag, dimension`

components/wmtk_components/extract_subset/extract_subset.cpp
```diff
@@ -0,0 +1,22 @@
...
+    switch (dimension){
+        case 2: {
+            // return internal::extrace_subset_2d();
+            std::vector<Tuple> vertices = m.get_all(wmtk::PrimitiveType::Vertex);
+            std::vector<Tuple> triangles = m.get_all(wmtk::PrimitiveType::Face);
+            return internal::extract_subset_2d(vertices, triangles, tag);
```
daniel-zint, line 12: A `Tuple` without it's corresponding mesh is useless. The tuple is a handle that can only be evaluated in combination with its mesh.

components/wmtk_components/extract_subset/extract_subset.cpp
```diff
@@ -0,0 +1,22 @@
...
+        }
+        case 3: {
```
daniel-zint, line 14: Add a `default` case that contains `throw "not implemented";` or something similar.

components/wmtk_components/extract_subset/internal/extract_subset_2d.cpp
```diff
@@ -0,0 +1,130 @@
+#include "extract_subset_2d.hpp"
+
+wmtk::TriMesh extract_subset_2d(
+    const std::vector<Eigen::Vector2d>& points,
+    Eigen::MatrixXi& triangles, std::vector<size_t> tag){
+
+    int nb_vertex = points.size();
```
daniel-zint, line 7: Should be `const` and it should be a `size_t` just like `points.size()`. If you want it to be an `int` for whatever reason, you should convert it explicitly because `size_t` to `int` is a narrowing conversion and it will produce a warning.

components/wmtk_components/extract_subset/internal/extract_subset_2d.cpp
```diff
@@ -0,0 +1,130 @@
...
+    int nb_vertex_in = 0;
+    int nb_tri_in = tag.size();
+    // maintain a vector of bool, true if an old vertex is preserved after extraction
+    std::vector<bool> vertices_in_bool(nb_vertex);
```
daniel-zint, line 11: You can set the values directly in the constructor
```
std::vector<bool> vertices_in_bool(nb_vertex, false);
```

components/wmtk_components/extract_subset/internal/extract_subset_2d.cpp
```diff
@@ -0,0 +1,130 @@
+#include "extract_subset_2d.hpp"
+
+wmtk::TriMesh extract_subset_2d(
+    const std::vector<Eigen::Vector2d>& points,
+    Eigen::MatrixXi& triangles, std::vector<size_t> tag){
+
+    int nb_vertex = points.size();
+    int nb_vertex_in = 0;
+    int nb_tri_in = tag.size();
```
daniel-zint, line 9: Same es for `nb_vertex`

components/wmtk_components/extract_subset/internal/extract_subset_2d.cpp
```diff
@@ -0,0 +1,130 @@
...
+    // maintain a vector of bool, true if an old vertex is preserved after extraction
+    std::vector<bool> vertices_in_bool(nb_vertex);
+    for (int k = 0; k < nb_vertex; ++k) {vertices_in_bool[k] = false;}
+
+    Eigen::MatrixXi faces_in;
+    faces_in.resize(nb_tri_in, 3);
+
+    //tag the preserved ones and count number of vertex in new extraction
+    for (size_t k = 0; k < nb_tri_in; ++k){
+        for (size_t k2 = 0; k2 < 3; ++k2) {
```
daniel-zint, line 19: I don't know why you are using `k` and `k2` here but the convention is to use `i` and `j`. Please change that, otherwise it might be confusing.

components/wmtk_components/extract_subset/internal/extract_subset_2d.cpp
```diff
@@ -0,0 +1,130 @@
...
+            vertices_in_bool[triangles(tag[k], k2)] = true;
+        }
+    }
+    for (bool b: vertices_in_bool) {
+        if (b) nb_vertex_in ++;
+    }
+
+    // construct a map from old vertex id to new new id
+    std::map<int, int> old2new;
+    int j = 0;
```
daniel-zint, line 30: `j` is more than a counter variable here, give it a more descriptive name.

components/wmtk_components/extract_subset/internal/extract_subset_2d.cpp
```diff
@@ -0,0 +1,130 @@
...
+            old2new.insert({i, j});
+            j++;
+        }
+    }
+    assert(j == nb_vertex_in);
+
+    wmtk::TriMesh mesh;
+    wmtk::RowVectors3l tris;
+    tris.resize(nb_tri_in, 3);
+    for (unsigned int i = 0; i < nb_tri_in; ++i){
```
daniel-zint, line 44: should also be `size_t`

components/wmtk_components/extract_subset/internal/extract_subset_2d.cpp
```diff
@@ -0,0 +1,130 @@
...
+        size_t tri = tag[i];
+        tris.row(i) << old2new[triangles(tri, 0)], old2new[triangles(tri, 1)], old2new[triangles(tri, 2)];
+    }
+    mesh.initialize(tris);
+
+    Eigen::MatrixXd points_in;
+    points_in.resize(nb_vertex_in, 2);
+    for (int i = 0; i < nb_vertex; ++i){
+        if (vertices_in_bool[i]){
+            points_in(old2new[i], 0) = points[i][0];
```
daniel-zint, line 55: `points_in.row(i) = points[i]`

tests/components/test_component_extract_subset.cpp
```diff
@@ -0,0 +1,8 @@
+#include <wmtk_components/extract_subset/internal/extract_subset_2d.hpp>
+#include <catch2/catch_test_macros.hpp>
+
+
+TEST_CASE("manual test case", "[components][extract_subset][2D]")
```
daniel-zint, line 5: Test cases should not have spaces in their name, use `_` instead: `manual_test_case`.

Furthermore, test case names should be very descriptive because all test cases share the same scope.

daniel-zint (changes requested):
I think the general idea of the algorithm itself is there but there are some issues on the coding side. Every method will get a `Mesh` as input and will return a `Mesh` as output. You are currently still using the IGL style matrices. They should be avoided if possible.

Oh and please set this PR as "draft".

components/wmtk_components/extract_subset/extract_subset.cpp
mtao: Btw if you throw a string literal it causes weird outputs. Make sure to throw a runtime error

mtao:
What is the progress here? Given the new multimesh structure there's extra things to develop in this concept of extracting a subset. There are others working on the extraction of a subset with codimension so it'd be good to unify the API for both/make sure both implementations have the appropriate similarities

tests/components/test_component_extract_subset.cpp
```diff
@@ -1,8 +1,25 @@
...
 #include <catch2/catch_test_macros.hpp>
+#include "../tools/TriMesh_examples.hpp"
+#include "wmtk/attribute/Attribute.hpp"
+#include "wmtk/Primitive.hpp"
+#include "wmtk/attribute/AttributeHandle.hpp"
+#include "wmtk/Mesh.hpp"
 
-
-TEST_CASE("manual test case", "[components][extract_subset][2D]")
+TEST_CASE("3 neighbors test case", "[components][extract_subset][2D]")
```
daniel-zint, line 10: write test case names with underscores: "3_neighbors_test_case"

tests/components/test_component_extract_subset.cpp
```diff
@@ -1,8 +1,25 @@
...
+    V.resize(tm.capacity( wmtk::PrimitiveType::Face), 1);
+    V.row(0) << 1;
+    V.row(1) << 0; // ont tri is not tagged
+    V.row(2) << 1;
+    V.row(3) << 1;
+    auto tag_handle = wmtk::mesh_utils::set_matrix_attribute(V, "tag", wmtk::PrimitiveType::Face, tm);
     
+    // NOTE: I don't know why the following line gives me a linking error of undefined reference...
+    // wmtk::TriMesh new_tm = wmtk::components::extract_subset(tm, tag_handle, 2);
+    wmtk::TriMesh new_tm = wmtk::components::internal::extract_subset_2d(tm, tag_handle);
```
daniel-zint, line 24: If I see it correctly, you didn't set `extract_subset_2d` in the `internal` namespace. That is why it cannot find it.

tests/components/test_component_extract_subset.cpp
```diff
@@ -1,8 +1,25 @@
...
 {
+    wmtk::TriMesh tm;
+    tm = wmtk::tests::three_neighbors(); // start with 4 tri
+    Eigen::Vector<long, 10> V; // init the data vector with any static length, then resize
+    V.resize(tm.capacity( wmtk::PrimitiveType::Face), 1);
+    V.row(0) << 1;
+    V.row(1) << 0; // ont tri is not tagged
+    V.row(2) << 1;
+    V.row(3) << 1;
+    auto tag_handle = wmtk::mesh_utils::set_matrix_attribute(V, "tag", wmtk::PrimitiveType::Face, tm);
```
daniel-zint, line 20: That way of setting values should be only used if you want to assign a value to each vertex at once. It is mostly for the case that you load data from a file. In general, you should follow this example:
https://github.com/wildmeshing/wildmeshing-toolkit/blob/764ca0e341b9ec83f3592dc1f92aebf40caeb6c6/tests/test_accessor.cpp#L54

`register_attribute` can also take a default value for its attribute:
https://github.com/wildmeshing/wildmeshing-toolkit/blob/764ca0e341b9ec83f3592dc1f92aebf40caeb6c6/src/wmtk/Mesh.hpp#L97
You might want to use that to default everything to something signaling invalidness like `-1`.

daniel-zint (commented):
Please take a look at the code I referenced for the usage of attributes.

tests/components/test_component_extract_subset.cpp
ahighmoon: According to the commit history https://github.com/wildmeshing/wildmeshing-toolkit/commit/f1b7140cefd57b767dd338aa4242778ab1c2b8c4, this functionality is update on Oct 5. So in our current branch there is no such interface right now.

ahighmoon: I did a merge just now.

mtao:
this task seems quite similar to https://github.com/wildmeshing/wildmeshing-toolkit/pull/502
@daniel-zint do you want to take a look and determine how these two PRs should interact?

PR #432, opened 2023-10-10
add counter to `Scheduler`

daniel-zint:


src/wmtk/Scheduler.cpp
```diff
@@ -17,6 +17,8 @@ void Scheduler::run_operation_on_all(PrimitiveType type, const std::string& name
     // run();
```
mtao, line 17: success/fail need to be reset here in the case that the scheduler gets called multiple times (this is the main reason we want this - to call teh same scheduler multiple times until all operations fail). A unit test that catches this should exist too

mtao (changes requested):
almost there, just need to reset the counters for every operation i tihnk

mtao (approved):
lgtm

PR #433, opened 2023-10-10
Check that input and output vertices for `Delaunay` are the same

daniel-zint:
Just adding a few `CHECK`s to the `Delaunay` tests.

### DIFF
--- a/tests/components/test_component_delaunay.cpp
+++ b/tests/components/test_component_delaunay.cpp
@@ -44,7 +44,7 @@ void check_p_is_contained_in_v(
     }
 }
 
-TEST_CASE("component_delaunay", "[components][delaunay][.]")
+TEST_CASE("component_delaunay", "[components][delaunay]")
 {
     std::map<std::string, std::filesystem::path> files;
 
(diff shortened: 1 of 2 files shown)


tests/components/test_component_delaunay.cpp
```diff
@@ -82,14 +83,15 @@ TEST_CASE("delaunay_2d_random", "[components][delaunay]")
     Eigen::MatrixXd vertices;
     Eigen::MatrixXi faces;
```
mtao, line 84: for future reference, could have used structured bindings for these to clean the notation up

mtao (approved):
lgtm

daniel-zint:
I just realized that the extra check in 3d might cause a bug because in 3d there might be Steiner points added. I need to fix that before merging.

tests/components/test_component_delaunay.cpp
daniel-zint: I did, didn't I? 
I just couldn't use the `auto [vertices, faces]` notation because the variable declaration would have been wrapped inside the `CHECK`.

PR #437, opened 2023-10-11
Add new component Regular Space

Zhouyuan-Chen:
Just a dummy skeleton now.


Zhouyuan-Chen:
I wrote the basic operations, SplitFace, SplitFaceWithTag and SplitEdgeWithTag. I would rename the last two classes as SplitFaceWithTodoList and SplitEdgeWithTodoList.

Zhouyuan-Chen:
I found an interesting "bug" in the DEBUG_TriMesh. That is when you collapse an edge, it seems the maximum of the indices will not be modified. As a result, when you only have 4 vertices, you still have a vertex with idx = 4.

Zhouyuan-Chen:
I will first test the 1d case since the 2d case gets involved in the split result's direction. Something should be more clear about the result of the split operation in the document.

daniel-zint:
> I found an interesting "bug" in the DEBUG_TriMesh. That is when you collapse an edge, it seems the maximum of the indices will not be modified. As a result, when you only have 4 vertices, you still have a vertex with idx = 4.

This is not a bug. The mesh class does not yet have a garbage collection.

src/wmtk/operations/tri_mesh/EdgeSplitWithTag.cpp
```diff
@@ -0,0 +1,98 @@
...
+    long vt0 = m_vertex_tag_accessor.scalar_attribute(input_tuple());
+    long vt1 = m_vertex_tag_accessor.scalar_attribute(mesh().switch_vertex(input_tuple()));
+    {
+        EdgeSplit split_op(mesh(), input_tuple(), m_settings.split_settings);
+        if (!split_op()) {
+            return false;
+        }
+        m_output_tuple = split_op.return_tuple();
+    }
+    m_pos_accessor.vector_attribute(m_output_tuple) = 0.5 * (p0 + p1);
```
daniel-zint, line 61: You can make this a compound operation using `EdgeSplitAtMidpoint`.

src/wmtk/operations/tri_mesh/EdgeSplitWithTag.hpp
```diff
@@ -0,0 +1,63 @@
...
+    double min_squared_length = -1;
+
+    void initialize_invariants(const TriMesh& m);
+
+    // debug functionality to make sure operations are constructed properly
+    bool are_invariants_initialized() const;
+};
+
+namespace tri_mesh {
+class EdgeSplitWithTag : public TriMeshOperation, private TupleOperation
```
daniel-zint, line 37: Also here, please explain what this operation is doing. 
We have little to no documentation in our code and this is an issue we need to tackle.

src/wmtk/operations/tri_mesh/EdgeSplitWithTag.cpp
```diff
@@ -0,0 +1,98 @@
...
+{}
+std::string EdgeSplitWithTag::name() const
+{
+    return "edge_split_with_tag";
+}
+Tuple EdgeSplitWithTag::return_tuple() const
+{
+    return m_output_tuple;
+}
+bool EdgeSplitWithTag::before() const
```
daniel-zint, line 43: You don't need to override `before()`. It is already defined in the base class. Just remove `before` completely from your class.

src/wmtk/operations/tri_mesh/FaceSplit.hpp
```diff
@@ -0,0 +1,53 @@
...
+    // this operation never set the position of the new vertex.
+    //     / | \
+    //    /  |  \
+    //   /  _*_  \
+    //  / _< f \_ \
+    //  |/_______\|
+    //   \       /
+    //    \     /
+    //     \   /
+    Tuple return_tuple() const;
```
daniel-zint, line 39: Very well documented, just change the format to doxygen style:
```
/**
* The return tuple is the new vertex, pointing to the original vertex.
* This operation does not set vertex positions.  <-- this belongs to the explanation of the class
* 
*           / | \
*          /  |  \
*        /  _*_  \
*      / _< f \_ \
*      |/_______\|
*       \          /
*         \      /
*           \  /
*/
```
Something like that. Sorry, I messed up the image 😅

src/wmtk/operations/tri_mesh/FaceSplitWithTag.hpp
```diff
@@ -0,0 +1,59 @@
...
+    Tuple m_output_tuple;
+
+    Accessor<double> m_pos_accessor;
+    Accessor<long> m_vertex_tag_accessor;
+    Accessor<long> m_edge_tag_accessor;
+    Accessor<long> m_split_todo_accessor;
+
+    const OperationSettings<FaceSplitWithTag>& m_settings;
+
+    Eigen::Vector3d p0;
```
daniel-zint, line 53: Please do not use the class as a "local storage". These variables should be defined in `execute`.

src/wmtk/operations/tri_mesh/FaceSplitWithTag.cpp
```diff
@@ -0,0 +1,92 @@
...
+    p2 = m_pos_accessor.vector_attribute(mesh().switch_vertex(mesh().switch_edge(input_tuple())));
+    {
+        FaceSplit split_op(mesh(), input_tuple(), m_settings.face_split_settings);
+        if (!split_op()) {
+            return false;
+        }
+        m_output_tuple = split_op.return_tuple();
+    }
+
+    m_pos_accessor.vector_attribute(m_output_tuple) = (p0 + p1 + p2) / 3.0;
```
daniel-zint, line 64: Just like for the `EdgeSplitWithTag` it would be cleaner if you would have a `FaceSplitAtMidpoint` operation that you call here.

tests/components/test_component_regular_space.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include <wmtk_components/mesh_info/mesh_info.hpp>
+#include <wmtk_components/regular_space/internal/RegularSpace.hpp>
+#include <wmtk_components/regular_space/regular_space.hpp>
+
+using json = nlohmann::json;
+using namespace wmtk;
+
+const std::filesystem::path data_dir = WMTK_DATA_DIR;
+
+TriMesh hex_plus_two()
```
daniel-zint, line 17: This examples exist in `tools/TriMesh_examples.hpp`. Use those.

tests/components/test_component_regular_space.cpp
```diff
@@ -0,0 +1,133 @@
...
+    V.row(5) << 2, 0, 0;
+    V.row(6) << 3, 0, 0;
+    V.row(7) << 0.5, -1, 0;
+    V.row(8) << 1.5, -1, 0;
+
+    mesh_utils::set_matrix_attribute(V, "position", PrimitiveType::Vertex, m);
+    return m;
+}
+
+TEST_CASE("file reading", "[components][regular_space][.]")
```
daniel-zint, line 59: Test cases must not use space. `file reading` --> `file_reading`.
Furthermore, be more specific with the test case name. The test cases have to have unique names across all tests.

tests/components/test_component_regular_space.cpp
```diff
@@ -0,0 +1,133 @@
...
+    std::map<std::string, std::filesystem::path> files;
+    std::map<std::string, long> tags_value;
+
+    json regular_space_jason = {
+        {"type", "regular space"},
+        {"input", "inputdir"}, /*input dir*/
+        {"output", "outputdir"}, /*output dir*/
+        {"demension", 1}, /*0 for vertex, 1 for edge, 2 for face, 3 for tet*/
+        {"tags_value", tags_value},
+        {"split_tag_value"}};
```
daniel-zint, line 70: This component does not do anything for now and I assume it should do more in the future.

Add a TODO comment, explaining what needs to be done and add a `REQUIRE(false)`.

tests/components/test_component_regular_space.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+TEST_CASE("1d case", "[components][regular_space][scheduler]")
+{
+    long input_value = 1;
```
daniel-zint, line 75: `const`

tests/components/test_component_regular_space.cpp
```diff
@@ -0,0 +1,133 @@
...
+    long embedding_value = 0;
+    long split_value = 2;
+
+    TriMesh m = hex_plus_two_with_position();
+    MeshAttributeHandle<double> pos_handle =
+        m.get_attribute_handle<double>(std::string("position"), wmtk::PrimitiveType::Vertex);
```
daniel-zint, line 81: `"position"` is sufficient. There is no need for `std::string( )`.

tests/components/test_component_regular_space.cpp
```diff
@@ -0,0 +1,133 @@
...
+        m,
+        pos_handle,
+        vertex_tag_handle,
+        edge_tag_handle,
+        input_value,
+        embedding_value,
+        split_value,
+        0);
+    rs.process();
+    ParaviewWriter writer1(data_dir / "my_result", "position", m, true, true, true, false);
```
daniel-zint, line 119: Despite the current issue with the `ParaviewWriter`, you can (and should) run checks here. You know what result you expect so iterate through the mesh and make sure that all attributes are as expected.

tests/test_2d_operations.cpp
```diff
@@ -1170,3 +1173,300 @@ TEST_CASE("swap_edge", "[operations][swap][2D]")
...
+        OperationSettings<tri_mesh::FaceSplit> settings;
+        settings.initialize_invariants(m);
+        wmtk::operations::tri_mesh::FaceSplit face_split_op(m, f, settings);
+        bool is_success = face_split_op();
+        Tuple ret = face_split_op.return_tuple();
+        spdlog::info("{}", m.id(ret, PV));
+        spdlog::info("{}", m.id(m.switch_vertex(ret), PV));
+        spdlog::info("{}", m.id(m.switch_vertex(m.switch_edge(ret)), PV));
+        REQUIRE(is_success);
+        REQUIRE(m.get_all(PV).size() == 4);
```
daniel-zint, line 1204: Everything from this line on should be `CHECK`s not `REQUIRE`s.

tests/test_2d_operations.cpp
```diff
@@ -1170,3 +1173,300 @@ TEST_CASE("swap_edge", "[operations][swap][2D]")
...
+        spdlog::info("{}", m.id(m.switch_vertex(split_ret), PV));
+        spdlog::info("{}", m.id(m.switch_vertex(m.switch_edge(split_ret)), PV));
+        const Tuple coll_input_tuple = m.switch_face(m.switch_edge(split_ret));
+        OperationSettings<tri_mesh::EdgeCollapse> collapse_settings;
+        collapse_settings.initialize_invariants(m);
+        tri_mesh::EdgeCollapse coll_op(m, coll_input_tuple, collapse_settings);
+        coll_op();
+        const Tuple& coll_ret = coll_op.return_tuple();
+        auto ret = m.switch_vertex(m.switch_edge(coll_ret));
+        spdlog::info("{} {}", m.id(ret, PV), m.id(m.switch_vertex(ret), PV));
```
daniel-zint, line 1245: I guess that is still debug code but `CHECK`s are missing and the tests should not contain any debug messages, if not necessary.

daniel-zint (changes requested):
The approach is definitely good. The code requires quite some clean-up though.

components/wmtk_components/regular_space/internal/RegularSpace.hpp
```diff
@@ -0,0 +1,44 @@
...
+{
+    TriMesh& m_mesh;
+    bool m_lock_boundary;
+
+    MeshAttributeHandle<double> m_position_handle;
+    MeshAttributeHandle<long> m_vertex_tag;
+    MeshAttributeHandle<long> m_edge_tag;
+    long m_input_tag_value;
+    long m_embedding_tag_value;
+    long m_split_tag_value;
```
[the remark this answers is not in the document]
Zhouyuan-Chen: when you split a simplicity, you will set m_split_tag_value to the new simplicity

Zhouyuan-Chen:
I still need to add the last few tests. Will be done soon.

Zhouyuan-Chen:
This component is done, wait to be reviewed

Zhouyuan-Chen:
I modified the code, but some codes I can't cover by tests, seems those codes are impossible to execute😅. however, for the safety of the code, those code is needed.

daniel-zint (approved):
Lgtm. Just resolve my comments from the previous review.

Zhouyuan-Chen:
Just remove the variable never used in the test file

daniel-zint (approved):
Good for now. It's ready to be merged.

PR #439, opened 2023-10-12
Add edge swap invariants

daniel-zint:
Refactoring the edge swap to use invariants instead of directly using `before` and `after` methods in the operation.


src/wmtk/invariants/MinIncidentValenceInvariant.cpp
```diff
@@ -0,0 +1,37 @@
...
+MinIncidentValenceInvariant::MinIncidentValenceInvariant(const Mesh& m, long min_valence)
+    : MeshInvariant(m)
+    , m_min_valence(min_valence)
+{}
+
+bool MinIncidentValenceInvariant::before(const Tuple& t) const
+{
+    const Tuple v0 = t;
+    const Tuple v1 = mesh().switch_vertex(t);
+    const long val0 = static_cast<long>(SimplicialComplex::vertex_one_ring(mesh(), v0).size());
```
mtao, line 18: lets not use the SimplicialComplex interface actively in new code

src/wmtk/invariants/MinIncidentValenceInvariant.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+    return val0 >= m_min_valence && val1 >= m_min_valence;
+}
+
+bool MinIncidentValenceInvariant::after(PrimitiveType type, const std::vector<Tuple>& t) const
+{
+    assert(type == PrimitiveType::Edge);
+
+    for (const Tuple& e : t) {
+        if (!before(e)) {
```
mtao, line 29: if we're going to call before in after might as well specify another function like `valence_improved(Tuple)` and call that. Then the semantic is that we are doing a type of check in both the before and after, rather than that we are checking preconditions in postdoncditiosn

src/wmtk/invariants/MinIncidentValenceInvariant.cpp
```diff
@@ -0,0 +1,37 @@
...
+    const Tuple v1 = mesh().switch_vertex(t);
+    const long val0 = static_cast<long>(SimplicialComplex::vertex_one_ring(mesh(), v0).size());
+    const long val1 = static_cast<long>(SimplicialComplex::vertex_one_ring(mesh(), v1).size());
+
+    return val0 >= m_min_valence && val1 >= m_min_valence;
+}
+
+bool MinIncidentValenceInvariant::after(PrimitiveType type, const std::vector<Tuple>& t) const
+{
+    assert(type == PrimitiveType::Edge);
```
mtao, line 26: don't assert - every type of tuple SHOULD be passed to all of these invariants - it's up to us to return true for the ones we don't care about

src/wmtk/operations/tri_mesh/EdgeSwapBase.cpp
```diff
@@ -0,0 +1,100 @@
...
+    //   /   \
+    //  /  f  \
+    // X--->---
+    //  \     /
+    //   \   /
+    //    \ /
+
+    Tuple split_ret;
+    {
+        OperationSettings<tri_mesh::EdgeSplit> op_settings;
```
mtao, line 48: the opsettings should be be members of the swap op rather than created every time - we don't want to initialize invariants in operations

src/wmtk/operations/tri_mesh/EdgeSwapValence.cpp
```diff
@@ -0,0 +1,61 @@
...
+    return "tri_mesh_edge_swap_valence";
+}
+
+bool EdgeSwapValence::execute()
+{
+    const Tuple v0 = input_tuple();
+    const Tuple v1 = mesh().switch_vertex(input_tuple());
+    const Tuple v2 = mesh().switch_vertex(mesh().switch_edge(input_tuple()));
+    const Tuple v3 = mesh().switch_vertex(mesh().switch_edge(mesh().switch_face(input_tuple())));
+    long val0 = static_cast<long>(SimplicialComplex::vertex_one_ring(mesh(), v0).size());
```
mtao, line 26: i think you do this long sequence in quite a few places - lets just make this a function that is shared somewhere/

src/wmtk/operations/tri_mesh/EdgeSwapValence.cpp
```diff
@@ -0,0 +1,61 @@
...
+{}
+
+std::string EdgeSwapValence::name() const
+{
+    return "tri_mesh_edge_swap_valence";
+}
+
+bool EdgeSwapValence::execute()
+{
+    const Tuple v0 = input_tuple();
```
mtao, line 22: could we implement this valence thing in the standard swap? we just have a bool for if we want to do it / have a std::optional<long> initial_valence` ?. after checkpoints/if checkpoints gets finished this should just be an invariant anyway

mtao (changes requested):

src/wmtk/operations/tri_mesh/EdgeSwapValence.cpp
daniel-zint: I opened #478 for that

src/wmtk/invariants/MinIncidentValenceInvariant.cpp
daniel-zint: #478

mtao (approved):
lgtm

PR #448, opened 2023-10-13
Zhouyuan/edge mesh input

Zhouyuan-Chen:
Reopen the edge mesh input. I closed this PR since I changed the name recently.

This part's EdgeMesh is outdated and need to update that part code.


src/wmtk/io/EdgeMeshReader.hpp
```diff
@@ -0,0 +1,51 @@
+#include <Eigen/Core>
+#include <filesystem>
+#include <fstream>
+
+namespace wmtk {
+class Mesh;
+
+class EdgeMeshReader
+{
+public:
```
daniel-zint, line 10: Make sure that only those functions are `public` that need to be. The rest should be `private`.

src/wmtk/io/EdgeMeshReader.hpp
```diff
@@ -0,0 +1,51 @@
...
+    enum data_type { V, L, VT, VN, VP, COMMENT };
+    enum file_type { OBJ, OFF };
+    EdgeMeshReader(const std::string& filename, const file_type type);
+    void read(
```
daniel-zint, line 14: All these functions need to covered in tests.

components/wmtk_components/input/input.cpp
```diff
@@ -52,11 +56,65 @@ void input(const nlohmann::json& j, std::map<std::string, std::filesystem::path>
         files[options.name] = cached_mesh_file;
         break;
     }
-    case 1:
+    case 1: {
         // edge mesh
-        throw "not implemented";
-        assert(false);
+        spdlog::warn("this edge mesh is old version, need to update according to the latest "
```
daniel-zint, line 61: Now as you have the `EdgeMesh` available, you should remove the warning and adapt the code.

daniel-zint (changes requested):
The new classes requires testing. I assume the code works but we need unit tests to prove it.

Zhouyuan-Chen:
Added the tests.
Edge Mesh Input finished.

daniel-zint:
Whenever you add the tag `[.]` to a test, it is ignored. You need to remove them.

daniel-zint (changes requested):
I'll review it once the tests are running in the CLI.

You can upload your OBJ file for testing here:
https://github.com/wildmeshing/data

Zhouyuan-Chen:
This component will not be used anymore, I will delete this branch later.

PR #454, opened 2023-10-16
Return tuple of `split_edge` is inconsistent

daniel-zint:
There was a bug in the return tuple of `split_edge`. The face inside the tuple was wrong in some scenarios.

### DIFF
--- a/src/wmtk/TriMeshOperationExecutor.cpp
+++ b/src/wmtk/TriMeshOperationExecutor.cpp
@@ -438,7 +438,7 @@ void TriMesh::TriMeshOperationExecutor::split_edge_single_mesh()
         ret = m_mesh.switch_face(ret);
         assert(m_mesh.id_face(ret) == new_tuple_fid);
     }
-    assert(m_mesh.is_valid_slow(ret));
+    assert(m_mesh.is_valid(ret, hash_accessor));
 }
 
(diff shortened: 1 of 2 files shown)


src/wmtk/TriMeshOperationExecutor.cpp
```diff
@@ -78,6 +78,12 @@ TriMesh::TriMeshOperationExecutor::TriMeshOperationExecutor(
         m_incident_face_datas.emplace_back(get_incident_face_data(f.tuple()));
     }
 
+    assert(m_incident_face_datas.size() <= 2);
+    if (m_incident_face_datas[0].fid != m.id_face(m_operating_tuple)) {
+        assert(m_incident_face_datas.size() == 2);
+        std::swap(m_incident_face_datas[0], m_incident_face_datas[1]);
+    }
```
zlyfunction, line 85: LGTM. According to my understanding, the reason for performing this swap is that when computing the `closed_star`, the returned faces are sorted by their IDs. However, since we may need to revert to the original order later on, why do we still use` closed_star` to retrieve `m_incident_face_datas`?

zlyfunction (approved):

src/wmtk/TriMeshOperationExecutor.cpp
daniel-zint: This was meant as a generalization for the 3d case. It is true that in this case, just requesting the two neighbors would have been easier.

PR #457, opened 2023-10-16
456 improve semantics of functions to clarify what simplex they operate on

mtao:


mtao:
changes shuld already be merged into yuf

mtao:
merged into the energy branch so https://github.com/wildmeshing/wildmeshing-toolkit/pull/350 can be checked

PR #460, opened 2023-10-16
Added Halfedge primitive type and refactored primitive type indexing

rjc8237:


rjc8237:
I added a HalfEdge attribute. The most potentially problematic changes are code snippets where a correspondence between dimension and a PrimitiveType is assumed, and especially where iteration over dimensions are used. In particular, the attribute manager in `Mesh` used `dimension+1` as the number of attributes and flags to initialize. I made the function `get_dimension_primitive_counts` to explicitly get the number of primitives for point, edge, surface, and volumetric meshes.

src/wmtk/Mesh.cpp
```diff
@@ -13,13 +13,17 @@ Mesh::Mesh(const Mesh& other) = default;
 Mesh& Mesh::operator=(const Mesh& other) = default;
 Mesh& Mesh::operator=(Mesh&& other) = default;
 Mesh::Mesh(const long& dimension)
-    : m_attribute_manager(dimension + 1)
-    , m_cell_hash_handle(register_attribute<long>("hash", static_cast<PrimitiveType>(dimension), 1))
+    : Mesh(dimension, dimension, get_primitive_type_from_id(dimension))
+{}
+
+Mesh::Mesh(const long& dimension, const long& max_primitive_type_id, PrimitiveType hash_type)
```
rjc8237, line 19: I made the type of the hash explicit as well instead of inferring from dimension.

src/wmtk/Mesh.cpp
```diff
@@ -13,13 +13,17 @@ Mesh::Mesh(const Mesh& other) = default;
...
+    : m_attribute_manager(max_primitive_type_id + 1)
+    , m_cell_hash_handle(register_attribute<long>("hash", hash_type, 1))
 {
-    m_flag_handles.reserve(dimension + 1);
-    for (long j = 0; j <= dimension; ++j) {
+    m_flag_handles.reserve(max_primitive_type_id + 1);
+    for (long j = 0; j <= max_primitive_type_id; ++j) {
         m_flag_handles.emplace_back(
-            register_attribute<char>("flags", static_cast<PrimitiveType>(j), 1));
+            register_attribute<char>("flags", get_primitive_type_from_id(j), 1));
```
rjc8237, line 26: I made a function to explicitly convert from IDs to primitive types. We could just use `static_cast` and initializers in the enumerator list (as was done before), but I think this more explicit conversion function is safer.

mtao (approved):
just add one include and try to remove as many derprecations as you can

PR #461, opened 2023-10-16
upload the embedding part code

Zhouyuan-Chen:
this part is the embedding function code.

not too much code to review, since we did it before. just copy and paste for the previous embedding code(from my repository).

This part code used the EdgeMeshReader class.


daniel-zint (changes requested):
tests are missing

Zhouyuan-Chen:
Ahh... Just realized that this component can't only be tested by checking the result with the Paraview.

But, confusion arose, do we really need this component now? As we already have the Delaunay component.

daniel-zint:
> Ahh... Just realized that this component can't only be tested by checking the result with the Paraview.
> 
> But, confusion arose, do we really need this component now? As we already have the Delaunay component.

Yes, it is still required. The Delaunay component only generates an embedding of a point cloud. With this component we also want to be able to embed an `EdgeMesh` in a `TriMesh` or `TetMesh`. Basically any lower dimension should be embeddable at some point. So, the Delaunay component should be called when embedding a point cloud in a `TriMesh` or `TetMesh`.

For your project, we could also generate test data only with the Delaunay component but sooner or later we need also the embedding of an `EdgeMesh`.

Zhouyuan-Chen:
I cleaned up this component code, and have checked the result, should work.

for the EdgeMesh, maybe we should add a small feature, convert EdgeMesh to E and V matrix.

Zhouyuan-Chen:
As the same, this component doesn't need too much review since we did it before.

components/wmtk_components/embedding/internal/Embedding.cpp
```diff
@@ -0,0 +1,152 @@
+#include "Embedding.hpp"
+#include "igl/writeOBJ.h"
+
+namespace wmtk::components::internal {
+Embedding::Embedding(
+    Eigen::Matrix<long, -1, -1>& m_edges_,
+    Eigen::MatrixXd& m_vertices_,
+    EmbeddingOptions& options_,
+    double m_blank_rate_
+    // double m_resolute_area_)
```
daniel-zint, line 10: remove old code

components/wmtk_components/embedding/internal/Embedding.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include <igl/adjacency_list.h>
+#include <igl/edges.h>
+#include <igl/triangle/triangulate.h>
+#include <wmtk/Scheduler.hpp>
+#include <wmtk/TriMesh.hpp>
+#include "EmbeddingOptions.hpp"
+
+namespace wmtk::components::internal {
+
+class Embedding
```
daniel-zint, line 12: add a description to the class

components/wmtk_components/embedding/internal/Embedding.hpp
```diff
@@ -0,0 +1,46 @@
...
+{
+public:
+    Embedding(
+        Eigen::Matrix<long, -1, -1>& m_edges_,
+        Eigen::MatrixXd& m_vertices_,
+        EmbeddingOptions& options_,
+        double m_blank_rate_ = 0.5
+        // double m_resolute_area = 0.1
+    );
+    void compute_bounding_value(double& max_x, double& max_y, double& min_x, double& min_y);
```
daniel-zint, line 22: add a description to this method

components/wmtk_components/embedding/internal/Embedding.cpp
```diff
@@ -0,0 +1,152 @@
...
+    m_marked_edges.reserve(m_edges.rows());
+
+    for (size_t i = 0; i < m_vertices.rows(); ++i) {
+        m_marked_vertices.emplace_back(i);
+    }
+    for (size_t i = 0; i < m_edges.rows(); ++i) {
+        m_marked_edges.emplace_back(std::pair<long, long>(m_edges(i, 0), m_edges(i, 1)));
+    }
+}
+void Embedding::compute_bounding_value(double& max_x, double& max_y, double& min_x, double& min_y)
```
daniel-zint, line 28: Use structured binding instead of reference input parameters:
```
std::tuple<double,double,double,double> Embedding::compute_bounding_value()
```

components/wmtk_components/embedding/internal/Embedding.cpp
```diff
@@ -0,0 +1,152 @@
...
+    max_y += blank_region_length_y;
+    min_y -= blank_region_length_y;
+}
+void Embedding::process()
+{
+    // find boundingbox
+    double max_x, min_x, max_y, min_y;
+    compute_bounding_value(max_x, max_y, min_x, min_y);
+
+    Eigen::MatrixXd bounding_box_vertices(4, 2);
```
daniel-zint, line 51: This has static size, so you can use `Eigen::Matrix<double, 4, 2>`.
I hope I got the syntax right but my main point is that you should avoid using `Eigen::MatrixXd` whenever possible.

components/wmtk_components/embedding/internal/Embedding.cpp
```diff
@@ -0,0 +1,152 @@
...
+    Eigen::Matrix<long, -1, -1> bounding_box_edges(4, 2);
```
daniel-zint, line 52: same here: `Eigen::Matrix<double, 4, 2>`

components/wmtk_components/embedding/internal/Embedding.cpp
```diff
@@ -0,0 +1,152 @@
...
+    Eigen::Matrix<long, -1, -1> temp_edges(m_edges.rows() + bounding_box_edges.rows(), 2);
+    temp_vertices << m_vertices, bounding_box_vertices;
+    temp_edges << m_edges, bounding_box_edges;
+
+    // this is for triangulate function, should be outside of the bounding box.
+    Eigen::MatrixXd outside_vertex;
+    outside_vertex.resize(1, 2);
+    outside_vertex << max_x + 1.0, max_y + 1.0;
+
+    // for (int i = 0; i < m_vertices.rows(); i++) {
```
daniel-zint, line 69: remove that if it's not needed anymore

components/wmtk_components/embedding/internal/Embedding.cpp
```diff
@@ -0,0 +1,152 @@
...
+    //     spdlog::info("{} {}", m_vertices(i, 0), m_vertices(i, 1));
+    // }
+
+    // for (int i = 0; i < m_edges.rows(); i++) {
+    //     spdlog::info("{} {}", m_edges(i, 0), m_edges(i, 1));
+    // }
+
+    // the fourth parameter is resolution, we should discuss it maybe.
```
daniel-zint, line 77: The embedding component should do as least as possible, i.e. the worst resolution that is still valid.

components/wmtk_components/embedding/internal/Embedding.cpp
```diff
@@ -0,0 +1,152 @@
...
+        m_faces);
+
+    Eigen::MatrixXd V(m_vertices.rows(), 3);
+    for (int i = 0; i < m_vertices.rows(); ++i) {
+        V(i, 0) = m_vertices(i, 0);
+        V(i, 1) = m_vertices(i, 1);
+        V(i, 2) = 0;
+    }
+    // const std::filesystem::path data_dir =
+    //     "/home/zhouyuan/workplace/embedding_part/wildmeshing-toolkit/data";
```
daniel-zint, line 93: that comment should be definitely removed

components/wmtk_components/embedding/internal/Embedding.cpp
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+    std::sort(temp_marked_vertices.begin(), temp_marked_vertices.end());
+    temp_marked_vertices.erase(
+        std::unique(temp_marked_vertices.begin(), temp_marked_vertices.end()),
+        temp_marked_vertices.end());
+    m_marked_edges = temp_marked_edges;
+    m_marked_vertices = temp_marked_vertices;
+
+    m_vertex_tags = std::vector<long>(m_vertices.rows(), options.embedding_tag_value);
```
daniel-zint, line 146: I think these tags should be edge attributes, am I right?

components/wmtk_components/embedding/internal/EmbeddingOptions.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <spdlog/spdlog.h>
+#include <nlohmann/json.hpp>
+
+namespace wmtk {
+namespace components {
+namespace internal {
+
+// this enum just for a test, we definitely should discuss the resolution design later
+enum Resolution_Level { LOW, MID, HIGH };
```
daniel-zint, line 11: resolution should be always "LOW"

components/wmtk_components/embedding/internal/Embedding.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+    // this variable is not used for now, relate to the resolution and need to be discussed
+    // double m_resolute_area;
+
+    Eigen::Matrix<long, -1, -1> m_edges;
+    Eigen::MatrixXd m_vertices;
+    Eigen::Matrix<long, -1, -1> m_faces;
+    // EmbeddingOptions options;
+
+    // 0 for scallfold, 1 for input
```
daniel-zint, line 39: this comment is outdated

src/wmtk/io/EdgeMeshReader.cpp
```diff
@@ -0,0 +1,164 @@
...
+    std::vector<std::vector<double>>& vertices_normal,
+    std::vector<std::vector<double>>& vertices_parameter)
+{
+    std::ifstream f(m_filename);
+    if (!f.is_open()) {
+        throw std::runtime_error("can't open file!");
+    }
+
+    std::string buffer;
+    while (!f.eof()) {
```
daniel-zint, line 86: Follow the concept from that stackoverflow answer:
https://stackoverflow.com/questions/7868936/read-file-line-by-line-using-ifstream-in-c

```
#include <sstream>
#include <string>

std::string line;
while (std::getline(infile, line))
{
    std::istringstream iss(line);
    int a, b;
    if (!(iss >> a >> b)) { break; } // error

    // process pair (a,b)
}
```

There is no need for using `f.eof()`.

src/wmtk/io/EdgeMeshReader.cpp
```diff
@@ -0,0 +1,164 @@
...
+        std::vector<std::string> tokens;
+        std::stringstream line(buffer);
+        std::string token;
+        getline(line, token, ' ');
+        switch (str_to_type(token)) {
+        case V: {
+            std::vector<double> positions;
+            while (getline(line, token, ' ')) {
+                // spdlog::info("lineinfo: {}", token);
+                positions.push_back(std::atof(token.c_str()));
```
daniel-zint, line 97: Please replace that by `std::stringstream`. Do not use `std::atof`.

src/wmtk/io/EdgeMeshReader.cpp
```diff
@@ -0,0 +1,164 @@
...
+        }; break;
+        case VN: {
+            std::vector<double> normal;
+            while (getline(line, token, ' ')) {
+                normal.push_back(std::atof(token.c_str()));
+            }
+            assert(normal.size() == 3);
+            double standard =
+                sqrt(normal[0] * normal[0] + normal[1] * normal[1] + normal[2] * normal[2]);
+            normal[0] /= standard;
```
daniel-zint, line 147: You can do the normalization using `Eigen::Map`:

```
Eigen::Map<Eigen::Vector3d>(normal).normalize()
```
I am not entirely sure if I got the syntax correct but it should be something like that.

Here is the documentation about `Eigen::Map`:
https://eigen.tuxfamily.org/dox-devel/classEigen_1_1Map.html

src/wmtk/io/EdgeMeshReader.hpp
```diff
@@ -0,0 +1,48 @@
+#include <Eigen/Core>
+#include <filesystem>
+#include <fstream>
+
+namespace wmtk {
+class Mesh;
+
+class EdgeMeshReader
```
daniel-zint, line 8: needs documentation

daniel-zint (changes requested):
The concepts seem fine but the code needs a bit of clean-up.

components/wmtk_components/embedding/internal/Embedding.cpp
Zhouyuan-Chen: Okay, then I will set the resolution constant as "a0.1q", which means
a0.1 means that the area of each triangle should be not greater than 0.1
q means that no angles will be smaller than 20 degrees

daniel-zint: If possible, remove this completely or set it to the minimum value possible. We just want a valid solution from the embedding, not necessarily a good or even reasonable one.

components/wmtk_components/embedding/internal/Embedding.cpp
Zhouyuan-Chen: m_marked_edge only contains the index of the edges on the input edges.
Since the index is not match to the toolkit's index, I convert these attribute into the trimesh in the file embedding.cpp

Zhouyuan-Chen:
changed the code and removed the c-style function. Maybe we can abandon the mesh_input_part code since there is mass overlap between the mesh_input_component and this component😅(both of them have a similar EdgeMeshReader class and the input component).

Zhouyuan-Chen:
Seems that the Windows Debug mode will lead the program to a dead loop. Not clear about the reason.

daniel-zint:
> changed the code and removed the c-style function. Maybe we can abandon the mesh_input_part code since there is mass overlap between the mesh_input_component and this component😅(both of them have a similar EdgeMeshReader class and the input component).
> 
> If we abandon the mesh_input branch, then I will move the test file into this branch.

The Embedding component should NOT read anything from file. It will only read from the cache. Any input has to be dealt with in the input component.
I did not realize that before but this component should not even have an `EdgeMeshReader`.

components/wmtk_components/embedding/internal/Embedding.cpp
```diff
@@ -41,15 +39,16 @@ void Embedding::compute_bounding_value(double& max_x, double& max_y, double& min
...
     max_y += blank_region_length_y;
     min_y -= blank_region_length_y;
+    return std::tuple<double, double, double, double>(max_x, max_y, min_x, min_y);
 }
 void Embedding::process()
 {
     // find boundingbox
     double max_x, min_x, max_y, min_y;
-    compute_bounding_value(max_x, max_y, min_x, min_y);
+    std::tie(max_x, max_y, min_x, min_y) = compute_bounding_value();
```
daniel-zint, line 48: `const auto [max_x, max_y, min_x, min_y] = compute_bounding_value();`

Zhouyuan-Chen:
Ahh, okay, then I will remove the EdgeMeshReader. Then the input of the embedding part should be the EdgeMesh. Got it.

components/wmtk_components/embedding/internal/Embedding.hpp
```diff
@@ -8,7 +8,24 @@
...
+ * m_edges: vertices' indices pairs of the edges
+ * m_vertices: vertex position
+ * options: please see the document in EmbeddingOptions.hpp
+ * m_black_rate: used to define the Bbox size
+ * @output
+ * m_faces: 2D mesh composed of faces
+ * m_edges: 2D mesh's edges
+ * m_vertex_tags: each vertex's tag
+ * m_marked_vertices: vertices on the input edges
+ * m_marked_edges: edges on the input edges
```
daniel-zint, line 27: The documentation itself is fine but please write it with Doxygen-style:
https://www.doxygen.nl/manual/docblocks.html

Zhouyuan-Chen:
> > changed the code and removed the c-style function. Maybe we can abandon the mesh_input_part code since there is mass overlap between the mesh_input_component and this component😅(both of them have a similar EdgeMeshReader class and the input component).
> > If we abandon the mesh_input branch, then I will move the test file into this branch.
> 
> The Embedding component should NOT read anything from file. It will only read from the cache. Any input has to be dealt with in the input component. I did not realize that before but this component should not even have an `EdgeMeshReader`.

Ahh, okay, then I will remove the EdgeMeshReader. Then the input of the embedding part should be the EdgeMesh. Got it.

daniel-zint (changes requested):
Please reformat the documentation to Doxygen style. That way, we can later on generate the documentation automatically.

Zhouyuan-Chen:
I still need to integrate the Delaunay triangulation into this component.

Zhouyuan-Chen:
I added 1d cases, hopefully, I will upload the code tonight. not too much to review I guess. then I will begin to review the code of the multimesh.

Zhouyuan-Chen:
the embedding code for now should be considered to have an input modification. since we used the old version .hdf5 before and there is a new attribute topsimplex. so I can't use my original .hdf5 file.

PR #463, opened 2023-10-17
save all primitives, including deleted

teseoch:


src/wmtk/Mesh.hpp
```diff
@@ -67,9 +67,10 @@ class Mesh : public std::enable_shared_from_this<Mesh>
     /**
      * Generate a vector of Tuples from global vertex/edge/triangle/tetrahedron index
      * @param type the type of tuple, can be vertex/edge/triangle/tetrahedron
+     * @param include_deleted if true returns also the deleted tuples (default false)
      * @return vector of Tuples referring to each type
      */
-    std::vector<Tuple> get_all(PrimitiveType type) const;
+    std::vector<Tuple> get_all(PrimitiveType type, const double include_deleted = false) const;
```
mtao, line 73: bool.

mtao: Also I'm not sure if I agree with this API - it will almost certainly break and SHOULD break in debug mode because in theory every access to a tuple should validate that the tuple is valid and the tuple belonging to an active cell should be active. 

The point is to allow paraviewwriter to figure out which indices are active or not, and this seems like something that could be read off by the paraviewwriter using the flag_attribute_handles() and the AttributeManager object directly using indices. Rather than introduce an implicit index-revealing semantic in the `get_all` function I'd rather let the mesh writers have access to indices

mtao (changes requested):
The writer should access hte AttributeManager directly rather than go through the public mesh interface - this PR introduces a hack to the public mesh interface to reveal indices/killed simplex indices.

daniel-zint (changes requested):
After discussing with Michael, I agree that we should restrict ourselves to the attributes within the writer and not use the public interface of the mesh (e.g. `get_all`).

src/wmtk/Mesh.cpp
```diff
@@ -34,14 +34,19 @@ PrimitiveType Mesh::top_simplex_type() const
 }
 
 std::vector<Tuple> Mesh::get_all(PrimitiveType type) const
+{
+    return get_all(type, false);
+}
+
+std::vector<Tuple> Mesh::get_all(PrimitiveType type, const double include_deleted) const
```
daniel-zint, line 41: That should be a `bool` not a `double` I assume.

src/wmtk/io/ParaviewWriter.cpp
```diff
@@ -84,7 +84,8 @@ ParaviewWriter::ParaviewWriter(
     for (int i = 0; i < 4; ++i) {
         auto pt = PrimitiveType(i);
         if (m_enabled[i]) {
-            const auto tuples = mesh.get_all(pt);
+            // include delted tuples so that attributes are aligned
```
daniel-zint, line 87: "deleted"

src/wmtk/io/ParaviewWriter.cpp
```diff
@@ -84,7 +84,8 @@ ParaviewWriter::ParaviewWriter(
     for (int i = 0; i < 4; ++i) {
```
daniel-zint, line 84: As you are on it already anyways, please change `int` to `size_t` here.

daniel-zint (commented):
looks good, just the `double` needs to be changed to `bool`

src/wmtk/io/ParaviewWriter.cpp
teseoch: 4 is a constant that can be an int, why size_t

mtao: For me it's a good practice when potentially using stl accessors to prevent the casting, I always assume the cast down is very cheap/easy too

mtao:
is this ready now?

src/wmtk/io/ParaviewWriter.cpp
teseoch: agree, but this one is just 4 :)

daniel-zint: Sure but you intend to use it in a context of a `size_t`. This one here is not an issue but I have quite a list of warnings in Windows because people convert `size_t` into `long` everywhere.

So, here it is fine (does not produce a warning) but I'd prefer if we are in general more careful with integer types.

daniel-zint (approved):
lgtm

src/wmtk/Mesh.cpp
```diff
@@ -38,16 +38,22 @@ PrimitiveType Mesh::top_simplex_type() const
...
+std::vector<Tuple> Mesh::get_all(PrimitiveType type, const bool include_deleted) const
 {
     ConstAccessor<char> flag_accessor = get_flag_accessor(type);
     const attribute::CachingAccessor<char>& flag_accessor_indices = flag_accessor.index_access();
     std::vector<Tuple> ret;
     long cap = capacity(type);
     ret.reserve(cap);
     for (size_t index = 0; index < cap; ++index) {
-        if ((flag_accessor_indices.const_scalar_attribute(index) & 1)) {
+        if (flag_accessor_indices.const_scalar_attribute(index) & 1)
```
mtao, line 53: i like always having braces no matter what

mtao (approved):
lgtm

PR #474, opened 2023-10-26
added marching component

Zhouyuan-Chen:
most code is copied from the regularspace component.


Zhouyuan-Chen:
Have tested the marching component code, and the result is fine. Then I will do the embedding part, and add the Delaunay for the points cloud.

components/wmtk_components/marching/internal/Marching.hpp
```diff
@@ -0,0 +1,42 @@
...
+{
+    TriMesh& m_mesh;
+    bool m_lock_boundary;
+
+    MeshAttributeHandle<double> m_position_handle; // record position
+    MeshAttributeHandle<long> m_vertex_tag; // record vertex tag value
+    MeshAttributeHandle<long> m_edge_tag; // record edge vertex tag value
+    long m_input_tag_value; // the value used to those simplicity you want to seperate
+    long m_embedding_tag_value; // the scalffold simplicities' tag value
+    long m_split_tag_value; // when you split a simplicity, you will set m_split_tag_value to the
```
daniel-zint, line 18: This component should create an isosurface. Therefore, all that is required is the isovalue.

In your case for example, you might have an embedding tag of 2 and an input tag of 0. Then you can set your isovalue to 1. Every edge whose vertices have one tag above 1 and one tag below 1 are then split. The new vertex gets then the tag 1.

components/wmtk_components/marching/internal/Marching.cpp
```diff
@@ -0,0 +1,112 @@
...
+    , m_position_handle(position_handle)
+    , m_vertex_tag(vertex_tag)
+    , m_edge_tag(edge_tag)
+    , m_input_tag_value(input_tag_value)
+    , m_embedding_tag_value(embedding_tag_value)
+    , m_split_tag_value(split_tag_value)
+    , m_dimension(dimension)
+    , m_scheduler(mesh)
+    , m_lock_boundary(lock_boundary)
+{}
```
daniel-zint, line 33: A `dimension = 0` never makes sense. Marching requires `dimension` to be at least 1 or greater. You can add an assert here to make sure of that already in the constructor.

components/wmtk_components/marching/internal/Marching.cpp
```diff
@@ -0,0 +1,112 @@
...
+    default: throw std::runtime_error("settings went wrong!"); break;
+    }
+}
+
+void Marching::process_in_2d()
+{
+    using namespace operations;
+
+    wmtk::MeshAttributeHandle<long> todo_edgesplit_same_handle = m_mesh.register_attribute<long>(
+        std::string("todo_edgesplit_same_tag"),
```
daniel-zint, line 49: You don't need to write `std::string("...")` you can just write `"..."` directly.

I guess the name should be `"todo_edgesplit_same_handle"`?

components/wmtk_components/marching/internal/Marching.cpp
```diff
@@ -0,0 +1,112 @@
...
+                (vt1 == m_input_tag_value && vt0 == m_embedding_tag_value)) {
+                acc_todo_edgesplit_same_tag.scalar_attribute(edge) = 1;
+            }
+        }
+        // using scheduler to do edge splitting
+        OperationSettings<tri_mesh::EdgeSplitWithTag> settings_split;
+        settings_split.edge_tag = m_edge_tag;
+        settings_split.vertex_tag = m_vertex_tag;
+        settings_split.embedding_tag_value = m_embedding_tag_value;
+        settings_split.need_embedding_tag_value = true;
```
daniel-zint, line 77: What is that setting? It sounds oddly specific.

components/wmtk_components/marching/internal/Marching.cpp
```diff
@@ -0,0 +1,112 @@
...
+        // settings_split_same.position = m_position_handle;
+        settings_split.split_with_tag_settings.split_settings.split_boundary_edges = true;
```
daniel-zint, line 79: I think this should depend on `m_lock_boundary`.

components/wmtk_components/marching/internal/Marching.cpp
```diff
@@ -0,0 +1,112 @@
...
+        while (true) {
+            m_scheduler.run_operation_on_all(PrimitiveType::Edge, "edge_split");
+            if (m_scheduler.number_of_successful_operations() == 0) {
+                break;
+            }
+        }
+    }
+
+    // link the split vertices
+    {
```
daniel-zint, line 95: There is no need to create a block around a for-loop. The loop itself is already a block.

components/wmtk_components/marching/internal/Marching.cpp
```diff
@@ -0,0 +1,112 @@
...
+        for (const Tuple& t : m_mesh.get_all(PrimitiveType::Edge)) {
+            long vt0, vt1;
+            vt0 = acc_vertex_tag.const_scalar_attribute(t);
+            vt1 = acc_vertex_tag.const_scalar_attribute(m_mesh.switch_vertex(t));
+            if (vt0 == m_split_tag_value && vt1 == m_split_tag_value) {
+                acc_edge_tag.scalar_attribute(t) = m_split_tag_value;
```
daniel-zint, line 101: This code sounds to me like it should not be inside this component. It is very specific to your project.

components/wmtk_components/marching/internal/MarchingOptions.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <spdlog/spdlog.h>
+#include <nlohmann/json.hpp>
+
+namespace wmtk {
+namespace components {
+namespace internal {
+
+struct MarchingOptions
+{
+    std::string type; // regular space
```
daniel-zint, line 12: `// marching`

components/wmtk_components/marching/marching.cpp
```diff
@@ -0,0 +1,65 @@
...
+    MeshAttributeHandle<double> pos_handle =
+        mesh.get_attribute_handle<double>(options.pos_attribute_name, PrimitiveType::Vertex);
+    MeshAttributeHandle<long> vertex_tag_handle =
+        mesh.get_attribute_handle<long>(options.vertex_tag_handle_name, PrimitiveType::Vertex);
+    MeshAttributeHandle<long> edge_tag_handle =
+        mesh.get_attribute_handle<long>(options.edge_tag_handle_name, PrimitiveType::Edge);
+
+    switch (dim) {
+    case 2: {
+        Marching mc(
```
daniel-zint, line 37: This call is independent of the dimension.

components/wmtk_components/marching/internal/Marching.hpp
```diff
@@ -0,0 +1,42 @@
...
+    int m_dimension; // operate dimension, 0 for vertex, 1 for edge, 2 for face, 3 for tet
+
+    Scheduler m_scheduler;
+
+    void process_in_2d();
+    void process_in_3d();
+
+public:
+    Marching(
+        TriMesh& mesh,
```
daniel-zint, line 29: This cannot be a `TriMesh` if this code should be able to handle also the 3D case.

src/wmtk/operations/tri_mesh/EdgeSplitWithTag.cpp
```diff
@@ -44,12 +44,15 @@ Tuple EdgeSplitWithTag::return_tuple() const
...
 bool EdgeSplitWithTag::execute()
 {
-    // Eigen::Vector3d p0 = m_pos_accessor.const_vector_attribute(input_tuple());
-    // Eigen::Vector3d p1 =
-    // m_pos_accessor.const_vector_attribute(mesh().switch_vertex(input_tuple()));
-    long et = m_edge_tag_accessor.scalar_attribute(input_tuple());
-    long vt0 = m_vertex_tag_accessor.scalar_attribute(input_tuple());
-    long vt1 = m_vertex_tag_accessor.scalar_attribute(mesh().switch_vertex(input_tuple()));
+    // long et = m_edge_tag_accessor.scalar_attribute(input_tuple());
+    long vt0, vt1;
```
daniel-zint, line 48: I really don't like that way of declaring variables and defining them afterwards. Please avoid it if possible and then also make them `const`.

tests/components/test_component_marching.cpp
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        // should be manifold
+        {
+            Accessor<long> acc_edge_tag = m.create_accessor(edge_tag_handle);
+            for (const Tuple& edge : m.get_all(PrimitiveType::Edge)) {
+                if (acc_edge_tag.scalar_attribute(edge) == split_tag_value) {
+                    Tuple t = m.switch_face(m.switch_edge(edge));
+                    int neighbour_num = 0;
+                    while (t.operator!=(edge)) {
```
daniel-zint, line 102: `t != edge`

tests/components/test_component_marching.cpp
```diff
@@ -0,0 +1,169 @@
...
+                        if (acc_edge_tag.scalar_attribute(t) == split_tag_value) {
+                            neighbour_num++;
```
daniel-zint, line 104: `++neighbour_num` use pre instead of post increment

And maybe switch to American English: neighbour --> neighbor

tests/components/test_component_marching.cpp
```diff
@@ -0,0 +1,169 @@
...
+        //    0---1---2
+        //   / \ / \ / \ .
+        //  3---4---5---6
+        //   \ / \ /  .
+        //    7---8
+        // set edge 4 as input
+        {
+            const std::vector<Tuple>& vertex_tuples = m.get_all(wmtk::PrimitiveType::Vertex);
+            Accessor<long> acc_vertex_tag = m.create_accessor(vertex_tag_handle);
+            acc_vertex_tag.scalar_attribute(vertex_tuples[4]) = input_tag_value;
```
daniel-zint, line 155: this block is not required for that test

tests/components/test_component_marching.cpp
```diff
@@ -0,0 +1,169 @@
...
+            }
+        }
+
+        if (false) {
+            ParaviewWriter
+                writer(data_dir / "marching_2d_result", "position", m, true, true, true, false);
+            m.serialize(writer);
+        }
+    }
+    SECTION("3d_case -- should be manifold")
```
daniel-zint, line 119: That is the exact same case as the one above. If this was meant a as a placeholder for the future 3D test, please don't do that. Someone else might see that and think that 3D works even though this is not actually testing 3D.

daniel-zint (changes requested):
The Marching component should be designed in a way that it inserts an isosurface into a higher dimensional mesh. For doing so, a scalar field is defined on the vertices and an isovalue is given. Whenever an edge crosses that isovalue (e.g. `v0 < j` && `v1 >= j` where `j` is the isovalue and `v0`,`v1` are the incident vertices) this edge needs to be split.

Currently, the component is very specific to our project. Please change the API to the way described above.

components/wmtk_components/marching/internal/Marching.hpp
```diff
@@ -0,0 +1,42 @@
...
+namespace wmtk::components::internal {
+
+class Marching
+{
+    TriMesh& m_mesh;
+    bool m_lock_boundary;
+
+    MeshAttributeHandle<double> m_position_handle; // record position
+    MeshAttributeHandle<long> m_vertex_tag; // record vertex tag value
+    MeshAttributeHandle<long> m_edge_tag; // record edge vertex tag value
```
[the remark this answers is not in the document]
Zhouyuan-Chen: this edge tag will recode our offset link information. thought it could be useful for visualizing the result later.

components/wmtk_components/marching/internal/Marching.cpp
Zhouyuan-Chen: I will make a comment here. when you do a split operation, the new four edges could be tagged with a specific value. By default, they will be tagged as embedding values. but I think sometimes we don't want to change the original edge's tag, so this is where the need_embedding_tag_value comes from.

Zhouyuan-Chen:
For the tag, I added the filter tags. The filter tags should be set by the user. 

For the parameter of the marching, for now, I want to use the polymorphic function to make sure this component can handle different mesh types. That is to say, I will have another function called `process(TetMesh )` to handle the 3D case. And The constructor of Marching class doesn't need to have mesh parameter anymore.

Zhouyuan-Chen:
online check occurs error. don't know what has happened to my code...

Zhouyuan-Chen:
modified some code mentioned earlier.

daniel-zint (changes requested):
Please merge main into this branch and re-request my review afterwards.

Zhouyuan-Chen:
I will change the API parameter later.

components/wmtk_components/marching/internal/MarchingOptions.hpp.orig
daniel-zint, line 1: This file should not be here.

daniel-zint (approved):
I think this part is fine to be merged. There are a few things that need to be refactored but we should do that on a new branch. This one is already very old.

Just make sure that you do not add any .orig files, i.e. remove `MarchingOptions.hpp.orig`.

PR #476, opened 2023-10-29
simplex optimizations and top_level to top_dimension

mtao:
he term "level" was goofy and not meaningful / used in any other circumstance.

made simplexcollection use binary search to find if an element already exists + added an assert to check for sortedness

removed some trimesh/tetmesh only switching logic


daniel-zint (approved):
lgtm

PR #482, opened 2023-10-30
472 edgemesh trimesh multimesh tuple map updates after operations

zlyfunction:
1. Implemented the functor for updating the map between the edge mesh and the triangle mesh:
    - Registered the map for the new edges created by the split operation.
    - Updated the hash in the map for those edges that potentially underwent hash changes.
2. Fixed bugs in the multi-mesh hash map functions.


mtao (approved):
we said we would accept bascially as is and then fix after right?
feel free to merge

PR #485, opened 2023-10-31
smooth refactor

yunfanzhou:


src/wmtk/operations/tri_mesh/VertexLaplacianSmooth.hpp
```diff
@@ -30,8 +30,6 @@ class VertexLaplacianSmooth : public VertexAttributesUpdateBase
 
     std::string name() const override;
 
-    static PrimitiveType primitive_type() { return PrimitiveType::Vertex; }
```
daniel-zint, line 33: Is there a reason why you removed those? Is it deprecated code?

src/wmtk/operations/tri_mesh/VertexSmoothUsingDifferentiableEnergy.cpp
```diff
@@ -10,6 +11,9 @@ void OperationSettings<tri_mesh::VertexSmoothUsingDifferentiableEnergy>::initial
     base_settings.initialize_invariants(m);
     base_settings.invariants.add(
         std::make_shared<TriangleInversionInvariant>(m, coordinate_handle));
+    if (!smooth_boundary) {
+        base_settings.invariants.add(std::make_unique<InteriorVertexInvariant>(m));
```
daniel-zint, line 15: That sounds to me like it should not be done by having an operation setting but actually by adding an invariant on the user side. If what you did here was for the purpose of backwards compatibility, please write a comment so that we can refactor it in the future.

src/wmtk/operations/tri_mesh/VertexTangentialLaplacianSmooth.cpp
```diff
@@ -5,6 +5,14 @@
 #include <wmtk/utils/mesh_utils.hpp>
 #include "VertexLaplacianSmooth.hpp"
 
+namespace wmtk::operations {
+void OperationSettings<tri_mesh::VertexTangentialLaplacianSmooth>::initialize_invariants(
```
daniel-zint, line 9: That is quite a bad sign if this function is never called. I think it should be called for every operation at least once. We really need to refactor this initialization stuff.. But that's not in the scope of this PR.

src/wmtk/optimization/LineSearch.cpp
```diff
@@ -26,7 +26,7 @@ bool LineSearch::check_state() const
     bool before_pass = m_invariants.before(m_interface.tuple());
     bool after_pass = true;
     for (const PrimitiveType pt : wmtk::utils::primitive_below(top_type)) {
-        after_pass |= m_invariants.after(pt, modified_simplices(pt));
+        after_pass &= m_invariants.after(pt, modified_simplices(pt));
```
daniel-zint, line 29: What is this function doing? It feels wrong to me that you are calling the invariant's `before` and `after` here at the same time.

tests/components/test_smoothing.cpp
```diff
@@ -29,15 +35,15 @@ TEST_CASE("smoothing_Newton_Method")
         scheduler.add_operation_type<operations::tri_mesh::VertexSmoothUsingDifferentiableEnergy>(
             "optimize_vertices",
```
daniel-zint, line 36: Why is there the `spdlog::info("HJELL?");`? That loos a lot like debug code that should be removed.

tests/components/test_smoothing.cpp
```diff
@@ -29,15 +35,15 @@ TEST_CASE("smoothing_Newton_Method")
...
             std::move(op_settings));
-    Tuple tuple = mesh.face_tuple_from_vids(2, 4, 5);
-    spdlog::warn("Initial valuenorm: {}", factory.settings().energy->get_one_ring_value(tuple));
+    Tuple tuple = mesh.face_tuple_from_vids(0, 1, 0);
+    spdlog::warn("Initial valuenorm: {}", factory.settings().energy->get_value(tuple));
```
daniel-zint, line 39: I am not a big fan of logging output in tests. They make the test output very long and I think they should be only here for debugging but removed when the test is finished.

tests/components/test_smoothing.cpp
```diff
@@ -29,15 +35,15 @@ TEST_CASE("smoothing_Newton_Method")
...
     spdlog::warn(
         "Initial gradient: norm: {}",
-        factory.settings().energy->get_local_gradient(tuple).norm());
-    while (factory.settings().energy->get_local_gradient(tuple).norm() > 1e-10) {
+        factory.settings().energy->get_gradient(tuple).norm());
+    while (factory.settings().energy->get_gradient(tuple).norm() > 1e-10) {
```
daniel-zint, line 43: I'd prefer a `for`-loop here with a sufficiently large number of iterations. With a `while`-loop, if your code does not work as expected you will create an endless loop in the tests.

tests/components/test_smoothing.cpp
```diff
@@ -50,36 +56,35 @@ TEST_CASE("smoothing_Newton_Method")
...
-    Tuple tuple = mesh.face_tuple_from_vids(2, 4, 5);
-    spdlog::warn(
-        "Initial gradient: norm: {}",
-        factory.settings().energy->get_one_ring_gradient(tuple).norm());
-    while (factory.settings().energy->get_one_ring_gradient(tuple).norm() > 1e-10) {
-        scheduler.run_operation_on_all(PrimitiveType::Vertex, "optimize_vertices");
-        tuple = mesh.face_tuple_from_vids(2, 4, 5);
-    }
-    ConstAccessor<double> pos = mesh.create_const_accessor(op_settings.coordinate_handle);
+// TEST_CASE("smoothing_Newton_Method_line_search")
```
daniel-zint, line 59: If that test case is outdated, please remove it. If there is a good reason to keep that commented out code, please add a comment to it so that we know it should not be deleted.

daniel-zint (changes requested):
There are a few confusions on my side but that might also just be me not knowing what you are doing here 😅 
At least the `while`-loop in the tests I really would like to see removed. More details in the comments below.

src/wmtk/operations/tri_mesh/VertexLaplacianSmooth.hpp
mtao: I just checked the tupleoperation class and it has a broken use of this function. We should make an independent issue to fix this

src/wmtk/operations/tri_mesh/VertexSmoothUsingDifferentiableEnergy.cpp
mtao: Isn't this a property in the base settings? We should just use that one's settings

src/wmtk/optimization/LineSearch.cpp
mtao: This is a funny piece of logic I introduced. In an iterative algorithm each iteration is like a vertex smooth. So you need to satisfy before and after between each step

tests/components/test_smoothing.cpp
mtao: We should set them to trace

mtao:
as soon as https://github.com/wildmeshing/wildmeshing-toolkit/pull/552 is merged then this can continue

mtao:
@yunfanzhou  i'm not a huge fan of this TriMeshAutodiffFunction itnerface that tries to take in 3 args - i think sticking to a generic autodiff interface that just takes in the domain simplex is honestly still easy enough to work through / people can just follow the pattern rather than get forced to pick specific args.

Issue #490, opened 2023-11-03
Simplex refactor


rjc8237:
The `Simplex` class should be renamed to `Cell` and store a tuple and a cell dimension (i.e., a long) instead of a tuple and `PrimitiveType` to reflect the introduction of general polygonal face cells.

To allow for gradual deprecation of the `Simplex`, a new `Cell` class can be introduced supporting implicit conversion from a `Simplex`.

rjc8237:
The `Simplex` class is being used to compare equality of geometric primitives. We want to be able to determine if two HalfEdges are equal, so replacing the `PrimitiveType` with dimension does not accurately reflect the desired semantic.

Instead, we can just rename `Simplex` to `Primitive`. This naming more accurately reflects the idea of a `Tuple` being a tuple of primitives, while under the hood the `Primitive` type actually uses the `Tuple` handle and a `PrimitiveType` specification to algorithmically determine the specified primitive.

Some care is needed to make sure that `Primitive` is only used in contexts where the primitive is not assumed to be a `Cell`. We can also have a separate `Cell` class that requires the `PrimitiveType` to be one of the cell types (e.g., not `HalfEdge`).

PR #491, opened 2023-11-03
Added new cell class

rjc8237:
Addresses issue #490

### DIFF
--- a/src/wmtk/Cell.hpp
+++ b/src/wmtk/Cell.hpp
@@ -1,0 +1,35 @@
+#pragma once
+
+#include "Simplex.hpp"
+#include "Tuple.hpp"
+
+namespace wmtk {
+
+class Cell
+{
+    long m_dimension;
+    Tuple m_tuple;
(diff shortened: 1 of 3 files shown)


mtao:
good start - could you add implicit casts from Simplex -> Cell and Cell -> Simplex?

src/wmtk/Cell.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+namespace wmtk {
+
+class Cell
+{
+    long m_dimension;
+    Tuple m_tuple;
+
+public:
+    Cell(const long& dimension, const Tuple& t)
```
mtao, line 14: put implementations of as many functions in the cpp please

mtao (changes requested):
i guess we'll just make adding one PR and then do smaller PRs for updating pertinent code later right?

mtao (approved):
nvm lets just accept this for now and do further modifications later

PR #493, opened 2023-11-05
making is_boundary interfaces take in a primitive type

mtao:
now there is only one is_boundary virtual which takes in a tuple and primitive type,
all other ones are treated as implementation details

### DIFF
--- a/src/wmtk/Mesh.cpp
+++ b/src/wmtk/Mesh.cpp
@@ -119,6 +119,20 @@ long Mesh::capacity(PrimitiveType type) const
     return m_attribute_manager.m_capacities.at(get_primitive_type_id(type));
 }
 
+
+bool Mesh::is_boundary(const Tuple& tuple) const
+{
+    long my_dim = top_cell_dimension() - 1;
+    PrimitiveType pt = static_cast<PrimitiveType>(my_dim);
+    return is_boundary(tuple, pt);
+}
+
+bool Mesh::is_boundary(const Simplex& s) const
+{
+    return is_boundary(s.tuple(), s.primitive_type());
+}
+
+
 bool Mesh::is_hash_valid(const Tuple& tuple, const ConstAccessor<long>& hash_accessor) const
 {
     const long cid = tuple.m_global_cid;
--- a/src/wmtk/Mesh.hpp
+++ b/src/wmtk/Mesh.hpp
@@ -288,16 +288,24 @@ public:
      */
     virtual bool is_ccw(const Tuple& tuple) const = 0;
     /**
-     * @brief check if all tuple simplices besides the cell are on the boundary
+     * @brief check if a simplex of codimension 1 is a boundary simplex
      *
      * @param tuple
      * @return true if all tuple simplices besides the cell are on the boundary
      * @return false otherwise
      */
-    virtual bool is_boundary(const Tuple& tuple) const = 0;
+    bool is_boundary(const Tuple& codim_1_simplex) const;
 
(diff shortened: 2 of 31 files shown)


src/wmtk/Mesh.cpp
```diff
@@ -119,6 +119,20 @@ long Mesh::capacity(PrimitiveType type) const
     return m_attribute_manager.m_capacities.at(get_primitive_type_id(type));
 }
 
+
+bool Mesh::is_boundary(const Tuple& tuple) const
+{
+    long my_dim = top_cell_dimension() - 1;
```
daniel-zint, line 125: `const` (I know it's 3 lines of code and it doesn't really matter here but I still think it's a good habit)

src/wmtk/Mesh.hpp
```diff
@@ -288,16 +288,24 @@ class Mesh : public std::enable_shared_from_this<Mesh>
...
     /**
-     * @brief check if all tuple simplices besides the cell are on the boundary
+     * @brief check if a simplex of codimension 1 is a boundary simplex
      *
      * @param tuple
      * @return true if all tuple simplices besides the cell are on the boundary
      * @return false otherwise
      */
-    virtual bool is_boundary(const Tuple& tuple) const = 0;
+    bool is_boundary(const Tuple& codim_1_simplex) const;
```
daniel-zint, line 297: I feel like this version of `is_boundary` should not exist in the API. There is quite a risk of people using it unintentionally for the wrong simplices.

src/wmtk/PointMesh.cpp
```diff
@@ -26,8 +26,19 @@ bool PointMesh::is_ccw(const Tuple&) const
...
+    switch (pt) {
+    case PrimitiveType::Vertex: return is_boundary_vertex(tuple);
+    case PrimitiveType::Edge:
+    case PrimitiveType::Face:
+    case PrimitiveType::Tetrahedron:
+    case PrimitiveType::HalfEdge:
+    default: break;
+    }
+    throw std::runtime_error(
+        "tried to compute hte boundary of an edge mesh for an invalid simplex dimension");
```
daniel-zint, line 40: "the"

src/wmtk/PointMesh.cpp
```diff
@@ -26,8 +26,19 @@ bool PointMesh::is_ccw(const Tuple&) const
...
+    return false;
```
daniel-zint, line 41: Two `return` statements and both of them are after throwing an error so it is dead code I think.

src/wmtk/TriMesh.cpp
```diff
@@ -61,9 +61,19 @@ long TriMesh::id(const Tuple& tuple, PrimitiveType type) const
...
+    case PrimitiveType::Vertex: return is_boundary_vertex(tuple);
+    case PrimitiveType::Edge: return is_boundary_edge(tuple);
+    case PrimitiveType::Face:
+    case PrimitiveType::Tetrahedron:
+    case PrimitiveType::HalfEdge:
+    default: break;
+    }
+    throw std::runtime_error(
+        "tried to compute hte boundary of an edge mesh for an invalid simplex dimension");
+    return false;
```
daniel-zint, line 76: `return` statement is dead code

daniel-zint (changes requested):
Looks good but I think we should not have a `is_boundary` function without a primitive type.

daniel-zint (approved):
Deprecating `is_boundary` will do the job for now. It cannot be deleted as other code might depend on it.

PR #499, opened 2023-11-06
added split and collapse operations

Zhouyuan-Chen:
Added split and collapse operations for tetrahedra mesh.

### DIFF
--- a/src/wmtk/operations/tet_mesh/TetSplit.cpp
+++ b/src/wmtk/operations/tet_mesh/TetSplit.cpp
@@ -113,27 +113,62 @@ Tuple TetSplit::return_tuple() const
 std::vector<Tuple> TetSplit::modified_primitives(PrimitiveType type) const
 {
     Simplex v(PrimitiveType::Vertex, mesh().switch_vertex(m_output_tuple));
-    auto sc = SimplicialComplex::open_star(mesh(), v);
+    const SimplicialComplex& sc = SimplicialComplex::closed_star(mesh(), v);
     std::vector<Tuple> ret;
     if (type == PrimitiveType::Tetrahedron) {
-        auto tets = sc.get_simplices(PrimitiveType::Tetrahedron);
-        for (const auto& tet : tets) {
-            ret.emplace_back(tet.tuple());
+        const wmtk::internal::SimplexSet& tets = sc.get_simplices(PrimitiveType::Tetrahedron);
+        if (!mesh().is_boundary(mesh().switch_face(
+                mesh().switch_edge(mesh().switch_tetrahedron(m_output_tuple))))) {
+            for (const Simplex& tet : tets) {
+                ret.emplace_back(tet.tuple());
+            }
+            ret.emplace_back(mesh().switch_tetrahedron(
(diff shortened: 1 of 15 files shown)


Zhouyuan-Chen:
Just want to mention that seems Boundary Detect in 3D is not implemented. In this case, I didn't use the related invariant.

src/wmtk/operations/tet_mesh/TetSplit.cpp
```diff
@@ -0,0 +1,91 @@
...
+    : TetMeshOperation(m)
+    , TupleOperation(settings.invariants, t)
+    , m_settings{settings}
+{
+    assert(m_settings.are_invariants_initialized());
+}
+
+bool TetSplit::execute()
+{
+    Tuple first_split_ret = mesh().split_edge(input_tuple(), hash_accessor()).m_output_tuple;
```
daniel-zint, line 40: You should not use atomic operations in here but `TetEdgeSplit` and `TetEdgeCollapse`.

Also, check the naming convention from the tri_mesh folder. `TetEdgeSplit` should only be called `EdgeSplit` and `TetEdgeCollapse` should be `EdgeCollapse`.

src/wmtk/operations/tet_mesh/TetSplit.cpp
```diff
@@ -0,0 +1,91 @@
...
+}
+
+bool OperationSettings<tet_mesh::TetSplit>::are_invariants_initialized() const
+{
+    return find_invariants_in_collection_by_type<ValidTupleInvariant>(invariants);
+}
+
+namespace tet_mesh {
+
+// TetSplit::TetSplit(Mesh& m, const Tuple& t, const OperationSettings<TetSplit>& settings)
```
daniel-zint, line 22: Delete commented out code.

src/wmtk/operations/tet_mesh/TetSplit.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+Tuple TetSplit::return_tuple() const
+{
+    return m_output_tuple;
+}
+
+std::vector<Tuple> TetSplit::modified_primitives(PrimitiveType type) const
+{
+    if (type == PrimitiveType::Face) {
+        // TODO
```
daniel-zint, line 83: if this is really a TODO, you want this code to throw an error if someone tries to run it.

daniel-zint (changes requested):
looks almost good

src/wmtk/operations/tet_mesh/TetSplit.cpp
Zhouyuan-Chen: I copied this from tri_mesh. I am not sure what is this supposed to do. But definitely, we need this function...

daniel-zint: Don't copy code without testing it and without knowing what it does. That can lead to very annoying debugging sessions later on. It is better if the code crashes or contains only half the functionality than having code that appears to the outside like it is looking the right thing.

Zhouyuan-Chen:
okay, I implemented the primitive_modified function

Zhouyuan-Chen:
I added the tests for the function primitive_modified.

daniel-zint (approved):
lgtm

PR #500, opened 2023-11-06
add link condition for edge mesh

zlyfunction:
add link cond for edge mesh:
check if the input edge tuple is the only edge in mesh

### DIFF
--- a/src/wmtk/SimplicialComplex.cpp
+++ b/src/wmtk/SimplicialComplex.cpp
@@ -8,8 +8,8 @@
 #include "simplex/simplex_boundary.hpp"
 #include "simplex/top_dimension_cofaces.hpp"
 
+#include "EdgeMesh.hpp"
 #include "TriMesh.hpp"
-
 namespace wmtk {
 
 internal::SimplexSet SimplicialComplex::get_simplices(const PrimitiveType& ptype) const
@@ -149,6 +149,7 @@ bool SimplicialComplex::link_cond(const Mesh& m, Tuple t)
 // work for 2-manifold case only for now
 bool SimplicialComplex::link_cond_bd_2d(const Mesh& m, Tuple t)
 {
+    assert(m.top_cell_dimension() == 2);
     // step1 check normal link condition
     if (!link_cond(m, t)) {
         return false;
@@ -201,6 +202,20 @@ bool SimplicialComplex::link_cond_bd_2d(const Mesh& m, Tuple t)
     return true;
 }
 
+bool SimplicialComplex::link_cond_bd_1d(const Mesh& m, Tuple t)
+{
+    assert(m.top_cell_dimension() == 1);
+    const Tuple t_switch_v = m.switch_tuple(t, PrimitiveType::Vertex);
+    if (m.is_boundary(t) && m.is_boundary(t_switch_v)) {
+        return false;
+    }
+    if (m.simplices_are_equal(
+            Simplex(PrimitiveType::Vertex, t),
+            Simplex(PrimitiveType::Vertex, t_switch_v))) {
+        return false;
+    }
+    return true;
+}
 
 // Toplogical-holding condition, not necessarily guarantee geometric embedding
 bool SimplicialComplex::edge_collapse_possible_2d(const TriMesh& m, const Tuple& t)
--- a/src/wmtk/SimplicialComplex.hpp
+++ b/src/wmtk/SimplicialComplex.hpp
@@ -149,6 +149,13 @@ public:
     static bool link_cond(const Mesh& m, Tuple t);
     static bool link_cond_bd_2d(const Mesh& m, Tuple t);
 
+    //////////////////////////////////
+    // link condition for edge mesh
+    // input Tuple t ---> edge (a,b)
+    // check if edge(a,b) is the only edge in EdgeMesh m
+    //////////////////////////////////
+    static bool link_cond_bd_1d(const Mesh& m, Tuple t);
+
     // could be a replacement for link_cond_bd_2d
     static bool edge_collapse_possible_2d(const TriMesh& m, const Tuple& t);
     //////////////////////////////////
--- a/src/wmtk/simplex/SimplexCollection.cpp
+++ b/src/wmtk/simplex/SimplexCollection.cpp
@@ -103,4 +103,15 @@ SimplexCollection SimplexCollection::get_intersection(
     return sc;
 }
 
+bool SimplexCollection::are_simplex_collections_equal(
+    const SimplexCollection& collection_a,
+    const SimplexCollection& collection_b)
+{
+    if (collection_a.m_simplices.size() != collection_b.m_simplices.size()) {
+        return false;
+    }
+    SimplexCollection sc_union = SimplexCollection::get_union(collection_a, collection_b);
+    return sc_union.m_simplices.size() == collection_a.m_simplices.size();
+}
+
 } // namespace wmtk::simplex
--- a/src/wmtk/simplex/link_condition.cpp
+++ b/src/wmtk/simplex/link_condition.cpp
@@ -1,0 +1,30 @@
+#include "link_condition.hpp"
+#include "link.hpp"
+
+namespace wmtk::simplex {
+bool link_condition_closed_trimesh(const TriMesh& mesh, const Tuple& edge)
+{
+    SimplexCollection lnk_a = link(mesh, Simplex(PrimitiveType::Vertex, edge), false); // link(a)
+    SimplexCollection lnk_b = link(
+        mesh,
+        Simplex(PrimitiveType::Vertex, mesh.switch_tuple(edge, PrimitiveType::Vertex)),
+        false); // link(b)
+    return true;
+    // return collection.simplex_vector().empty();
+}
+
+bool link_condition(const EdgeMesh& mesh, const Tuple& edge)
+{
+    // SimplexCollection collection = link(mesh, simplex, false);
+    return true;
+    // return collection.simplex_vector().empty();
+}
+
+bool link_condition(const TriMesh& mesh, const Tuple& edge)
+{
+    // SimplexCollection collection = link(mesh, simplex, false);
+    return true;
+    // return collection.simplex_vector().empty();
+}
+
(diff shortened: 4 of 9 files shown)


mtao (dismissed):
mostly fine, i just want to see the files moved around to better locations. might as well move the 2d link condition as well.

src/wmtk/SimplicialComplex.cpp
```diff
@@ -201,6 +202,20 @@ bool SimplicialComplex::link_cond_bd_2d(const Mesh& m, Tuple t)
     return true;
 }
 
+bool SimplicialComplex::link_cond_bd_1d(const Mesh& m, Tuple t)
+{
+    assert(m.top_cell_dimension() == 1);
```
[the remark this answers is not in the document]
daniel-zint: I marked that conversation as "unresolved". At least I don't see how it was resolved.

src/wmtk/SimplicialComplex.hpp
```diff
@@ -149,6 +149,13 @@ class SimplicialComplex
     static bool link_cond(const Mesh& m, Tuple t);
     static bool link_cond_bd_2d(const Mesh& m, Tuple t);
 
+    //////////////////////////////////
+    // link condition for edge mesh
+    // input Tuple t ---> edge (a,b)
+    // check if edge(a,b) is the only edge in EdgeMesh m
+    //////////////////////////////////
+    static bool link_cond_bd_1d(const Mesh& m, Tuple t);
```
daniel-zint, line 157: I like it a lot that you are adding documentation. Please use the doxygen style:
```
/**
* @brief short explanation
* long explanation if necessary
*
* @param m explain parameter "m"
* @param whatever explain parameter "whatever"
* @return true if blabla, false otherwise (or something like that)
*/
```

You can find the full documentation on Doxygen documentation [here](https://www.doxygen.nl/manual/docblocks.html)

src/wmtk/simplex/SimplexCollection.cpp
```diff
@@ -103,4 +103,15 @@ SimplexCollection SimplexCollection::get_intersection(
...
 }
 
+bool SimplexCollection::are_simplex_collections_equal(
+    const SimplexCollection& collection_a,
+    const SimplexCollection& collection_b)
+{
+    if (collection_a.m_simplices.size() != collection_b.m_simplices.size()) {
+        return false;
+    }
+    SimplexCollection sc_union = SimplexCollection::get_union(collection_a, collection_b);
```
daniel-zint, line 113: I think you need to use intersection here instead of union. Otherwise `collection_a` could be a subset of `collection_b`.

Please also add a test for `are_simplex_collections_equal`.

src/wmtk/simplex/link_condition.cpp
```diff
@@ -0,0 +1,90 @@
+#include "link_condition.hpp"
+#include "link.hpp"
+#include "open_star.hpp"
+
+namespace wmtk::simplex {
+bool link_condition_closed_trimesh(const TriMesh& mesh, const Tuple& edge)
+{
+    SimplexCollection lnk_a = link(mesh, Simplex(PrimitiveType::Vertex, edge), true); // link(a)
```
daniel-zint, line 8: * `const`
* Just write `link_a` instead of `lnk_a`.
* You don't need to explicitly write `true` here. The default is already true.
* If you write something like `const Simplex v_a = Simplex(PrimitiveType::Vertex, edge)` you can shorten down this line by a lot and make it much more readable. The same holds for `v_b` and `e`.

src/wmtk/simplex/link_condition.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <wmtk/EdgeMesh.hpp>
+#include <wmtk/TriMesh.hpp>
+#include "SimplexCollection.hpp"
+
+namespace wmtk::simplex {
+bool link_condition(const EdgeMesh& mesh, const Tuple& edge);
```
daniel-zint, line 8: Please add documentation.

src/wmtk/simplex/link_condition.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    SimplexCollection lnk_a_lnk_b_intersection = SimplexCollection::get_intersection(lnk_a, lnk_b);
+
+    return SimplexCollection::are_simplex_collections_equal(lnk_a_lnk_b_intersection, lnk_ab);
+}
+
+bool link_condition(const EdgeMesh& mesh, const Tuple& edge)
+{
+    const Tuple edge_switch_v = mesh.switch_tuple(edge, PrimitiveType::Vertex);
+    if (mesh.is_boundary_vertex(edge) && mesh.is_boundary_vertex(edge_switch_v)) {
```
daniel-zint, line 23: I feel like this should not be part of the link condition. I already mentioned it yesterday when we talked. Let's discuss that tomorrow during our group meeting. Just keep it here for now but I feel like it needs to be moved to some other place.

src/wmtk/simplex/link_condition.cpp
```diff
@@ -0,0 +1,90 @@
...
+        return false;
+    }
+
+    // check if dummy vertex w is included in the lhs
+    auto get_boundary_edges = [&mesh](const Tuple& _v) {
+        Simplex input_v(PrimitiveType::Vertex, _v);
+        std::vector<Tuple> ret;
+        // get one_ring_edges from open_star
+        auto one_ring_edges = open_star(mesh, input_v).simplex_vector(PrimitiveType::Edge);
+        for (const auto& _e : one_ring_edges) {
```
daniel-zint, line 47: `Simplex` not `auto`

src/wmtk/simplex/link_condition.cpp
```diff
@@ -0,0 +1,90 @@
...
+}
+
+bool link_condition(const TriMesh& mesh, const Tuple& edge)
+{
+    // step1 check link condition for closed case
+    if (!link_condition_closed_trimesh(mesh, edge)) {
+        return false;
+    }
+
+    // check if dummy vertex w is included in the lhs
```
daniel-zint, line 41: What is "dummy vertex w"?

daniel-zint (changes requested):
I didn't go through the tests. I think the readability of the code can be improved. Also, there is some code that I just couldn't understand which either means you need better variable names or you need comments explaining why you are doing something.

zlyfunction:
It’s a notation in this paper http://elib.mi.sanu.ac.rs/files/journals/publ/86/n080p023.pdf

src/wmtk/simplex/link_condition.cpp
zlyfunction: It’s a notation in this paper http://elib.mi.sanu.ac.rs/files/journals/publ/86/n080p023.pdf

src/wmtk/simplex/SimplexCollection.cpp
zlyfunction: It cannot be. I already checked the length before

src/wmtk/SimplicialComplex.cpp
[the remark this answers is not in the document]
zlyfunction: I have added new files to the simplex directory. The new link condition is defined within the simplex_collection class. To avoid confusion, I have use link_condition and link_condition_for_closed_trimesh instead.

src/wmtk/simplex/link_condition.cpp
daniel-zint: I see. However, the code should be understandable even without reading a paper. You can still add a hint in the documentation of the method that this function implements the paper but reading the paper should not be required to understand the code.

src/wmtk/simplex/link_condition.cpp
```diff
@@ -0,0 +1,90 @@
...
+    auto get_boundary_edges = [&mesh](const Tuple& _v) {
+        Simplex input_v(PrimitiveType::Vertex, _v);
+        std::vector<Tuple> ret;
+        // get one_ring_edges from open_star
+        auto one_ring_edges = open_star(mesh, input_v).simplex_vector(PrimitiveType::Edge);
```
daniel-zint, line 46: These are not the one-ring edges by the way. These are the incident edges. One-ring edges are the edges in the link.

src/wmtk/simplex/link_condition.cpp
daniel-zint: We will keep that as it is for now and probably later on create a second more simple "link condition" (or whatever it will be called).

src/wmtk/simplex/link_condition.cpp
```diff
@@ -5,16 +5,18 @@
 namespace wmtk::simplex {
 bool link_condition_closed_trimesh(const TriMesh& mesh, const Tuple& edge)
 {
-    SimplexCollection lnk_a = link(mesh, Simplex(PrimitiveType::Vertex, edge), true); // link(a)
-    SimplexCollection lnk_b = link(
-        mesh,
-        Simplex(PrimitiveType::Vertex, mesh.switch_tuple(edge, PrimitiveType::Vertex)),
-        true); // link(b)
-    SimplexCollection lnk_ab = link(mesh, Simplex(PrimitiveType::Edge, edge), true); // link(ab)
+    const Simplex v_a = Simplex(PrimitiveType::Vertex, edge);
```
daniel-zint, line 8: You can use `Simplex::vertex(edge)` and `Simplex::edge(edge)`. That's a bit shorter.

tests/test_simplex_collection.cpp
```diff
@@ -1233,4 +1233,47 @@ TEST_CASE("simplex_link_condtion_trimesh", "[simplex_collection]")
         REQUIRE(link_condition(m, t2) == false);
         REQUIRE(link_condition(m, t3) == true);
     }
+}
+
+TEST_CASE("are_simplex_collections_equal", "[simplex_collection]")
```
daniel-zint, line 1238: The test is not sufficient but it is fine for now. Please write a TODO that this test should be extended to cases where the two sets are not equal. I am very sure that the code works correctly but it would be good to have full coverage of all possible cases.

daniel-zint (approved):
Should be fine.

PR #518, opened 2023-11-13
create tags for multimesh construction

yunfanzhou:
a generic implementation in multimesh::untils to create tags for edges. The tags will be used for multimesh construction


src/wmtk/multimesh/utils/create_tag.cpp
```diff
@@ -0,0 +1,57 @@
+#include "create_tag.hpp"
+
+namespace wmtk::multimesh::utils {
+bool is_root(std::vector<int>& parents, int vid)
+{
+    return parents[vid] == vid;
```
mtao, line 6: Const

src/wmtk/multimesh/utils/create_tag.cpp
```diff
@@ -0,0 +1,57 @@
...
+}
+
+int get_root(std::vector<int>& parents, int vid)
+{
+    if (parents[vid] == -1) {
```
mtao, line 11: This case wouldn't happen if the system is initialized properly

src/wmtk/multimesh/utils/create_tag.hpp
```diff
@@ -0,0 +1,19 @@
...
+bool is_root(std::vector<int>& parents, int vid);
+
+std::vector<int> create_parents(std::vector<Tuple>& tuples);
+
+void set_root(std::vector<int>& parents, int vid, int root);
+int get_root(std::vector<int>& parents, int vid);
+
+void set_union(std::vector<int>& parents, int vid1, int vid2);
+
+void union_find(std::set<int>& critical_points, std::vector<int>& parents, int vid1, int vid2);
```
mtao, line 18: None of these functions create the tag/they don't need to be in this header

mtao (commented):
Looks ok so far

mtao (changes requested):
API should be that the TupleTag class is an "invisible" API called in tuple_tag.cpp, not something that is constructed and invoked publicly (visible in the header)

mtao (approved):
lgtm

PR #521, opened 2023-11-15
503 multimesh remeshing

JcDai:
the multimesh version of isotropic remeshing. Aiming to test the multimesh invariants and operations


src/wmtk/MultiMeshManager.cpp
```diff
@@ -479,12 +479,19 @@ void MultiMeshManager::update_map_tuple_hashes(
...
+            } else {
+                is_gid_visited[original_parent_gid] = true;
+            }
             const char parent_flag = Mesh::get_index_access(parent_flag_accessor)
                                          .const_scalar_attribute(original_parent_gid);
             bool exists = 1 == (parent_flag & 1);
             if (!exists) {
-                continue;
+                // It's okay if it does not exist, this is important for the edge_collapse case
+                // where ear edge might be replaced by a new edge continue;
```
daniel-zint, line 494: I guess the `continue;` should not be in the comment.

tests/components/test_component_isotropic_remeshing.cpp
```diff
@@ -657,3 +658,43 @@ TEST_CASE("remeshing_with_boundary", "[components][isotropic_remeshing][2D]")
         // mesh.serialize(writer);
     }
 }
+
+TEST_CASE("remeshing_preserve_topology", "[components][isotropic_remeshing][2D]")
```
daniel-zint, line 662: A test case that is under development can be removed from the CI by adding `[.]` to the tags. Please do it because currently, your tests block our Windows test machines as they run into an endless loop.

tests/test_multi_mesh.cpp
```diff
@@ -25,6 +25,30 @@ constexpr PrimitiveType PF = PrimitiveType::Face;
 
 namespace {} // namespace
 
+void print_tuple_map(const DEBUG_TriMesh& parent, const DEBUG_MultiMeshManager& p_mul_manager)
```
daniel-zint, line 28: That looks like debug code. Please add a comment stating that and use `wmtk::logger().debug()` for that.

daniel-zint (commented):
This is not a full review. Just a few comments.

src/wmtk/Mesh.hpp
```diff
@@ -7,6 +7,7 @@
 #include <memory>
 #include <wmtk/io/ParaviewWriter.hpp>
 #include <wmtk/multimesh/utils/extract_child_mesh_from_tag.hpp>
+#include <wmtk/operations/utils/UpdateVertexMultiMeshMapHash.hpp>
```
mtao, line 10: this should not be included here - at most a predeclaration

src/wmtk/Mesh.hpp
```diff
@@ -57,6 +59,8 @@ class TupleTag;
 }
 } // namespace multimesh
 
+class SimplicialComplex;
```
mtao, line 62: we're trying to deprecate this class - why is it being added to more things?

src/wmtk/invariants/MultiMeshTopologyInvariant.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include <wmtk/Mesh.hpp>
+#include <wmtk/PointMesh.hpp>
+#include <wmtk/SimplicialComplex.hpp>
+#include <wmtk/TetMesh.hpp>
+#include <wmtk/TriMesh.hpp>
+#include <wmtk/multimesh/MultiMeshVisitor.hpp>
+#include <wmtk/simplex/Simplex.hpp>
+
+namespace wmtk {
+MultiMeshEdgeTopologyInvariant::MultiMeshEdgeTopologyInvariant(
```
mtao, line 14: write unit tests to improve coverage here...

src/wmtk/invariants/MultiMeshTopologyInvariant.cpp
```diff
@@ -0,0 +1,40 @@
...
+    const Tuple v1 = t;
+    const Tuple v2 = mesh().switch_vertex(t);
+
+    // if less or equal than one vertex is in the child mesh, return true
+    if (mesh().map_to_child_tuples(m_child_mesh, Simplex(PrimitiveType::Vertex, v1)).size() == 0 ||
+        mesh().map_to_child_tuples(m_child_mesh, Simplex(PrimitiveType::Vertex, v2)).size() == 0)
+        return true;
+
+    // if both vertices are in the child mesh, and the edge is not in the child mesh, return false
+    if (mesh().map_to_child_tuples(m_child_mesh, Simplex(PrimitiveType::Edge, t)).size() == 0)
```
mtao, line 35: checking for the edge is more efficient than checkign for the vertex - so check for that first as a shortcut. also you can use the `empty()` function. perhaps it'd be worth adding a `has_map` function which is much cheaper than trying the actual map

src/wmtk/invariants/MultiMeshTopologyInvariant.hpp
```diff
@@ -0,0 +1,23 @@
+#include "MeshInvariant.hpp"
+
+namespace wmtk {
+class MultiMeshEdgeTopologyInvariant : public MeshInvariant
+{
+public:
+    MultiMeshEdgeTopologyInvariant(const Mesh& parent, const Mesh& child, const PrimitiveType pt);
```
mtao, line 7: if the child mesh must be an EdgeMesh why not check for that explicitly instead of passing a generic `Mesh` ref?

src/wmtk/operations/utils/UpdateEdgeOperationMultiMeshMapFunctor.cpp
```diff
@@ -32,6 +32,68 @@ void UpdateEdgeOperationMultiMeshMapFunctor::update_all_hashes(
...
+                    wmtk::multimesh::utils::vector5_to_tuple(parent_to_child_data.tail<5>());
+                parent_tuple = m.resurrect_tuple(parent_tuple, parent_hash_accessor);
+                child_tuple = child_ptr->resurrect_tuple(child_tuple, child_hash_accessor);
+
+                // TODO: Currently using global ids to identify the tuple
+                Tuple opt_tuple =
+                    m.tuple_from_global_ids(ear_fid_other, ears[1].eid, fmoe.m_spine_vids[1]);
+
+                if (m.id_vertex(parent_tuple) == ear_vid) {
+                    wmtk::multimesh::utils::symmetric_write_tuple_map_attributes(
```
mtao, line 79: make some test that does this pathway

src/wmtk/operations/utils/UpdateVertexMultiMeshMapHash.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <wmtk/EdgeMesh.hpp>
+#include <wmtk/Mesh.hpp>
+#include <wmtk/PointMesh.hpp>
+#include <wmtk/SimplicialComplex.hpp>
+#include <wmtk/TetMesh.hpp>
+#include <wmtk/TriMesh.hpp>
+#include <wmtk/multimesh/utils/tuple_map_attribute_io.hpp>
+
+namespace wmtk::operations::utils {
+void update_vertex_operation_hashes(Mesh& m, const Tuple& vertex, Accessor<long>& hash_accessor)
```
mtao, line 11: why not implement this as a static member function of the MultiMeshManager class to avoid having to friend this?

tests/components/test_component_isotropic_remeshing.cpp
```diff
@@ -18,12 +19,51 @@
...
+#include "../tools/EdgeMesh_examples.hpp"
+#include "../tools/TriMesh_examples.hpp"
+
 using json = nlohmann::json;
 using namespace wmtk;
 using namespace wmtk::tests;
 
 const std::filesystem::path data_dir = WMTK_DATA_DIR;
 
+void print_tuple_map_iso(const DEBUG_TriMesh& parent, const DEBUG_MultiMeshManager& p_mul_manager)
```
mtao, line 41: just write out as print tuple map without hte iso (use ana anonymous namespace?)

src/wmtk/Mesh.hpp
JcDai: Vertexsmoothing is using simplicial complex so I just use it to the new function to adopt it. Can I add an issue and change it to simplexcollection later?

src/wmtk/operations/utils/UpdateVertexMultiMeshMapHash.cpp
JcDai: updated, I made them a static member functions of multimesh manager and create a wrapper function for them in Mesh class.

PR #532, opened 2023-11-20
Embedded remeshing pipline

Zhouyuan-Chen:
This is the embedded remeshing pipeline.


Zhouyuan-Chen:
2D 3D offset relocation is done. we should discuss the energy relocation later maybe...

Zhouyuan-Chen:
I used a dumb way to do the is_invert function in the 3D case, I just checked the volume of the neighbor tethedra before and after the operation, if the sign of the column changes, then it is inverted.

Therefore, I need to guarantee that the initial state is uninverted.

Zhouyuan-Chen:
I still need to add some tests for the atomic operation such as pushing and smoothing.

Zhouyuan-Chen:
waiting for the remeshing component

Zhouyuan-Chen:
added and tested the Taubin smoothing.

PR #539, opened 2023-11-27
Delete `tuple_vector` from `SimplexCollection`

mtao:
This function returns tuples from a heterogeneous collection of simplices, which does not have a clear semantic and therefore should not exist.
If a function needs a collection of Tuple objects then it should simply return a `vector<Tuple>`.


daniel-zint (approved):
lgtm

PR #540, opened 2023-11-27
future proofing `primitive_type_name`

mtao:
Added some specifications on how `primitive_type_name operates to make sure that we only pass in valid arguments even if the number of primitives change.

The main motivation for this was that gcc 13 on my personal laptop was unhappy with the degree in which we were constraining accesses into the local `names` array so I added some extra specification to make sure that it would build.


daniel-zint (approved):
lgtm but you might want to add a test for `primitive_type_name`

rjc8237:
Looks fine to me as well.

I had some concerns about the computation of the number of ids as it depends on the use of an array of strings, but I suppose that is unlikely to be changed.

PR #542, opened 2023-11-28
541 fixing and testing multimap with bijections

mtao:
the default map when two meshes are bijective in a trivial way was not tested / used. Fixed an issue where reserve was used but resizing never happened in the vector. also changed the name from surjection to bijection, changed asserts to throw so catch2 can check that the exceptions do the desired things.


yunfanzhou (commented):
lgtm

PR #546, opened 2023-11-29
544 default cache and subcaching

mtao:
Thought is to enable a workflow like

```
CacheStack stack("opt",".",false);
for(int iter = 0; iter < 100; ++iter) {
   SubCacheHandle h(fmt::format("{:04}", iter), stack);
   Cache c = stack.get_current_stack();
  // all files created by c are put in the "path/0002_unique_stuff/..."
}
```


daniel-zint:
When is it advantageous to use subcaching? I don't quite see the use case for that.

PR #559, opened 2023-12-03
MultiMesh Invariants

mtao:
In order for invariants to work with multimesh we need to be able to map tuples between invariants.
This means that invariantcollection now needs a concept of mesh(), which means that all of the default initialized settings need to have a constructor fora mesh().


src/wmtk/operations/tet_mesh/TetSplitWithTags.hpp
```diff
@@ -19,8 +19,10 @@ struct OperationSettings<tet_mesh::TetSplitWithTags>
     MeshAttributeHandle<double> pos_handle;
     long split_vertex_tag_value;
 
+    OperationSettings(const Mesh& m): invariants(m) {}
+
     InvariantCollection invariants;
-    void initialize_invariants(const TetMesh& m);
+    void initialize_invariants();
```
mtao, line 25: initializing invariants no longer takes an arg as it should read the mesh() from the invariantcollection's mesh held internally

src/wmtk/operations/tet_mesh/TetSplitWithTags.hpp
```diff
@@ -19,8 +19,10 @@ struct OperationSettings<tet_mesh::TetSplitWithTags>
     MeshAttributeHandle<double> pos_handle;
     long split_vertex_tag_value;
 
+    OperationSettings(const Mesh& m): invariants(m) {}
```
mtao, line 22: the operationsettings objects now need to default to default initialize the invariants with a mesh - lets just put this in headers for now to be lazy

src/wmtk/operations/tet_mesh/TetSplitWithTags.cpp
```diff
@@ -53,8 +54,8 @@ bool TetSplitWithTags::execute()
         opposite_tt = acc_tt.scalar_attribute(mesh().switch_tetrahedron(input_tuple()));
     }
 
-    OperationSettings<TetSplit> op_settings;
-    op_settings.initialize_invariants(mesh());
+    OperationSettings<TetSplit> op_settings(mesh());
```
mtao, line 57: example of how compound op settings have to be done now

src/wmtk/operations/tet_mesh/TetSplitWithTags.cpp
```diff
@@ -8,8 +8,9 @@
 
 namespace wmtk::operations {
 
-void OperationSettings<tet_mesh::TetSplitWithTags>::initialize_invariants(const TetMesh& m)
+void OperationSettings<tet_mesh::TetSplitWithTags>::initialize_invariants()
 {
+    const TetMesh& m = static_cast<TetMesh&>(invariants.mesh());
```
mtao, line 13: here's how we should read the mesh from the held invariantcollection from now on

src/wmtk/operations/tet_mesh/TetSplitWithTags.hpp
```diff
@@ -19,8 +19,10 @@ struct OperationSettings<tet_mesh::TetSplitWithTags>
...
+
     InvariantCollection invariants;
```
mtao, line 24: to be honest the cleanest thing to do is to make the InvariantCollection held here a shared_ptr. If you don't mind going through that work in every operation then all of this constructor changing stuff changes.

All you need to do is make sure the invariant-collection code everywhere creates a shared_ptr (basic_invariatn_collection or whatnot should return a shared_ptr now), and then you cahnage a bunch of dot notation to arrow notation

src/wmtk/operations/tet_mesh/TetSplitWithTags.hpp
mtao: (dont do this if you want to use shared_ptr)

src/wmtk/operations/tet_mesh/TetSplitWithTags.hpp
mtao: if you use shared_ptr then just create the invariantcollection using make_Shared and then add things as we currently are doing

src/wmtk/operations/tet_mesh/TetSplitWithTags.cpp
mtao: (dont do this if you want to use shared_ptr)

mtao (commented):
@yunfanzhou please look at these changes and the comments i made.
I started out thinking about keeping the held `InvariantCollection` as values, but in retrospect i think tehse should all be shared_ptr to invariant collections. please do that

yunfanzhou:
Who append InvariantsC from setting to the OperationFactory’ InvCol? 
OperationSettings now has to have a default InvCol?
OperationFac owns a copy of InvCol. Every OperaSetting implemenet function that create InvarCol and return a shared_ptr to the InvarC. 
TupleOperation.before() checks tuple validity in the Operation.mesh() yes
OperationFactory still construct using a setting? yes
OperationSettings.create_invariants() should take a mesh as input argument?
OpSetting store a ptr to any mesh that’s required. create_invaraints() in OpSetting will return a InvCol that contains all the invariants that’s constructed on the given meshes

daniel-zint:
I did the majority of refactoring but some things do not work or need to be changed:
1. Not all tests pass. I could fix some things but for the rest I am a bit clueless right now.
2. There is still the static mesh hack in the code that needs to be removed.
3. Tests for multi mesh invariants are missing. We need more invariants tests in general.
4. The `after` function was not updated to function with multi mesh.
5. We need to check in general that the behavior of the code is now as desired.
6. `TupleOperation` needs a bit of refactoring because now it is actually a `SimplexOperation`. It also has a function `input_tuple()` which should now be `input_simplex()`.

I think we should address these problems in the written order.

PR #563, opened 2023-12-06
Dzint/simplex face optimization

daniel-zint:
- make `faces_single_dimension` return a `SimplexCollection` and add a `_tuples` version which returns a vector of tuples
- rewrite `simplex::faces` without recursion
- overload `simplex::open_star` with `TriMesh` specific version


mtao (approved):
awesome!

PR #565, opened 2023-12-06
Multimesh Attribute Scoping

mtao:
Added a multimesh attribute scope handle - it works by collecting a vector of a single-mesh attribute scope handles.


mtao:
@JcDai @zlyfunction could you provide a test / tell me how to run something taht was causing splits to fail in the past? it'd be good to test that this fixes the issue you two were facing

PR #577, opened 2023-12-13
add link condition for 3d

JcDai:
added 3d link condition to simplex/link_condition. added tests for it


src/wmtk/simplex/link_condition.cpp
```diff
@@ -98,4 +98,93 @@ bool link_condition(const TriMesh& mesh, const Tuple& edge)
     return true;
 }
 
+bool link_condition_closed_tetmesh(const TetMesh& mesh, const Tuple& edge)
+{
+    // for closed mesh, if link(a) \intersect link(b) == link(ab)
+    const Simplex v_a = Simplex::vertex(edge);
+    const Simplex v_b = Simplex::vertex(mesh.switch_tuple(edge, PrimitiveType::Vertex));
```
daniel-zint, line 105: `mesh.switch_vertex(edge)`

src/wmtk/simplex/link_condition.cpp
```diff
@@ -98,4 +98,93 @@ bool link_condition(const TriMesh& mesh, const Tuple& edge)
...
+    // if ab is a boundary edge, then if any boundary face incident a shares a common edge with any
+    // boundary face incident b, return false.
+    // if ab is not boundary edge, then if a is not on
+    // boundary or b is not on boundary, return true, otherwise false
+
+    auto get_boundary_faces = [&mesh](const Tuple& _v) {
+        const Simplex input_v(PrimitiveType::Vertex, _v);
+        std::vector<Tuple> ret;
+        // get incident_faces from open_star
+        auto incident_faces = open_star(mesh, input_v).simplex_vector(PrimitiveType::Face);
```
daniel-zint, line 135: you should use `cofaces_single_dimension` for that instead of `open_star`

src/wmtk/simplex/link_condition.cpp
```diff
@@ -98,4 +98,93 @@ bool link_condition(const TriMesh& mesh, const Tuple& edge)
...
+        for (const Simplex& _f : incident_faces) {
+            if (mesh.is_boundary(_f.tuple(), PrimitiveType::Face)) {
+                if (utils::SimplexComparisons::equal(
+                        mesh,
+                        Simplex(PrimitiveType::Vertex, _f.tuple()),
+                        input_v)) {
+                    // tuple edge is _v - x
+                    ret.push_back(mesh.switch_edge(mesh.switch_vertex(_f.tuple())));
+                } else {
+                    if ((utils::SimplexComparisons::equal(
```
daniel-zint, line 145: you can write that whole stuff as
```
if() {
} else if() {
} else {
}
```

For whatever reason you nested the second if-statement inside the else case. That gives you an unnecessary level of indention.

Additionally, this whole if-else stuff should not be necessary if you use `cofaces_single_dimension` because I think it guarantees that the tuple in there contains the input simplex. But check the documentation once more to make sure what I am saying is correct. And you can at least add an assert to your code here to be certain.

src/wmtk/simplex/link_condition.cpp
```diff
@@ -98,4 +98,93 @@ bool link_condition(const TriMesh& mesh, const Tuple& edge)
...
+            }
+        }
+        return ret;
+    };
+
+    const auto boundary_neighbors_a = get_boundary_faces(edge);
+    const auto boundary_neighbors_b =
+        get_boundary_faces(mesh.switch_tuple(edge, PrimitiveType::Vertex));
+    if (mesh.is_boundary_edge(edge)) {
+        for (auto f_a : boundary_neighbors_a) {
```
daniel-zint, line 165: `const Simplex &` in both for-loops

src/wmtk/simplex/link_condition.cpp
```diff
@@ -98,4 +98,93 @@ bool link_condition(const TriMesh& mesh, const Tuple& edge)
...
+    }
+
+    // add dummy vertex and link it to boundary faces
+    // the check can be convert from link_dummy(a) \intersect link_dummy(b) == link_dummy(ab) to:
+    // if ab is a boundary edge, then if any boundary face incident a shares a common edge with any
+    // boundary face incident b, return false.
+    // if ab is not boundary edge, then if a is not on
+    // boundary or b is not on boundary, return true, otherwise false
+
+    auto get_boundary_faces = [&mesh](const Tuple& _v) {
```
daniel-zint, line 131: I think the code is easier to read if you make that input a `Simplex`. The invocation of that one confused me.

src/wmtk/simplex/link_condition.cpp
```diff
@@ -98,4 +98,93 @@ bool link_condition(const TriMesh& mesh, const Tuple& edge)
...
+
+    const auto boundary_neighbors_a = get_boundary_faces(edge);
+    const auto boundary_neighbors_b =
+        get_boundary_faces(mesh.switch_tuple(edge, PrimitiveType::Vertex));
+    if (mesh.is_boundary_edge(edge)) {
+        for (auto f_a : boundary_neighbors_a) {
+            for (auto f_b : boundary_neighbors_b) {
+                if (utils::SimplexComparisons::equal(
+                        mesh,
+                        Simplex(PrimitiveType::Edge, f_a),
```
daniel-zint, line 169: I am very confused. In `get_boundary_faces` you are collecting triangles, right? Then why do you now consider the triangle here as an edge?
I think you should use `Simplex` more often to make the code easier to read. I cannot really follow the reasoning of that code right now.

tests/test_simplex_collection.cpp
```diff
@@ -1357,3 +1357,144 @@ TEST_CASE("are_simplex_collections_equal", "[simplex_collection]")
...
+        std::array<Tuple, 6> e;
+        // cannot collapse
+        e[0] = m.edge_tuple_from_vids(0, 1);
+        e[1] = m.edge_tuple_from_vids(0, 2);
+        e[2] = m.edge_tuple_from_vids(0, 3);
+        e[3] = m.edge_tuple_from_vids(1, 2);
+        e[4] = m.edge_tuple_from_vids(1, 3);
+        e[5] = m.edge_tuple_from_vids(2, 3);
+
+        for (int i = 0; i < 6; ++i) {
```
daniel-zint, line 1377: `size_t`
Or even better, use a for-each loop: `for(const Tuple& t : e)`

tests/test_simplex_collection.cpp
```diff
@@ -1357,3 +1357,144 @@ TEST_CASE("are_simplex_collections_equal", "[simplex_collection]")
...
+            REQUIRE(link_condition(m, e[i]) == false);
```
daniel-zint, line 1378: `CHECK` instead of `REQUIRE` also in the rest of the test.

Instead of writing `... == false` you can either write `!...` or `CHECK_FALSE(...)`

daniel-zint (approved):
Test looks solid so I assume the implementation works. The comments are mostly on coding style.

src/wmtk/simplex/link_condition.cpp
JcDai: The reason is I want to check: after adding a dummy vertex and connecting it to all of the vertices in the incident boundary faces of v_a and v_b, do link(a) and link(b) share new common faces (one of the endpoint should be the dummy vertex). So the real check is to check if any face from boundary_neighbors_a shares a common edge with any face from boundary_neighbors_b, because if they shares an edge, there will be a common face consisting of this edge and the dummy vertex sharing between link(a) and link(b), which does not belongs to link(ab).

src/wmtk/simplex/link_condition.cpp
```diff
@@ -141,17 +142,15 @@ bool link_condition(const TetMesh& mesh, const Tuple& edge)
                         input_v)) {
                     // tuple edge is _v - x
                     ret.push_back(mesh.switch_edge(mesh.switch_vertex(_f.tuple())));
+                } else if ((utils::SimplexComparisons::equal(
```
daniel-zint, line 145: As code coverage indicates, I am pretty sure that `cofaces_single_dimension` already gives you the tuples with the desired orientation.

PR #579, opened 2023-12-14
Dzint/578 attribute checkpoints

daniel-zint:


src/wmtk/attribute/AttributeManager.hpp
```diff
@@ -137,6 +142,21 @@ MeshAttributeHandle<T> AttributeManager::register_attribute(
     return r;
 }
 template <typename T>
+inline T AttributeManager::parent_scope(std::function<T()> f)
+{
+    change_to_parent_scope();
```
mtao, line 147: this change to parent scope needs to happen for all meshes simultaneously (in case the user wants to specify reading some previous state of another mesh in the system?). The change_to_parent_scope should probably invoke some functor like teh AttributeScopeHandle object did to switch all meshes to a parent scope or child scope

src/wmtk/attribute/AttributeManager.hpp
```diff
@@ -137,6 +142,21 @@ MeshAttributeHandle<T> AttributeManager::register_attribute(
...
+    T return_value = f();
```
mtao, line 148: I'm not sure we need to make sure everything passes into a std::function. in reality we just want something that satisfies the `std::is_invokable` concept (https://en.cppreference.com/w/cpp/concepts/invocable). Too bad we're not on c++20, but i think it might be cleaner to static assert for invokability rather than to force everything get mapped to a std::function

src/wmtk/attribute/AttributeScopeHandle.hpp
```diff
@@ -2,14 +2,40 @@
 
 namespace wmtk::attribute {
 struct AttributeManager;
+
+/**
+ * This handle is a wrapper for the MeshManager scope funtions.
+ * It's main task is to enable RAII for attribute scopes.
+ *
+ * It is NOT a handle for `AttributeScope`.
```
mtao, line 10: in abstract it is - this "handle" effectively corresponds to the lifetime of some AttributeScope - the implementation just doesn't track which scope. But yeah we could rename it / make the impl a bit more aware of which scope it is supposed to track?

src/wmtk/attribute/AttributeScopeStack.cpp
```diff
@@ -93,6 +101,19 @@ AttributeScope<T> const* AttributeScopeStack<T>::get_checkpoint(long index) cons
         return m_checkpoints.at(index);
     }
 }
+template <typename T>
+void AttributeScopeStack<T>::change_to_parent_scope()
+{
+    assert(!empty());
+    m_current = m_current->parent();
```
mtao, line 108: assert that m_current->parent() != nullptr?
can m_current ever be nullptr / what is the spec like?

mtao (changes requested):
generally in good shape - i just think there's another level of scope creation that needs to be propagated to all of the meshes in a MM. I also think using std::invoke / making the interface a little variadic could make the tool more powerful than just satisfying some standard prototype but maybe that can be added later on as a TODO

src/wmtk/attribute/AttributeScopeStack.cpp
daniel-zint: I assume in the accessor that `m_current->parent()` can be a nullptr. In that case I am accessing the base instead of the cache.

PR #582, opened 2023-12-14
add tet invertion invariant

JcDai:


src/wmtk/invariants/TetrahedronInversionInvariant.cpp
```diff
@@ -0,0 +1,37 @@
...
+TetrahedronInversionInvariant::TetrahedronInversionInvariant(
+    const Mesh& m,
+    const MeshAttributeHandle<double>& coordinate)
+    : Invariant(m)
+    , m_coordinate_handle(coordinate)
+{}
+
+bool TetrahedronInversionInvariant::after(PrimitiveType type, const std ::vector<Tuple>& tets) const
+{
+    if (type != PrimitiveType::Tetrahedron) return true;
```
danielepanozzo, line 16: Why not an exception? This should not happen.

danielepanozzo (approved):
looks good to me

src/wmtk/invariants/TetrahedronInversionInvariant.cpp
mtao: the generic underlying invariant calls basically invoke `after` on every type of simplex possible. We could/should decrease this by letting invariants register which dimensions they care about.

PR #584, opened 2023-12-15
add tetmesh link condition to multimesh link condition invariant

JcDai:
very tiny change. add link condition check for tetmesh to multimeshlinkcondition invariant and replace the code from deprecated simplicialcomplex


daniel-zint (approved):
lgtm

PR #590, opened 2023-12-17
Invariants::after takes the operation input simplex.

daniel-zint:


mtao:
should it take the input simplex or should it take the input equivalent set of simplices? we're mixing semantics atm i think

daniel-zint:
> should it take the input simplex or should it take the input equivalent set of simplices? we're mixing semantics atm i think

I was thinking about that too. The thing is, we cannot get the equivalent without knowing the operation. Let's discuss that tomorrow.

src/wmtk/operations/tri_mesh/EdgeSwapBase.cpp
```diff
@@ -103,9 +103,12 @@ bool EdgeSwapBase::execute()
...
 {
-    std::vector<Simplex> s;
-    s.emplace_back(simplex::Simplex::edge(m_output_tuple));
-    return s;
+    return {simplex::Simplex::edge(m_output_tuple)};
+}
+
+std::vector<Simplex> EdgeSwapBase::unmodified_primitives() const
+{
+    return {input_simplex()};
```
mtao, line 111: weren't you the one who had a compiler complain about this return notation? i've been avoiding this for you - but i guess i can do this again ;-) 
(i think this was particularly with std::array your compiler was not happy)

src/wmtk/optimization/LineSearch.cpp
```diff
@@ -28,7 +28,9 @@ bool LineSearch::check_state() const
                                                // vertex here but I didn't know for sure.
     bool after_pass = true;
     for (const PrimitiveType pt : wmtk::utils::primitive_below(top_type)) {
-        after_pass &= m_invariants.after(pt, modified_simplices(pt));
+        after_pass &= m_invariants.after(
```
mtao, line 31: don't worry about this guess, it's being deleted anyway ;-)

mtao (approved):
seems in good state to me

src/wmtk/operations/tri_mesh/EdgeSwapBase.cpp
daniel-zint: No my compiler never complained about that stuff. It was only that I didn't know that for `std::array` you should write `{{...}}` but it never was an issue.

PR #595, opened 2023-12-20
593 tet swap operation

JcDai:
add edge swap for tetmesh. 
modify return data of split and collapse


JcDai:
deprecate after implementing in another branch

PR #597, opened 2023-12-22
Consolidate

danielepanozzo:


src/wmtk/EdgeMesh.hpp
```diff
@@ -60,6 +60,8 @@ class EdgeMesh : public Mesh
 
     bool is_connectivity_valid() const override;
 
+    std::vector<std::vector<TypedAttributeHandle<long>>> connectivity_attributes();
```
mtao, line 63: could be const

src/wmtk/attribute/Attribute.cpp
```diff
@@ -170,6 +171,30 @@ void Attribute<T>::clear_current_scope()
     }
 }
 
+template <typename T>
+void Attribute<T>::consolidate(const std::vector<long>& new2old)
+{
+    for (long i = 0; i< new2old.size(); ++i)
```
mtao, line 177: just to point out, eigen can do this: https://eigen.tuxfamily.org/dox/group__TutorialSlicingIndexing.html#TutorialSlicingArray

not worth doing for this task, but i've had good performance boosts on similar tasks by trusting the lib to do striding / packing right (we're striding properly here anyway)

mtao (approved):
just one constness thing, lgtm!

PR #602, opened 2023-12-23
Tet mesh edge swap

JcDai:
real branch for tetmesh edge swap and other operations


src/wmtk/TetMeshOperationExecutor.cpp
```diff
@@ -213,7 +213,7 @@ TetMesh::TetMeshOperationExecutor::get_split_simplices_to_delete(
     return ids;
 }
 
-const std::array<std::vector<long>, 4>
+const std::pair<std::array<std::vector<long>, 4>, std::array<std::vector<Tuple>, 4>>
```
mtao, line 216: returning all this is an optimization that isn't clear / obfuscates the code a bit imo

src/wmtk/TetMeshOperationExecutor.cpp
```diff
@@ -674,7 +689,9 @@ void TetMesh::TetMeshOperationExecutor::split_edge()
 
 void TetMesh::TetMeshOperationExecutor::collapse_edge()
 {
-    simplex_ids_to_delete = get_collapse_simplices_to_delete(m_operating_tuple, m_mesh);
+    auto simples_to_delete = get_collapse_simplices_to_delete(m_operating_tuple, m_mesh);
+    simplex_ids_to_delete = std::get<0>(simples_to_delete);
```
mtao, line 693: `std::tie(simplex_ids_to_delete, simplex_tuples_to_delete) = simplices_to_delete)`

src/wmtk/operations/EdgeCollapse.cpp
```diff
@@ -49,9 +51,10 @@ std::vector<Simplex> EdgeCollapse::unmodified_primitives(
 ////////////////////////////////////
 std::vector<Simplex> EdgeCollapse::execute(TetMesh& mesh, const Simplex& simplex)
 {
-    Accessor<long> accessor = hash_accessor();
-    auto return_data = mesh.collapse_edge(simplex.tuple(), accessor);
-    return {Simplex::vertex(return_data.m_output_tuple)};
+    auto return_data = operations::utils::multi_mesh_edge_collapse(mesh, simplex.tuple());
+    const operations::tet_mesh::EdgeOperationData& my_data = return_data.get(mesh, simplex);
+    deleted_edges = my_data.deleted_simplex_tuples(mesh, PrimitiveType::Edge);
```
mtao, line 56: I thought edgecollapse shouldn't store state - did we agree to have this around for this round?

src/wmtk/operations/EdgeSplit.cpp
```diff
@@ -50,10 +51,16 @@ std::vector<Simplex> EdgeSplit::unmodified_primitives(const TriMesh& mesh, const
 ///////////////////////////////
 std::vector<Simplex> EdgeSplit::execute(TetMesh& mesh, const Simplex& simplex)
 {
-    Accessor<long> accessor = hash_accessor();
-    auto return_data = mesh.split_edge(simplex.tuple(), accessor);
+    auto return_data = utils::multi_mesh_edge_split(mesh, simplex.tuple());
+
+    spdlog::trace("{}", primitive_type_name(simplex.primitive_type()));
```
mtao, line 56: use the logger now (teseo just removed all other spdlog non-logger statements)

src/wmtk/operations/EdgeSplit.cpp
```diff
@@ -71,7 +78,26 @@ std::pair<Tuple, Tuple> EdgeSplit::new_spine_edges(const Mesh& mesh, const Tuple
...
     constexpr static PrimitiveType PE = PrimitiveType::Edge;
     constexpr static PrimitiveType PF = PrimitiveType::Face;
-    return {new_vertex, mesh.switch_tuples(new_vertex, {PE, PF, PE})};
+    constexpr static PrimitiveType PT = PrimitiveType::Tetrahedron;
+
+    std::pair<Tuple, Tuple> ret;
+
+    switch (mesh.top_simplex_type()) {
+    case PE: ret = {new_vertex, mesh.switch_tuples(new_vertex, {PE})};
+    case PF: ret = {new_vertex, mesh.switch_tuples(new_vertex, {PE, PF, PE})};
```
mtao, line 87: don't you need to break from each of these?

src/wmtk/operations/EdgeSplit.hpp
```diff
@@ -25,6 +27,11 @@ class EdgeSplit : public MeshOperation
     std::vector<Simplex> execute(TetMesh& mesh, const Simplex& simplex) override;
     std::vector<Simplex> unmodified_primitives(const TetMesh& mesh, const Simplex& simplex)
         const override;
+
+private:
+    std::vector<Tuple> new_edges;
```
mtao, line 32: did we want to add state?

src/wmtk/operations/composite/TetEdgeSwap.cpp
```diff
@@ -0,0 +1,150 @@
...
+    const auto split_simplicies = m_split(simplex);
+    if (split_simplicies.empty()) return {};
+    assert(split_simplicies.size() == 1);
+
+    // after split
+    const Tuple& split_ret = split_simplicies.front().tuple();
+    assert(!mesh().is_boundary(split_ret));
+    auto iter_tuple = split_ret;
+    std::vector<Tuple> candidate_edge_tuples;
+    while (true) {
```
mtao, line 28: couldn't calling top dim cofaces work here? i thought we had some order semantic for edges?

src/wmtk/operations/composite/TetEdgeSwap.cpp
```diff
@@ -0,0 +1,150 @@
...
+        candidate_edge_tuples.push_back(mesh().switch_edge(iter_tuple));
+        iter_tuple = mesh().switch_tetrahedron(mesh().switch_face(iter_tuple));
+        if (iter_tuple == split_ret) {
+            break;
+        }
+    }
+
+    assert(m_collapse_index < candidate_edge_tuples.size());
```
mtao, line 36: i thought the point was that you would run multiple passes - wouldn't this make the whole system stop running after hitting some edge with a valence below your desired valence?

src/wmtk/operations/composite/TetEdgeSwap.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    // check if #edge_generated_by_split > #edges_deleted_by_collapse, if not, return face instead
+    assert(edges_generated_by_split.size() >= simplices_deleted_by_collapse[1].size());
+
+    bool able_to_return_edges = true;
+    if (edges_generated_by_split.size() == simplices_deleted_by_collapse[1].size())
+        able_to_return_edges = false;
+
+    if (able_to_return_edges) {
+        // can use std::difference with sorted vector, worth?
```
mtao, line 65: yes

src/wmtk/operations/composite/TetEdgeSwap.cpp
```diff
@@ -0,0 +1,150 @@
...
+        for (const auto& e_split : edges_generated_by_split) {
+            bool preserve_flag = true;
+            for (const auto& e_collapse : simplices_deleted_by_collapse[1]) {
```
mtao, line 68: i think for both the edge and face cases daniel's simplex collection code has the ability to find / it sorts things for you

src/wmtk/operations/composite/TetEdgeSwap.cpp
```diff
@@ -0,0 +1,150 @@
...
+    const auto collapse_simplicies =
+        m_collapse(Simplex(m_collapse.primitive_type(), collapse_tuple));
+    if (collapse_simplicies.empty()) return {};
+    assert(collapse_simplicies.size() == 1);
+
+
+    // resurrect the edge tuples
+    if (able_to_return_edges) {
+        for (long i = 0; i < edge_tuples_generated_by_swap.size(); ++i) {
+            edge_tuples_generated_by_swap[i] = resurrect_tuple(edge_tuples_generated_by_swap[i]);
```
mtao, line 122: did every edge tuple from the swap surrive? you need to delete one from this set right?

mtao: why did you do two loops here?
there's also afunction for this in wmtk::simplex::utils::homgogeneous*

src/wmtk/operations/tet_mesh/EdgeOperationData.cpp
```diff
@@ -49,17 +51,117 @@ std::array<Tuple, 2> EdgeOperationData::input_endpoints(const TetMesh& m) const
 
 std::vector<Tuple> EdgeOperationData::collapse_merged_ear_edges(const TetMesh& m) const
 {
```
mtao, line 53: what happened with all of these impl? are they replaced somewhere?

mtao: the ordering in these functions was important for attribute transfer BTW.

src/wmtk/operations/tet_mesh/EdgeOperationData.cpp
```diff
@@ -49,17 +51,117 @@ std::array<Tuple, 2> EdgeOperationData::input_endpoints(const TetMesh& m) const
...
+    const
+{
+    std::vector<Tuple> ret;
+    switch (pt) {
+    case PrimitiveType::Vertex: {
+        return simplex_tuples_to_delete[0];
+        break;
+    }
+    case PrimitiveType::Edge: {
+        return simplex_tuples_to_delete[1];
```
mtao, line 115: why do a switch case rather than read a single index?

src/wmtk/operations/tet_mesh/EdgeOperationData.hpp
```diff
@@ -148,5 +155,12 @@ class EdgeOperationData : public wmtk::operations::EdgeOperationData
     std::vector<TetCollapseData> tet_collapse_data;
     std::vector<FaceCollapseData> face_collapse_data;
     std::vector<TetSplitData> tet_split_data;
+
+    // data required for swap
+    // TODO: can be used as cache for speedup
+    std::vector<long> m_new_tet_ids;
```
mtao, line 161: there is a lot more structure to this than represented here. i would rather store these values according to their structure and then reconstruct htis when necessary later rather than just lose the structure.
also if you're going to create 4 vectors like this just do an array of vectosr anyway

mtao (changes requested):

src/wmtk/operations/EdgeCollapse.cpp
JcDai: yes this is not used. I'll remove it

src/wmtk/operations/tet_mesh/EdgeOperationData.cpp
JcDai: these didn't compile before I commented them out

src/wmtk/operations/composite/TetEdgeSwap.cpp
JcDai: top_dimension_cofaces does bfs, I guess the order is not what we want

src/wmtk/operations/composite/TetEdgeSwap.cpp
JcDai: Every edge in this set survives.

Got dumb for the two loops  :/

src/wmtk/operations/composite/TetEdgeSwap.cpp
```diff
@@ -0,0 +1,152 @@
...
+    std::vector<Tuple> candidate_edge_tuples;
+    while (true) {
+        candidate_edge_tuples.push_back(mesh().switch_edge(iter_tuple));
+        iter_tuple = mesh().switch_tetrahedron(mesh().switch_face(iter_tuple));
+        if (iter_tuple == split_ret) {
+            break;
+        }
+    }
+
+    assert(m_collapse_index < candidate_edge_tuples.size());
```
teseoch, line 36: we should fail the op not assert

teseoch (commented):
we need some unit test

mtao (approved):
lgtm

PR #603, opened 2023-12-23
Attribute update strategy api updates

mtao:


tests/io/test_io.cpp
```diff
@@ -121,15 +121,21 @@ TEST_CASE("msh_3d", "[io]")
...
-        wmtk::operations::NewAttributeStrategy::CollapseBasicStrategy::CopyTuple);
+    auto attribute_handle = m.register_attribute<long>(std::string("test_attribute"), PE, 1);
+
+    {
+        // get the casted attribute types
+        auto& split_strat = attribute_handle.trimesh_standard_split_strategy();
+        auto& collapse_strat = attribute_handle.trimesh_standard_collapse_strategy();
+
+        // set the strategies
+        split_strat.set_standard_split_strategy(
```
teseoch, line 132: can you add an example with the lambda too?

teseoch (approved):

PR #604, opened 2023-12-23
Wildmeshing

teseoch:


src/wmtk/Scheduler.cpp
```diff
@@ -13,30 +16,40 @@ Scheduler::~Scheduler() = default;
...
 
-    const auto tups = op.mesh().get_all(type);
 
-    std::vector<Simplex> simplices =
-        wmtk::simplex::utils::tuple_vector_to_homogeneous_simplex_vector(tups, type);
+    logger().info("Executing on {} simplices", simplices.size());
 
-    spdlog::info("Executing on {} simplices", simplices.size());
+    {
+        POLYSOLVE_SCOPED_STOPWATCH("Sorting", res.sorting_time, logger());
```
mtao, line 33: why did this need a macro? if this is just some RAII thing i always just create a class ;-)

src/wmtk/invariants/FunctionInvariant.hpp
```diff
@@ -0,0 +1,30 @@
...
+    FunctionInvariant(
+        const PrimitiveType type,
+        const std::shared_ptr<function::PerSimplexFunction>& func);
+
+    bool after(
+        const std::vector<Tuple>& top_dimension_tuples_before,
+        const std::vector<Tuple>& top_dimension_tuples_after) const override;
+
+private:
+    std::shared_ptr<function::PerSimplexFunction> m_func;
```
mtao, line 26: do you intend on having this function shrared between multiple places? if you're not aware, std::function has a check if it has a value or not https://en.cppreference.com/w/cpp/utility/functional/function/operator_bool

mtao (approved):
lgtm

src/wmtk/invariants/FunctionInvariant.hpp
teseoch: yes it will be shared

src/wmtk/Scheduler.cpp
teseoch: it comes from polysolve, i will not refactor it.

PR #609, opened 2023-12-27
Attribute update

teseoch:


src/wmtk/attribute/AttributeInitializationHandle.hpp
```diff
@@ -28,20 +28,8 @@ namespace wmtk::attribute {
 class AttributeInitializationHandleBase
```
mtao, line 28: I think this handle is no long necsesary - i will just delete it soon

src/wmtk/operations/tri_mesh/PredicateAwareCollapseNewAttributeStrategy.cpp
```diff
@@ -7,7 +7,14 @@ namespace wmtk::operations::tri_mesh {
...
     wmtk::attribute::MeshAttributeHandle<T>& h)
-    : CollapseNewAttributeStrategy(dynamic_cast<TriMesh&>(h.mesh()))
+    : PredicateAwareCollapseNewAttributeStrategy(h, h.mesh())
+{}
+
+template <typename T>
+PredicateAwareCollapseNewAttributeStrategy<T>::PredicateAwareCollapseNewAttributeStrategy(
+    const wmtk::attribute::MeshAttributeHandle<T>& h,
+    Mesh& m)
+    : CollapseNewAttributeStrategy(dynamic_cast<TriMesh&>(m))
```
mtao, line 17: i don't think we actually need the tri-mesh ness here right?

mtao (approved):
lgtm

PR #613, opened 2023-12-28
change edgemesh split return tup to Vnew Enew[1]

JcDai:
change edgemesh split return tuple to adopt the convention in Tri and TetMesh, that is, the new vertex and the edge v_new-v1 (suppose the input is v0-v1)


src/wmtk/EdgeMeshOperationExecutor.cpp
```diff
@@ -151,7 +151,12 @@ Tuple EdgeMesh::EdgeMeshOperationExecutor::split_edge_single_mesh()
     delete_simplices();
 
     // prepare return Tuple
-    return m_mesh.edge_tuple_from_id(new_eids[0]);
+    auto ret_edge = m_mesh.edge_tuple_from_id(new_eids[1]);
+    if (m_mesh.id_vertex(ret_edge) != v_new) {
+        ret_edge = m_mesh.switch_vertex(ret_edge);
```
mtao, line 156: add some asserts for all the other values the simplex should have?

mtao (changes requested):
add more checks for the other components of hte return value (check all local indices are the same as the input one + poinst to the right hting?)

PR #615, opened 2023-12-29
removed default transfer

teseoch:


src/wmtk/operations/tri_mesh/BasicSplitNewAttributeStrategy.cpp
```diff
@@ -60,7 +58,7 @@ void BasicSplitNewAttributeStrategy<T>::assign_split(
     const std::array<Tuple, 2>& split_simplices)
 {
     if (!bool(m_split_op)) {
-        return;
+        throw std::runtime_error("Spine split attribute needs to have a transfer");
```
mtao, line 61: we should leave a comment specifying this behavior in a header somewhere

mtao (approved):
lgtm, but i think we need to start commenting these conventions a bit more

PR #621, opened 2023-12-30
removed unnecessary classes and fix inversion invariant

teseoch:


mtao (approved):
stylistic complaint about having a large if/else if/else if statement taht is hard to see at once

PR #627, opened 2023-12-31
default strat throw

teseoch:


mtao (approved):
lgtm after warning as error gets fixed?

PR #629, opened 2024-01-02
use cache in app

daniel-zint:
Update components and wmtk app to use the `io::Cache` instead of a map of files. Furthermore, update components options and other outdated things.


mtao (approved):
lgtm?

teseoch (approved):

PR #635, opened 2024-01-03
added component to load an uv mesh

teseoch:


components/wmtk_components/multimesh/multimesh.cpp
```diff
@@ -0,0 +1,33 @@
...
+    std::shared_ptr<Mesh> child = cache.read_mesh(options.child);
+
+    if (parent->top_simplex_type() == child->top_simplex_type() &&
+        parent->capacity(parent->top_simplex_type()) ==
+            child->capacity(child->top_simplex_type())) {
+        std::vector<int64_t> parent_simplices(parent->capacity(parent->top_simplex_type()));
+        std::iota(parent_simplices.begin(), parent_simplices.end(), 0);
+
+        auto child_map =
+            multimesh::same_simplex_dimension_surjection(*parent, *child, parent_simplices);
```
mtao, line 21: i guess for now this is just pipelining to enable obj-style mappings? we could clarify that i guess. need a better name for all these things

mtao (approved):
lgtm?

components/wmtk_components/multimesh/multimesh.cpp
teseoch: yes, i have a throw for the other cases, we can add as needed

PR #640, opened 2024-01-05
add max function invarianr

JcDai:


mtao (approved):
seems reasonable to me

PR #643, opened 2024-01-06
Multimesh tetmesh update

JcDai:


src/wmtk/TetMeshOperationExecutor.cpp
```diff
@@ -127,9 +129,32 @@ TetMesh::TetMeshOperationExecutor::TetMeshOperationExecutor(
...
     for (const simplex::Simplex& t :
          hash_update_region.simplex_vector(PrimitiveType::Tetrahedron)) {
         cell_ids_to_update_hash.push_back(m_mesh.id(t));
+
+        auto faces = wmtk::simplex::faces(m, t, false);
+        faces.add(t);
+        faces.sort_and_clean();
+
+        auto load = [&](PrimitiveType pt, size_t index) {
+            auto simps = faces.simplex_vector(pt);
```
mtao, line 143: there's an idiosycnracy where simplex_vector(pt) is much slower than simplex_vector().
probably better to just go through simplex_vector() and insert into the right global_simplex..hashes for each one

mtao (approved):
lgtm

PR #648, opened 2024-01-08
Envelope

teseoch:


daniel-zint (commented):
Looks good so far but you have compilation errors.

src/wmtk/invariants/EnvelopeInvariant.hpp
```diff
@@ -0,0 +1,31 @@
...
+}
+
+namespace wmtk::invariants {
+class EnvelopeInvariant : public Invariant
+{
+public:
+    EnvelopeInvariant(
+        const Mesh& m,
+        const TypedAttributeHandle<double>& coordinate,
+        const Mesh& envelope_mesh,
```
daniel-zint, line 19: The mesh we want to put an envelope around could be non-manifold. As much as I dislike it, I think we should use IGL-style matrices here.

PR #651, opened 2024-01-08
Tetmesh multimesh tests

JcDai:


src/wmtk/multimesh/utils/extract_child_mesh_from_tag.cpp
```diff
@@ -114,7 +114,53 @@ std::shared_ptr<Mesh> internal::TupleTag::extract_and_register_child_mesh_from_t
...
+
+        // for (auto key : parent_to_child_vertex_map) {
+        //     std::cout << key.first << " --> " << key.second << std::endl;
+        // }
+
+        // std::cout << tet_mesh_matrix << std::endl;
+
+        std::shared_ptr<TetMesh> child_ptr = std::make_shared<TetMesh>();
+        auto& child = *child_ptr;
+        child.initialize(tet_mesh_matrix);
```
mtao, line 151: ok, this is following what was done previously and lets leave it as this for now, but in the future we really shouldn't need indices here lol. the EV / FE / TF matrices should suffice for generating meshes lol

src/wmtk/operations/utils/UpdateEdgeOperationMultiMeshMapFunctor.cpp
```diff
@@ -156,7 +155,13 @@ void UpdateEdgeOperationMultiMeshMapFunctor::update_ear_replacement(
                     continue;
                 }
 
-                const int64_t parent_old_vid = m.id_vertex(parent_tuple);
+
+                parent_tuple = m.parent_scope(
```
mtao, line 159: add A TODO / comment as to why this is necessary. this existing is a bug i think right? the parent_tuple should already be valid in the parent scope

mtao (approved):
there's perhaps cleanup to be done later on - just leave comments that things need t obe checked later on?

src/wmtk/operations/utils/UpdateEdgeOperationMultiMeshMapFunctor.cpp
JcDai: I also think parent_tuple should be already valid in the parent scope. But there is a case I met in tet, where if I don't resurrect it in the parent  scope,  it fails is_valid check in the parent scope. need to look into it later

PR #657, opened 2024-01-10
656 refactor optimization smooth

zlyfunction:
Refactor optimizationSmooth.*pp so that it can smooth the energy defined(registered) on a child mesh.

### DIFF
--- a/src/wmtk/operations/OptimizationSmoothing.cpp
+++ b/src/wmtk/operations/OptimizationSmoothing.cpp
@@ -118,8 +118,10 @@ bool OptimizationSmoothing::WMTKProblem::is_step_valid(const TVector& x0, const 
 }
 
 
-OptimizationSmoothing::OptimizationSmoothing(std::shared_ptr<wmtk::function::Function> energy)
-    : AttributesUpdate(energy->mesh())
+OptimizationSmoothing::OptimizationSmoothing(
+    Mesh& m,
+    std::shared_ptr<wmtk::function::Function> energy)
+    : AttributesUpdate(m)
     , m_energy(energy)
 {
     m_linear_solver_params = R"({"solver": "Eigen::LDLT"})"_json;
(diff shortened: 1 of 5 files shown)


src/wmtk/operations/OptimizationSmoothing.cpp
```diff
@@ -104,11 +104,23 @@ bool OptimizationSmoothing::WMTKProblem::is_step_valid(const TVector& x0, const
...
+            domain.end(),
+            std::back_inserter(dom_tmp),
+            [](const simplex::Simplex& s) { return s.tuple(); });
+    } else {
+        std::transform(
+            domain.begin(),
+            domain.end(),
+            std::back_inserter(dom_tmp),
+            [&](const simplex::Simplex& s) {
+                return m_energy.mesh().map_tuples(m_invariants.mesh(), s).front();
```
mtao, line 120: Why can you just pick the first one? Shouldn't the default behavior be too pick everything? For instance if you check the energy of a position vertex that is on a seam then you should check the energy on both sides of the seam?

src/wmtk/operations/OptimizationSmoothing.cpp
```diff
@@ -140,17 +154,27 @@ void OptimizationSmoothing::create_solver()
...
-        simplex,
-        m_invariants,
-        *m_energy);
-
-    auto x = problem.initial_value();
-    try {
-        m_solver->minimize(problem, x);
-    } catch (const std::exception&) {
-        return {};
+    std::vector<simplex::Simplex> simplices_to_smooth;
```
mtao, line 157: It's not clear to be whether the multi mesh behavior or the optimization should be to trigger multiple optimizations. This should be a user driven choice

src/wmtk/operations/OptimizationSmoothing.hpp
```diff
@@ -21,7 +21,7 @@ class OptimizationSmoothing : public AttributesUpdate
     class WMTKProblem;
 
 public:
-    OptimizationSmoothing(std::shared_ptr<wmtk::function::Function> energy);
+    OptimizationSmoothing(Mesh& m, std::shared_ptr<wmtk::function::Function> energy);
```
mtao, line 24: The original constructor skills still exist where it forwards the energy mesh

mtao (changes requested):
When there's time we should discuss how to make your choices more generic

PR #664, opened 2024-01-12
add cmake for volumeremesher

JcDai:


src/wmtk/utils/Rational.hpp
```diff
@@ -180,6 +180,12 @@ class Rational
         os << mpq_get_d(r.value);
         return os;
     }
+
+    template <typename T>
+    void init(const T& v)
```
mtao, line 185: is this used anywhere?

mtao (approved):
lets go over this more later i guess. it dosn't hurt anything fro now lol

PR #666, opened 2024-01-14
Basic Multimesh IsBoundary functionality

mtao:
Implemented a struct to allow for multimesh is_boundary checks - question remains where we should enter this
* do we dump this into the mesh class?
* do we have functions opt in through a parameter?

i don't care for forcing the definition of a multimesh boundary into the mesh class as different operations might have different demands on this front. On the other hand it's quite annoying to have to propagate the definition of a boundary as an argument to all of these operators


src/wmtk/MultiMeshManager.hpp
```diff
@@ -1,13 +1,12 @@
...
 
 #include <tuple>
-#include "Accessor.hpp"
 #include "Tuple.hpp"
 #include "attribute/AttributeScopeHandle.hpp"
 #include "attribute/MeshAttributes.hpp"
 // included to make a friend as this requires IDs
-#include <wmtk/multimesh/same_simplex_dimension_surjection.hpp>
 #include <wmtk/operations/utils/UpdateVertexMultiMeshMapHash.hpp>
+#include <wmtk/multimesh/same_simplex_dimension_surjection.hpp>
```
teseoch, line 9: we don't need it, right?

src/wmtk/invariants/MultiMeshLinkConditionInvariant.cpp
```diff
@@ -15,21 +15,23 @@ namespace {
...
-    {
-        return simplex::link_condition(m, s.tuple());
-    }
-    bool operator()(const TetMesh& m, const simplex::Simplex& s) const
-    {
-        return simplex::link_condition(m, s.tuple());
+    bool operator()(const Mesh& m, const simplex::Simplex& s) const { 
+        return simplex::link_condition(m,s.tuple());
     }
+    //bool operator()(const PointMesh& m, const simplex::Simplex& s) const { return true; }
```
teseoch, line 21: this can be deleted?

src/wmtk/attribute/MeshAttributeHandle.hpp
```diff
@@ -29,6 +29,8 @@ class MeshAttributeHandle
         TypedAttributeHandle<double>,
         TypedAttributeHandle<Rational>>;
 
+    using ValueVariant = std::variant<char, int64_t, double, Rational>;
```
teseoch, line 32: i don't see where tis is used

mtao: i extended extract_mesh_from_tag to support arbitrary tag types - so you need to be able to pass in a variadic input type to avoid a template. Consider this fake function:
```get_isocontour(const MeshAttributeHandle& handle, const MeshAttributeHandle::ValueVariant& value)```
. technically this should support eigen vector types, i'll fix that later

src/wmtk/MultiMeshManager.hpp
mtao: was having trouble building this without it for some reason

PR #672, opened 2024-01-15
added link_condition test

Zhouyuan-Chen:
I added the shark fin example and 3-order vertex example.

for shark fin, I think everything goes smoothly.

However, for the 3-order case, I am not quite sure if the behavior is correct, so I added a comment. this code is in test_2d_invariants.cpp 762 line.

### DIFF
--- a/tests/test_2d_invariants.cpp
+++ b/tests/test_2d_invariants.cpp
@@ -8,11 +8,11 @@
 #include "tools/TriMesh_examples.hpp"
 
 #include <wmtk/EdgeMesh.hpp>
+#include <wmtk/attribute/TypedAttributeHandle.hpp>
 #include <wmtk/invariants/MinIncidentValenceInvariant.hpp>
 #include <wmtk/invariants/MultiMeshTopologyInvariant.hpp>
 #include <wmtk/invariants/TetMeshSubstructureTopologyPreservingInvariant.hpp>
 #include <wmtk/invariants/TriMeshSubstructureTopologyPreservingInvariant.hpp>
-#include <wmtk/attribute/TypedAttributeHandle.hpp>
 #include <wmtk/multimesh/utils/extract_child_mesh_from_tag.hpp>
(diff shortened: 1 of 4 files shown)


JcDai:
Every grid cell should have 6 tets, but in the TV matrix in tetmesh examples has only 5 tets for a cell instead?

Zhouyuan-Chen:
@JcDai I used two patterns to construct the whole TetMesh. I guess it should be okay? I am not sure if it would cause some problems. If it will, let me know.

JcDai:
> @JcDai I used two patterns to construct the whole TetMesh. I guess it should be okay? I am not sure if it would cause some problems. If it will, let me know.

Oh yah nvm. This is not the setting I'm thinking of

Zhouyuan-Chen:
> > @JcDai I used two patterns to construct the whole TetMesh. I guess it should be okay? I am not sure if it would cause some problems. If it will, let me know.
> 
> Oh yah nvm. This is not the setting I'm thinking of

okay, then I need to clean them up, just double check, your expectation is the following?

https://www.researchgate.net/figure/Decomposition-of-a-grid-cell-into-6-tetrahedra-All-of-them-share-an-edge-which-is-one-of_fig5_362015540

JcDai:
> https://www.researchgate.net/figure/Decomposition-of-a-grid-cell-into-6-tetrahedra-All-of-them-share-an-edge-which-is-one-of_fig5_362015540

no you don't need to change your setting, it's fine

Zhouyuan-Chen:
There is likely a bug existing in the link3D. (boundary face case).
when I want to get link faces, some faces are the coface of the original vertex, which is wrong.
I checked my tetmesh example, and think it should be right. you can have a double check for that. `two_by_two_by_two_grids_tets`

I added a comment to the place where strange happened. `test_simplex_collection.cpp 1803 line`

I am hundred percent sure.

PR #709, opened 2024-01-29
Dzint/avoid memory allocations

daniel-zint:
Remove small memory allocations by passing `SimplexCollection`s as references.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,6 +158,8 @@ if(MSVC)
     target_compile_options(wildmeshing_toolkit PUBLIC $<$<CONFIG:RELWITHDEBINFO>:/GL>)
     target_compile_options(wildmeshing_toolkit PUBLIC $<$<CONFIG:RELWITHDEBINFO>:/Ob2>) # inline whenever suitable
     target_compile_options(wildmeshing_toolkit PUBLIC $<$<CONFIG:RELWITHDEBINFO>:/Ot>) # favor faster code over binary size
+    target_compile_options(wildmeshing_toolkit PUBLIC $<$<CONFIG:RELEASE>:/GL>)
+    target_compile_options(wildmeshing_toolkit PUBLIC $<$<CONFIG:RELEASE>:/Ot>)
     target_compile_definitions(wildmeshing_toolkit PUBLIC _SILENCE_ALL_MS_EXT_DEPRECATION_WARNINGS)
 endif()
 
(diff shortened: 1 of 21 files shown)


src/wmtk/TetMeshOperationExecutor.cpp
```diff
@@ -124,18 +138,16 @@ TetMesh::TetMeshOperationExecutor::TetMeshOperationExecutor(
 
     // update hash on all tets in the two-ring neighborhood
     simplex::SimplexCollection hash_update_region(m);
-    for (const simplex::Simplex& v : edge_closed_star.simplex_vector(PrimitiveType::Vertex)) {
-        const simplex::SimplexCollection v_closed_star =
-            simplex::top_dimension_cofaces(m_mesh, v, false);
-        hash_update_region.add(v_closed_star);
+    for (const simplex::Simplex& v : edge_closed_star_vertices.simplex_vector()) {
+        simplex::top_dimension_cofaces(v, hash_update_region, false);
```
mtao, line 142: I'd rather pass a mesh and a vector for top dimension cofaces than use the SimplexCollection interface - it's just making code more opaque here (the mesh is implicit now, is there ever a use of this interface where you don't pass in false?)

src/wmtk/TetMeshOperationExecutor.cpp
```diff
@@ -124,18 +138,16 @@ TetMesh::TetMeshOperationExecutor::TetMeshOperationExecutor(
...
     global_simplex_ids_with_potentially_modified_hashes.resize(4);
     simplex::SimplexCollection faces(m_mesh);
+    faces.reserve(hash_update_region.simplex_vector().size() * 15);
 
-    for (const simplex::Simplex& t :
-         hash_update_region.simplex_vector(PrimitiveType::Tetrahedron)) {
+    for (const simplex::Simplex& t : hash_update_region.simplex_vector()) {
         cell_ids_to_update_hash.push_back(m_mesh.id(t));
 
         faces.add(wmtk::simplex::faces(m, t, false));
```
mtao, line 153: if anything this should be one of the places we reduce allocations

src/wmtk/simplex/link_single_dimension.cpp
```diff
@@ -0,0 +1,207 @@
...
+                all_cofaces.emplace_back(Simplex::vertex(t));
+            }
+        } else if (link_type == PrimitiveType::Edge) {
+            all_cofaces.reserve(cell_tuples.size() * 3);
+            for (Tuple t : cell_tuples) {
+                t = mesh.switch_tuples(t, {PV, PE, PF});
+                all_cofaces.emplace_back(Simplex::edge(t));
+                t = mesh.switch_tuples(t, {PV, PE});
+                all_cofaces.emplace_back(Simplex::edge(t));
+                t = mesh.switch_tuples(t, {PV, PE});
```
mtao, line 128: we need to make sure switch_tuple is all inlined and whatnot

mtao (approved):
looks like some good improvements.
However I want to doublecheck that the SimplexCollection object still makes sense - the performance deficiency we noted is that doing sorting / uniqueness on Tuple-based objects is inefficient so we should store IDs - but these ID invocations are the majority of the benefit that SimplexCollection provides over a std::vector. Furthermore, I'm still disatisfied with this implicit sorted state business - there's some nontrivial logic about managing sorted and unsorted simplex collections that gets shoved around + at this point the vast majority of our internal use of SimplexCollection returners involves saying not to sort?

CMakeLists.txt
```diff
@@ -156,6 +156,9 @@ target_link_libraries(wildmeshing_toolkit PUBLIC
 if(MSVC)
     target_compile_options(wildmeshing_toolkit PUBLIC /MP)
     target_compile_options(wildmeshing_toolkit PUBLIC $<$<CONFIG:RELWITHDEBINFO>:/GL>)
+    target_compile_options(wildmeshing_toolkit PUBLIC $<$<CONFIG:RELWITHDEBINFO>:/Ob2>) # inline whenever suitable
```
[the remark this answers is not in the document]
daniel-zint: The flag /Ob2 is set in Release by default but I added the other ones to Cmake too now.

src/wmtk/TetMeshOperationExecutor.cpp
daniel-zint: The vector is cleaned right after that loop. I think for now it's simpler to use the SimplexCollection but this code needs to be cleaned up in general.

PR #711, opened 2024-01-31
fix the iterate index error in simplex_link_3d test

Zhouyuan-Chen:
the iteration index is wrong in the first test of `simplex_link_3d`


daniel-zint (approved):
lgtm

PR #712, opened 2024-02-03
remove vtable calls

daniel-zint:


src/wmtk/PointMesh.hpp
```diff
@@ -10,6 +10,9 @@ namespace wmtk {
 // topologies
 class PointMesh : public Mesh
 {
+public:
+    friend class Mesh;
```
mtao, line 14: sure, but i don't know why this would be necessary (anything that is hidden in mesh shoudl be made protected?)

mtao (approved):
lgtm

PR #717, opened 2024-02-11
reduce accessor types

mtao:
removed const and mutable accessor types


src/wmtk/attribute/Accessor.hpp
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include "TupleAccessor.hpp"
+
+#include <Eigen/Dense>
+
+namespace wmtk::attribute {
+
+template <typename T>
+class Accessor: public TupleAccessor<T> {
```
daniel-zint, line 10: `Accessor` seems to be redundant. Maybe just replace it by `TupleAccessor`?

src/wmtk/attribute/MutableAccessor.hpp
```diff
@@ -1,67 +1,8 @@
...
-    using ConstAccessorType::ConstAccessorType;
-
-    using ConstAccessorType::const_scalar_attribute;
-    using ConstAccessorType::const_vector_attribute;
-    using ConstAccessorType::dimension;
-    using ConstAccessorType::reserved_size;
-    using ConstAccessorType::scalar_attribute;
-    using ConstAccessorType::vector_attribute;
-
+using MutableAccessor = TupleAccessor<T>;
```
daniel-zint, line 6: I think `MutableAccessor` can be removed completely.

src/wmtk/attribute/TupleAccessor.hpp
```diff
@@ -19,6 +25,9 @@ class TupleAccessor : protected CachingAccessor<T>
     friend class wmtk::Mesh;
     friend class wmtk::TetMesh;
     friend class wmtk::TriMesh;
+    friend class wmtk::EdgeMesh;
+    friend class wmtk::PointMesh;
+    friend class wmtk::TriMeshOperationExecutor;
```
daniel-zint, line 30: It's a bit odd that `TriMeshOperationExecutor` needs to be a friend but `TetMeshOperationExecutor` does not.

src/wmtk/attribute/TupleAccessor.hpp
```diff
@@ -30,10 +39,10 @@ class TupleAccessor : protected CachingAccessor<T>
...
     using CachingBaseType::CachingBaseType;
 
-    TupleAccessor(TupleAccessor&&) = default;
-    TupleAccessor& operator=(TupleAccessor&&) = default;
-    TupleAccessor(const TupleAccessor&) = delete;
-    TupleAccessor& operator=(const TupleAccessor&) = delete;
+    //TupleAccessor(TupleAccessor&&) = default;
+    //TupleAccessor& operator=(TupleAccessor&&) = default;
+    //TupleAccessor(const TupleAccessor&) = delete;
+    //TupleAccessor& operator=(const TupleAccessor&) = delete;
```
daniel-zint, line 45: delete?

daniel-zint (commented):
Looks good but I think there is some clean-up required still.

src/wmtk/attribute/MutableAccessor.hpp
mtao: i thought i'd deleted this file :upside_down_face: . wilco

src/wmtk/attribute/Accessor.hpp
mtao: i'm a bit torn on this - i like having the explicit semantic that a TupleAccessor is an Accessor. I guess i'll just replace TupleAccessor with Accessor (i don't want users to think about the Tuple aspect too much)

src/wmtk/attribute/TupleAccessor.hpp
mtao: yeah, i hadn't thought about that, but for whatever reason this was required

src/wmtk/attribute/TupleAccessor.hpp
mtao: yep. forwarding of constructors wasn't working properly and i was testing.

PR #719, opened 2024-02-12
Dzint/simplex holds global

daniel-zint:
Adding the global ids to `Simplex` allows for fast access to attributes as it is not necessary to query the id every time.


mtao:
superceded by #723

PR #724, opened 2024-02-14
Zhouyuan/optimize the marching component

daniel-zint:


Zhouyuan-Chen:
had added the position update, but still have a problem with it.

Also, as I have a closer look at our current code, I don't think it is practical to do so since we may need some modification for more files to fulfill this component's demand.

Zhouyuan-Chen:
I realized there is a bug in the marching component. I will fix it tomorrow.

components/tests/test_component_marching.cpp
```diff
@@ -144,6 +152,21 @@ TEST_CASE("marching_component_tri", "[components][marching]")
         CHECK(isosurface_vertex_num == expected_isosurface_vertex_num);
     }
 
+    const auto& edges = m.get_all(PrimitiveType::Edge);
+    wmtk::attribute::MeshAttributeHandle edge_tag_handle =
+        m.get_attribute_handle<int64_t>("marching_edge_tag", PrimitiveType::Edge);
```
daniel-zint, line 157: The names for the edge and face tag attributes should be given by the user.

components/tests/test_component_marching.cpp
```diff
@@ -167,25 +190,32 @@ TEST_CASE("marching_component_tri", "[components][marching]")
         }
     }
 
-    if (false) {
-        wmtk::io::ParaviewWriter
-            writer("marching_2d_result", "vertices", m, true, false, true, false);
+    if (true) {
+        wmtk::io::ParaviewWriter writer(
+            "/home/zhouyuan/workplace/toolkit/wildmeshing-toolkit/data/marching_2d_result",
```
daniel-zint, line 195: No absolute paths please.
Also this should not print anything by default --> `if(false)`

components/wmtk_components/marching/wmtk/components/marching/internal/Marching.cpp
```diff
@@ -83,18 +114,66 @@ void Marching::process()
...
+
+        op_split.add_transfer_strategy(etag_strategy);
+
+        /**************************face tag******************************/
+        auto compute_face_label =
+            [this](const Eigen::MatrixX<int64_t>& labels) -> Eigen::VectorX<int64_t> {
+            assert(labels.cols() == 3);
+            if (labels(0, 0) == m_output_value && labels(0, 1) == m_output_value &&
+                labels(0, 2) == m_output_value)
+                return Eigen::VectorX<int64_t>::Constant(1, m_output_value);
```
daniel-zint, line 158: add brackets around the return statement

daniel-zint (commented):
Some minor comments. Please add the attributes to the json options. Otherwise I think it's fine.

PR #730, opened 2024-02-20
Increase permissivity of Tuple::is_null

mtao:
the check now only confirms that the global id of the tuple is -1


daniel-zint (approved):
lgtm

PR #732, opened 2024-02-23
Dzint/component multi mesh from tag

daniel-zint:
This component adds a child mesh that is a manifold representation from tags that were potentially non-manifold.


components/wmtk_components/multimesh_from_tag/wmtk/components/multimesh_from_tag/multimesh_from_tag.cpp
```diff
@@ -0,0 +1,56 @@
...
+        keeps.emplace_back(substructure_label);
+        mesh.clear_attributes(keeps);
+    }
+
+    {
+        std::map<std::string, std::vector<int64_t>> names;
+        names[options.input] = mesh.absolute_multi_mesh_id();
+        names[options.output] = mmft.substructure()->absolute_multi_mesh_id();
+        names[options.output + "_soup"] =
+            mmft.substructure_soup()->absolute_multi_mesh_id(); // TODOfix: should be deleted
```
mtao, line 51: remember to delete

src/wmtk/Tuple.hpp
```diff
@@ -78,6 +78,10 @@ class Tuple
     /// Checks if a tuple is "null". This merely implies the global index is -1
     bool is_null() const;
     Tuple with_updated_hash(int64_t new_hash) const;
+
+    int8_t local_vid() const;
```
mtao, line 82: IIRC we purposely were leaving these as implementation details - please don't make public

mtao: (just to increase the friction of most people using these to do things, like access topological attributes directly)

src/wmtk/attribute/AttributeManager.hpp
```diff
@@ -86,6 +86,9 @@ class AttributeManager : public wmtk::utils::MerkleTreeInteriorNode
 
     std::string get_name(const attribute::TypedAttributeHandleVariant& attr) const;
 
+    template <typename T>
+    void set_name(const TypedAttributeHandle<T>& attr, const std::string& name);
```
mtao, line 90: just to be clear-  this should only be public in the manager class but the mesh class doesn't expose this right?

src/wmtk/simplex/tuples_preserving_primitive_types.hpp
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <vector>
+#include <wmtk/Tuple.hpp>
+
+
+namespace wmtk::simplex {
+
+/**
+ * @brief Compute all tuples that contain simplex(ptype1, t) and simplex(ptype2, t).
```
mtao, line 10: and are contained by simplex(ptype2) ?

src/wmtk/simplex/tuples_preserving_primitive_types.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+    Tuple t_iter = t;
+
+    std::vector<Tuple> intersection_tuples;
+    do {
+        intersection_tuples.emplace_back(t_iter);
+        for (size_t i = 1; i < switch_tuple_types.size() - 1; ++i) {
+            t_iter = mesh.switch_tuple(t_iter, switch_tuple_types[i]);
+        }
+    } while (t != t_iter);
```
mtao, line 26: does this general function really work via circulation? i think we might ahve discussed how this is implemented in some nice way, but i don't undertand it right now

mtao (commented):
some things to change / discuss perhaps but generally fine

components/wmtk_components/multimesh_from_tag/wmtk/components/multimesh_from_tag/internal/MultiMeshFromTag.hpp
```diff
@@ -0,0 +1,113 @@
+#pragma once
+
+#include <deque>
+#include <wmtk/TetMesh.hpp>
+#include <wmtk/TriMesh.hpp>
+#include <wmtk/io/HDF5Writer.hpp>
```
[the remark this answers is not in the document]
daniel-zint: Seem like I forgot to clean them up. I removed them now and also a bunch of other unnecessary includes.

src/wmtk/attribute/AttributeManager.hpp
daniel-zint: Correct. It is not in the mesh class at all.

src/wmtk/simplex/tuples_preserving_primitive_types.cpp
daniel-zint: I hope it does. It worked in my few test cases and I cannot think of a case where it does not work but I cannot guarantee that it works all the time.

mtao (approved):
had forgtten one other comment apparently

src/wmtk/Tuple.hpp
daniel-zint: I made the component a friend of Tuple and made the getters private.

PR #748, opened 2024-04-08
enhanced rational class

teseoch:


src/wmtk/utils/Rational.hpp
```diff
@@ -1,195 +1,438 @@
...
+        } else {
+            mpq_init(value);
+            std::string tmp = num + "/" + denom;
+            mpq_set_str(value, tmp.c_str(), 10);
+        }
+    }
 
 private:
     mpq_t value;
+    double d_value;
```
daniel-zint, line 391: I'd give all those members some default values here.

daniel-zint (commented):
Looks reasonable, just some minor comments for now.

daniel-zint (approved):
lgtm. Of course, you need to bring back the other operations in wildmeshing but I trust you with that.

PR #762, opened 2024-05-07
Collapse before simulation

JcDai:


daniel-zint (approved):
Let's just get that monster merged as soon as possible.

PR #771, opened 2024-05-12
Mtao/operator before ptr check

mtao:


src/wmtk/operations/Operation.cpp
```diff
@@ -109,14 +109,20 @@ bool Operation::before(const simplex::Simplex& simplex) const
...
     const Mesh& invariant_mesh = m_invariants.mesh();
 
-    // TODO check if this is correct
-    const std::vector<simplex::Simplex> invariant_simplices =
-        m_mesh.map(invariant_mesh, simplex_resurrect);
-
-    for (const simplex::Simplex& s : invariant_simplices) {
-        if (!m_invariants.before(s)) {
+    if (&invariant_mesh == &mesh()) {
+        if (!m_invariants.before(simplex_resurrect)) {
```
daniel-zint, line 113: This if statement is not required. You can just
```
return m_invariants.before(simplex_resurrect);
```

daniel-zint (approved):
lgtm. Just one minor comment.

src/wmtk/operations/Operation.cpp
mtao: agreed - i've been trying to keep a habit of keeping one return value if possible under the faith that it can induce more RVO when it doesn't obfuscate the code too much (this also helps with automatic return type deduction in templated functions).

PR #778, opened 2024-06-21
Update Operation.cpp

daniel-zint:
The `is_valid(simplex)` check must be performed after the `is_removed` check, because `is_valid` has an assert that checks for `is_removed`.


mtao (approved):
Lgtm

PR #784, opened 2024-07-16
Dzint/fixing integration tests

daniel-zint:
Creating a new integration test is now simpler. The "tests" object in the JSON is automatically created.

### DIFF
--- a/components/tests/integration_guide.md
+++ b/components/tests/integration_guide.md
@@ -1,10 +1,9 @@
 # Guide for Using Integration Test
 
-When you want to start a new integration test, then you need to have following a few steps to set up the test.
+1. Write a JSON file under the folder `/data/unit_test/`, you can refer to those existing JSON files to learn how to write it.
+2. Put the mesh for your test under the folder `/data/unit_test/meshes`.
+3. Add the JSON file's name to `/components/tests/integration_test.cpp`, i.e. add the line `WMTK_INTEGRATION("YOUR TEST NAME", false);`.
+4. The first time you run your integration test, set the `DO_VALIDATION` parameter `false`, i.e. `WMTK_INTEGRATION("YOUR TEST NAME", false);`. The integration test will then add the test output (number of simplices) to your JSON. Afterward, set the parameter to `true`. Now, the integration test will compare the result with your original output and fail if the two do not match.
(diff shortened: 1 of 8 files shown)


components/tests/integration_guide.md
```diff
@@ -1,10 +1,9 @@
 # Guide for Using Integration Test
 
-When you want to start a new integration test, then you need to have following a few steps to set up the test.
+1. Write a JSON file under the folder `/data/unit_test/`, you can refer to those existing JSON files to learn how to write it.
+2. Put the mesh for your test under the folder `/data/unit_test/meshes`.
```
teseoch, line 4: maybe suggest to use one of the existing meshes?

src/wmtk/io/Cache.hpp
```diff
@@ -147,6 +147,11 @@ class Cache
     /// Unsets the mesh held by each cached mm - useful for debugging whether cache loading works
     void flush_multimeshes();
 
+    /**
+     * @brief Get all names of the meshes stored in cache.
+     */
+    std::vector<std::string> mesh_names();
```
mtao, line 153: const?

components/tests/integration_test.cpp
```diff
@@ -109,22 +109,32 @@ int authenticate_json(const std::string& json_file, const bool compute_validatio
...
                 return 2;
             }
             if (n_tetrahedra != expected_tetrahedra) {
                 spdlog::error(
-                    "Violating Authenticate for tetrahedra {}!={}",
+                    "Violating Authenticate in mesh `{}` for tetrahedra {} != {}",
+                    meshes[i],
                     n_tetrahedra,
                     expected_tetrahedra);
                 return 2;
```
mtao, line 140: while we're at it could we remove these random numbers with an enum or something?
also aren't we trying to use wmtk::logger?

mtao (approved):
cool

components/tests/integration_test.cpp
daniel-zint: I can switch to wmtk::logger, yes. Actually, it would be better if we could have catch2 checks with custom messages but AFAIK they don't exist yet.

Please feel free to take a look also at the rest of the code. As I am already at it, I can clean it up a bit more.

PR #793, opened 2024-08-22
cache without folder

daniel-zint:
The cache has now a default constructor. If used, the cache does not write to disk but keeps meshes purely in memory.

### DIFF
--- a/src/wmtk/io/Cache.cpp
+++ b/src/wmtk/io/Cache.cpp
@@ -74,6 +74,16 @@ Cache::Cache(const std::string& prefix, const std::filesystem::path location, bo
     , m_delete_cache(delete_cache)
 {
     m_cache_dir = create_unique_directory(prefix, location);
+    wmtk::logger().info(
+        "Cache stores in folder: {}\nThe folder will be deleted automatically after successful "
+        "completion.",
+        fs::absolute(m_cache_dir));
+}
+
+Cache::Cache()
+    : m_delete_cache(false)
+{
+    wmtk::logger().info("Cache does not write to disk.");
 }
 
 Cache::~Cache()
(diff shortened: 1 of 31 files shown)


src/wmtk/io/Cache.cpp
```diff
@@ -90,48 +126,48 @@ Cache::~Cache()
...
 
-const std::filesystem::path& Cache::create_unique_file(
+const std::filesystem::path& Cache::create_cache_file(
     const std::string& filename,
-    const std::string& extension,
-    size_t max_tries)
+    const std::string& extension)
 {
-    const std::string timestamp = number_to_hex(nanoseconds_timestamp());
+    throw_memory_only_mode();
```
mtao, line 133: is this worth a throw or just a warning?

mtao: either way i prefer the 2 lines of "if memory only mode throw memory only mode" - this code doesn't really imply a branch and the change in control flow isn't obvious

src/wmtk/io/Cache.cpp
```diff
@@ -195,18 +254,32 @@ void Cache::write_mesh(
...
+        } else {
+            p = it->second;
+        }
+        HDF5Writer writer(p);
+        m.serialize(writer, &m);
     } else {
-        p = it->second;
+        // write in meshes list
+        const auto it = m_meshes.find(name);
+        std::shared_ptr<Mesh> pm = const_cast<Mesh&>(m).shared_from_this();
```
mtao, line 273: this const cast shouldn'lt be necessary, as shared_from_this can return a `std::shared_ptr<const Mesh>` ? this occurs later on as well

mtao (approved):
lgtm, minor cleanup comments

PR #795, opened 2024-09-01
Operations Undo Exceptions

mtao:
Previously if an exception were thrown in an operation the operation would not undo properly.
Operations now catch and rethrow exceptions to mark operations as failed if an exception occurs.


mtao:
depends on fixes in mtao/optimize_ops to work

PR #803, opened 2024-09-19
wildmeshing component refactor/ tetwild/ triwild

JcDai:


src/wmtk/invariants/FrozenVertexInvariant.cpp
```diff
@@ -0,0 +1,41 @@
+#include "FrozenVertexInvariant.hpp"
+#include <wmtk/Mesh.hpp>
+
+namespace wmtk::invariants {
+FrozenVertexInvariant::FrozenVertexInvariant(
```
daniel-zint, line 5: If this invariant is only used in your component, please move it to your component. Only move it to the src folder, if it is used in multiple components.

src/wmtk/invariants/FrozenVertexInvariant.hpp
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+#include <optional>
+#include <wmtk/attribute/TypedAttributeHandle.hpp>
+#include "Invariant.hpp"
+
+namespace wmtk::invariants {
+class FrozenVertexInvariant : public Invariant
```
daniel-zint, line 8: Please add a line explaining what this invariant does.

src/wmtk/invariants/FrozenVertexInvariant.cpp
JcDai: will do in the next pr

PR #806, opened 2024-09-26
Refactor shortestedge_collapse API.

daniel-zint:
Refactor shortestedge_collapse API.

Additionally, I fixed the computation of the bbox. It was only checking for x and y coordinates, indipendently of the dimension of the position attribute.

### DIFF
--- a/components/shortestedge_collapse/wmtk/components/shortestedge_collapse/shortestedge_collapse.cpp
+++ b/components/shortestedge_collapse/wmtk/components/shortestedge_collapse/shortestedge_collapse.cpp
@@ -23,15 +23,8 @@
 
 
 namespace wmtk::components {
-std::shared_ptr<Mesh> shortestedge_collapse(
-    TriMesh& mesh,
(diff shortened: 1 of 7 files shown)


JcDai (approved):
lgtm

components/shortestedge_collapse/wmtk/components/shortestedge_collapse/shortestedge_collapse.cpp
```diff
@@ -81,9 +77,9 @@ std::shared_ptr<Mesh> shortestedge_collapse(
 
     //////////////////////////////////
     // computing bbox diagonal
-    Eigen::VectorXd bmin(mesh.top_cell_dimension());
+    Eigen::VectorXd bmin(position_handle.dimension());
```
mtao, line 80: we should have been using Eigen::AlignedBox for bounding boxes to begin with. There's at least 3 sorts of bugs that can happen with manual bbox initialization could have been avoided

PR #816, opened 2024-10-10
IdSimplex ID use with derived meshes + Accessors

mtao:
Made MeshCRTP use any Mesh id overloads in a protected fashion, which enables forwarding to fully derived Mesh types.

Also made accessor access patterns work for any type that `Accessor::index` function exists - implementing a constraint would be nice with cpp20 :-P )


daniel-zint (approved):
👍

PR #825, opened 2024-10-24
Per-Application Integration Test Granularity

mtao:
adding support for doing one integration test suite per application


applications/integration_test.py
```diff
@@ -133,6 +147,7 @@ def test_all(self):
 
     parser.add_argument('-c', '--test_config', help="Path to the json config file")
     parser.add_argument('-b', '--binary_folder', help="Path to the folder that contains the apps binaries")
+    parser.add_argument('-s', '--single', help="Single test to run")
     parser.add_argument('-t', '--test', help="Runs a single test")
```
teseoch, line 151: what is the difference between s and t?

mtao: i hadn't noticed it - cleaned it up in a recent commit i'll push soon

PR #832, opened 2024-10-29
Convert tool to convert wmtk mesh types

mtao:
We currently support 2-3 mesh formats within the toolkit (vtu / hdf5 / msh) and it would be useful to be able to directly create/manipulate meshes using them. This is particularly annoying because hdf5 supports multimesh configurations while others do not - For building examples that use multimesh we can create simple input/output pairs by sticking to hdf5 / use this tool to help build hdf5 files from inputs


applications/convert/test_config.json
```diff
@@ -0,0 +1,7 @@
+{
+    "tests": [
```
teseoch, line 2: we need tests for this

PR #833, opened 2024-10-30
hiding GCC14 warning for null-dereference in eigen

mtao:
adding ignore for warning in eigen that gcc14 catches - also localizing our use of eigen


mtao:
> this will likely disable accelerate

what do you mean?

PR #834, opened 2024-10-30
tetwild triwild cleanup

JcDai:


applications/triwild/triwild_test_config.json
```diff
@@ -0,0 +1,12 @@
+{
+    "test_directory": "unit_test",
+    "tests": [],
+    "release_only_tests": [
+        "triwild_siggraph_icon_out.json"
```
daniel-zint, line 5: Is this slow? If it doesn't take more than a minute, I'd vote for running this test in debug and release.

components/triangle_insertion/wmtk/components/triangle_insertion/triangle_insertion.cpp
```diff
@@ -64,10 +64,10 @@ std::tuple<std::shared_ptr<wmtk::TetMesh>, ChildMeshes> triangle_insertion(
         utils::generate_raw_tetmesh_from_input_surface(V, F, Vbg, Fbg);
 
     ChildMeshes child_meshes;
-    std::shared_ptr<Mesh> surface_mesh = child_meshes.surface_mesh;
-    std::shared_ptr<Mesh> open_boundary_mesh = child_meshes.open_boundary_mesh;
-    std::shared_ptr<Mesh> nonmanifold_edge_mesh = child_meshes.nonmanifold_edge_mesh;
-    std::shared_ptr<Mesh> bbox_mesh = child_meshes.bbox_mesh;
+    std::shared_ptr<Mesh>& surface_mesh = child_meshes.surface_mesh;
```
daniel-zint, line 67: That is really odd, using a reference of a shared pointer.

I'd say, you either create a new shared pointer (no reference) or you dereference directly the Mesh inside:
```
Mesh& surface_mesh = *child_meshes.surface_mesh;
```

scripts/convert_vtu_to_obj.py
```diff
@@ -0,0 +1,23 @@
+import meshio as mio 
+import sys
+
+def __main__():
```
daniel-zint, line 4: Isn't that something `meshio` can do? If not, having such a script is definitely nice, but please add a few lines of documentation, i.e. what does the script do, what are the arguments, and what is the output.

src/wmtk/invariants/Swap2dUnroundedVertexInvariant.cpp
```diff
@@ -0,0 +1,42 @@
...
+    constexpr static PrimitiveType PV = PrimitiveType::Vertex;
+    constexpr static PrimitiveType PE = PrimitiveType::Edge;
+    constexpr static PrimitiveType PF = PrimitiveType::Triangle;
+
+    auto accessor = mesh().create_const_accessor(m_coordinate_handle);
+    int dim = mesh().get_attribute_dimension(m_coordinate_handle);
+
+    // get the coords of the vertices
+    // input face end points
+    const std::array<Tuple, 4> vs = {
```
daniel-zint, line 25: You can use `simplex::faces_single_dimension` for that. Or even better, use `Mesh::orient_vertices()` which is more efficient.

daniel-zint (approved):
lgtm, just some minor clean-up comments

scripts/convert_vtu_to_obj.py
JcDai: meshio cannot convert mesh to obj with "line", I've added some docs

components/triangle_insertion/wmtk/components/triangle_insertion/triangle_insertion.cpp
JcDai: changed

src/wmtk/invariants/Swap2dUnroundedVertexInvariant.cpp
JcDai: I need a vertex in the opposite triangle so the last switch is needed anyway. and this code will be replaced by a swap2d before check, I will change it there

PR #835, opened 2024-10-31
split tests in slow and fast to run the fast one in debug

teseoch:


applications/cdt_sec/cdt_sec_test_config.json
```diff
@@ -1,6 +1,7 @@
 {
     "test_directory": "unit_test",
-    "tests": [
+    "tests": [],
+    "slow_tests": [
```
daniel-zint, line 4: Maybe you want to call these `release_only_tests` or something like that?

JcDai (approved):

PR #836, opened 2024-11-01
Add padding to Tuple so that it aligns with 2*int64_t.

daniel-zint:


teseoch (approved):

mtao (commented):
lgtm, maybe comment that this can be optimized in the tuple_io file though?

PR #840, opened 2024-11-14
Fix is_valid debug messages

mtao:
TriMesh allowed is_valid to assert upon failures in debug mode and also had some debug printouts available but they would only show up after an assertion had happened, making it useless. This PR does the debug message and then assertions to show all of the tuple state before identifying the first part of the tuple that was invalid.

On top of reordering the assert/print in TriMesh, this PR pushes this behvaior to EdgeMesh and TetMesh to all nontrivial mesh types.


daniel-zint (approved):
lgtm

PR #841, opened 2024-11-21
always export vertex attributes in paraview

teseoch:


tests/io/test_io.cpp
```diff
@@ -92,6 +92,16 @@ TEST_CASE("paraview_2d", "[io]")
     mesh->serialize(writer);
 }
 
+TEST_CASE("paraview_2d_vtag", "[io]")
```
daniel-zint, line 95: please add a [.] this is not a real test

daniel-zint (approved):
I trust

PR #851, opened 2025-01-10
Dzint/tetwild msh converter

daniel-zint:


src/wmtk/io/ParaviewWriter.cpp
```diff
@@ -150,13 +150,15 @@ ParaviewWriter::ParaviewWriter(
...
     m_writers[3].init(filename.string() + "_tets.vtu", vertices_name, cells[3], m_enabled[3]);
 
-    paraviewo::VTMWriter vtm;
-    if (m_enabled[0]) vtm.add_dataset("verts", "mesh", filename.string() + "_verts.vtu");
-    if (m_enabled[1]) vtm.add_dataset("edges", "mesh", filename.string() + "_edges.vtu");
-    if (m_enabled[2]) vtm.add_dataset("faces", "mesh", filename.string() + "_faces.vtu");
-    if (m_enabled[3]) vtm.add_dataset("tets", "mesh", filename.string() + "_tets.vtu");
-
-    vtm.save(filename.string() + ".vtm");
+    if (m_enabled[0] + m_enabled[1] + m_enabled[2] + m_enabled[3] > 1) {
```
mtao, line 153: shouldn't we shortcircuit writing at the beginning of this function with a return instead of down here?

mtao (approved):
sure - this is very similar to an application i have lying around in my branch called "convert" - we should discuss merging the two at some point. it does this functionality but i got stuck writing a generic but simple cmdline interface for it

src/wmtk/io/ParaviewWriter.cpp
daniel-zint: Not sure what you mean. This is just to avoid writing the additional VTM file. We still want all the VTUs to be written.

PR #853, opened 2025-01-29
Make ids public.

daniel-zint:
By making the `Tuple` IDs public, all the `friend` stuff in there could be removed. I had to touch a lot of files because of that but, I think the code is cleaner now.


mtao (approved):
delete tupleinspector but other than that lgtm

PR #855, opened 2025-02-12
updating repo locations

mtao:
updated all dataset repos to be the current main after merging them all


CMakeLists.txt
```diff
@@ -167,12 +170,11 @@ if(MSVC)
...
-    # Unit tests
-    enable_testing()
-    # data for unit tests
-    set(WMTK_DATA_ROOT "${CPM_SOURCE_CACHE}/wmtk_data/" CACHE PATH "Where should the toolkit download and look for test data?")
-endif()
+
+# git repo used for wmtk data if it is required.
+# use wmtk_download_data to obtain the path to the data
+set(WMTK_DATA_REPOSITORY https://github.com/wildmeshing/data.git CACHE STRING "Git repo for testing data" )
+set(WMTK_DATA_TAG 6f171d11283d2977913a1b4e6129f9808357baf9 CACHE STRING "git tag for testing data" )
```
daniel-zint, line 177: I like that this is right away in the main CMake file. Makes it very easy to find and update.

applications/CMakeLists.txt
```diff
@@ -10,6 +10,12 @@ include(wmtk_add_application)
 include(wmtk_register_integration_test)
 include(wmtk_generate_test_config)
 
+
+include(wmtk_download_data)
+wmtk_download_data(DATA_DIR 
+    https://github.com/wildmeshing/data.git
+    6f171d11283d2977913a1b4e6129f9808357baf9
```
daniel-zint, line 17: Shouldn't that one use the `WMTK_DATA_TAG` and repo?

cmake/wmtk_download_data.cmake
```diff
@@ -1,30 +1,16 @@
...
-        GIT_REPOSITORY ${GIT_REPOSITORY}
-        GIT_TAG ${GIT_TAG}
-        SOURCE_DIR ${WMTK_DOWNLOAD_DATA_FOLDER}
-    )
-    set(DATA_DIR ${WMTK_DOWNLOAD_DATA_FOLDER} PARENT_SCOPE)
+     CPMAddPackage(
+         NAME wmtk_data
+         GIT_REPOSITORY ${WMTK_DATA_REPOSITORY}
+         GIT_TAG ${WMTK_DATA_TAG}
+         GIT_SHALLOW TRUE
```
daniel-zint, line 9: `GIT_SHALLOW` will cause trouble when someone updates the data folder. That can be really annoying and I would recommend not using it if we want to pull a specific commit.

daniel-zint (changes requested):
I think you forgot to use the data and repo variables. Also `GIT_SHALLOW` should not be used here in my opinion.

applications/CMakeLists.txt
mtao: was an accident. am deleting.

cmake/wmtk_download_data.cmake
mtao: changing it

PR #856, opened 2025-03-25
Dzint/test delete attribute

daniel-zint:
Adding some validation functionality for handles and attributes. The behavior right now is not what I would like but we need to discuss what behavior we want.

My suggestion:
* The attribute vector keeps on growing by default. A `clear` should rarely happen, although having the option might be nice.
* The handles map should stay up-to-date, and deleted attributes should be deleted from the handles map instead of resetting the pointers.
* Attribute handles store the attribute name. That is convenient for debugging and verifying the validity of a handle.


mtao (approved):
lgtm; updated logic to not remove attributes, but leaving clear attributes in case someone feels like it.

merge when you feel liek it

PR #857, opened 2025-04-01
Attribute Documentation

daniel-zint:


src/wmtk/attribute/TypedAttributeHandle.hpp
```diff
@@ -50,7 +50,6 @@ class TypedAttributeHandle
     TypedAttributeHandle(const MeshAttributeHandle&);
 
 public:
-    using value_type = T;
```
mtao, line 53: why remove this? it's pretty potentially useful for template metaprogramming

src/wmtk/attribute/TypedAttributeManager.cpp
```diff
@@ -296,74 +320,22 @@ void TypedAttributeManager<T>::remove_attribute(const AttributeHandle& attribute
...
-            logger().warn(
-                "Attribute name is not the same as the name in the handles map. Attribute: {}, "
-                "handles map: {}",
-                attr_name,
-                name);
-            return false;
-        }
-
-        return true;
+    if (!m_attributes[handle.index()]) {
```
mtao, line 334: I get cleaning up the number of lines in some sense, but the debug info was sorta nice I thought?

src/wmtk/operations/utils/VertexLaplacianSmooth.cpp
```diff
@@ -24,7 +23,7 @@ bool VertexLaplacianSmooth::operator()(Mesh& mesh, const simplex::Simplex& simpl
 
     std::visit(
         [&](const auto& handle) {
-            using T = typename std::decay_t<decltype(handle)>::value_type;
+            using T = typename std::decay_t<decltype(handle)>::Type;
```
mtao, line 26: oh i see, i made a previous comment about value_type. i guess we already had this type name. lately i've been more into using the STL type value_type for externally visible typedefs and Type for internal ones, i don't think we have a general policy on this :shrug:

tests/attributes/test_accessor.cpp
```diff
@@ -451,10 +432,167 @@ TEST_CASE("validate_handle", "[attributes]")
...
-    CHECK(m.validate_attributes());
     CHECK_FALSE(a1.is_valid());
     CHECK_FALSE(a2.is_valid());
     CHECK_FALSE(a3.is_valid());
     CHECK_THROWS(m.get_attribute_handle<double>("a1", wmtk::PrimitiveType::Vertex));
 }
+
+namespace {
+// template magic for overload pattern
+template <class... Ts>
```
mtao, line 449: the overload should go into the main utils folder perhaps (in hte metaprogramming namesapce)?/

mtao (approved):

src/wmtk/attribute/TypedAttributeHandle.hpp
daniel-zint: A few lines above there is a `using Type = T;`. This one was just redundant.

src/wmtk/attribute/TypedAttributeManager.cpp
daniel-zint: That just looks weird here in Github. I removed `validate()` because it was basically checking if the map was consistent with the attribute vector. Now the map does not exist anymore, so this function doesn't do anything. The debug info is still there in `validate_handle`.

tests/attributes/test_accessor.cpp
daniel-zint: I honestly do not like this version. I just wanted to show some different options for writing the visit. I think the `PrintVisitor` and `PrintVisitorT` are the better examples. They are much easier to understand. If you want to use this version then yes, we should move it to the utils folder.

PR #924, opened 2026-06-19
Python bindings: integration tests, Release build, and PyPI wheel CI

danielepanozzo:
## Summary
Packaging work to publish the `pywmtk` bindings on PyPI.

- **Integration tests** (`tests/python/`) mirroring `tests/integration_tests.cpp`: the same `data/integration_tests` inputs are driven through the binding the way the C++ `wmtk_wrapper`/`app/main.cpp` does it (parse JSON → inject `json_input_file` → dispatch on `application`).
  - Serial inputs (`num_threads` <= 1) are deterministic → output compared **byte-for-byte** vs native `wmtk_app` (logs excluded).
  - Multithreaded inputs (`num_threads` > 1, e.g. `tetwild_octocat`) → **termination only** (parallel scheduler is non-deterministic).
  - When `wmtk_app` is absent (e.g. testing an installed wheel) the suite still verifies every app runs end-to-end on real data; the byte comparison is reported as skipped via a warning.
  - `test_smoke.py`: data-free import/dispatch check.
- **Release build** for the published bindings (no debug symbols) in `pyproject.toml`.
- **PyPI wheel/sdist CI** — new `.github/workflows/wheels.yml`:
  - `build_wheels`: `cibuildwheel` on **macOS arm64** (`macos-14`), **Windows x86_64** (`windows-2022`), **Linux x86_64** (`ubuntu-latest`), CPython 3.9–3.13.
  - `make_sdist`: source distribution.
  - `test_wheels`: installs each freshly built wheel and runs `tests/python` on real data to verify correctness.
  - `publish`: uploads wheels + sdist to PyPI on a GitHub Release (Trusted Publishing).
  - `pyproject.toml` `[tool.cibuildwheel]`: skip pypy/musllinux/32-bit/win-arm; per-wheel smoke test.

## Verified locally
- C++/Python result parity: `31 passed` (30 byte-for-byte vs `wmtk_app`; `tetwild_octocat` termination-only).
- Wheel-CI no-app path passes with warning; smoke tests pass.
- `cibuildwheel --print-build-identifiers` confirms `cp39–cp313 × {macosx_arm64, win_amd64, manylinux_x86_64}`.
- sdist builds (`python -m build --sdist`).

## Publishing
The `publish` job runs on a published GitHub Release and uses PyPI **Trusted Publishing** (no token in repo). See PR discussion / commit for the one-time PyPI setup. An API-token fallback is documented inline in the workflow.


tests/python/_run_binding.py
```diff
@@ -0,0 +1,46 @@
+"""Subprocess helper: run a single integration-test JSON through the pywmtk binding.
+
```
daniel-zint, line 2: I don't know what this is used for or if it is necessary.

tests/python/test_integration.py
```diff
@@ -0,0 +1,263 @@
...
+#
+# The byte-for-byte comparison is only meaningful when the reference app and the
+# binding are built from the same configuration: the iterative geometry code is
+# sensitive enough that two independently-configured build trees (e.g. a stock
+# `cmake build/` vs. the scikit-build `build/<config>/` tree the wheel is built
+# in) can take divergent discrete decisions on the heaviest cases. So we prefer
+# the `wmtk_app` that scikit-build co-builds alongside the binding under
+# `build/<config>/app/`, and only fall back to a top-level `build/app/`.
+# Override explicitly with WMTK_APP.
+def _find_reference_app():
```
daniel-zint, line 64: I dislike the byte-for-byte comparison as it makes the Python tests so messy. I'd say we should just run the integration tests.

tests/python/test_integration.py
```diff
@@ -0,0 +1,263 @@
...
+
+# Subprocess helper that loads a JSON and calls pywmtk.wmtk (see _run_binding.py).
+RUN_BINDING = HERE / "_run_binding.py"
+
+# Per-run wall-clock cap (seconds) so a hang fails loudly instead of stalling.
+RUN_TIMEOUT = 1800
+
+# Exactly the `input_files` list from tests/integration_tests.cpp.
+# (image_simulation_energies_2d is commented out there as well.)
+INPUT_FILES = [
```
daniel-zint: This list is now duplicated. We should probably make it a JSON and refer to it in both integration tests.

tests/python/test_integration.py
```diff
@@ -0,0 +1,263 @@
...
+from pathlib import Path
+
+import pytest
+
+HERE = Path(__file__).resolve().parent
+REPO_ROOT = HERE.parents[1]  # tests/python -> tests -> repo root
+
+# Test data (same directory the C++ integration tests read from).
+DATA_DIR = Path(
+    os.environ.get("WMTK_DATA_DIR", REPO_ROOT / "data" / "integration_tests")
```
daniel-zint, line 47: I'd not check environment variables. This is different to what the C++ tests do and could lead to inconsistencies. Just using the relative path should be fine.

tests/python/test_integration.py
```diff
@@ -0,0 +1,263 @@
...
+    "topological_offset_3d_edge_input.json",
+    "topological_offset_3d.json",
+    "manifold_extraction_3d.json",
+    "triwild_puzzle.json",
+]
+
+# Applications whose `output` default carries a ".obj" extension. The others
+# (tetwild, image_simulation, topological_offset, ...) take an extension-less
+# base name and append their own (".msh", "_surface.obj", "_<n>.vtu", ...).
+_OBJ_OUTPUT_APPS = {"isotropic_remeshing", "qslim", "shortest_edge_collapse"}
```
daniel-zint: One more reason not to compare results with the C++ API.

PR #930, opened 2026-07-06
Spenbert/offsets

Spenbert02:
topological_offsets component updated json and bugfix


Spenbert02:
@daniel-zint you may want to review this briefly. I modified the interfaces for both components slightly. Manifold extraction now inherits from the TopoOffsetMesh classes directly which is cleaner. The missing tets bug I fixed by changing the (BVH distance - target_distance) root finding method. This is heuristic but it works better

PR #939, opened 2026-07-16
skip good regions when smoothing

danielepanozzo:
Stacked on #938. Smoothing runs on **every vertex every pass**, but a vertex surrounded by already-good tets barely moves — its smoothing is wasted work. This restricts vertex smoothing to vertices incident to an **active** tet (energy ≥ `skip_good_regions_margin · stop_energy`).

## Result
On the spencer meshes: **smoothing passes 14–16× faster** (e.g. 55928: 1.07s → 0.068s), ~**10% total** wall-time, **no change to the result and no element-count growth**.

## Only smoothing is gated (the important finding)
I tried gating the topology/sizing ops (split/collapse/swap) too — it's **not safe**. Skipping them starves the optimizer → the mesh stalls → the stuck-refine (num_worst=50) fires and **blows up the element count** (crown 8–18×, mesh/run-dependent). Dilating the active region by tet-rings didn't fix it (non-monotonic, variance-dominated). Smoothing changes no topology or sizing, so skipping good regions can never starve the optimizer or bloat the mesh — it's the one genuinely-wasteful op on good regions.

## Details
- `active_vertices()` — vids incident to a tet with `m_quality ≥ active_quality_threshold()` (`= (margin·stop_energy)³`, since `m_quality` is AMIPS³).
- `skip_good_regions` (default **true**) + `skip_good_regions_margin` (0.9), JSON-configurable; env override `TW_SKIP_GOOD_REGIONS`.
- `test_sizing_refine.cpp`: `active_vertices` returns exactly the bad tet's vertices / none when all-good / all when all-bad.

Full tetwild suite **735 assertions** green; clang-format clean. 7 files.


components/tetwild/wmtk/components/tetwild/tetwild.cpp
```diff
@@ -161,6 +161,13 @@ TetWildMesh::ExportStruct tetwild_with_export(nlohmann::json json_params)
     if (const char* e = std::getenv("TW_STUCK_REFINE"))
         params.stuck_refine = (std::string(e) != "0");
 
+    // Skip good regions.
+    params.skip_good_regions = json_params.value("skip_good_regions", params.skip_good_regions);
+    params.skip_good_regions_margin =
+        json_params.value("skip_good_regions_margin", params.skip_good_regions_margin);
+    if (const char* e = std::getenv("TW_SKIP_GOOD_REGIONS"))
```
daniel-zint: I dislike the environment variable. I don't see why we should ever want that.

components/tetwild/wmtk/components/tetwild/tetwild.cpp
```diff
@@ -161,6 +161,13 @@ TetWildMesh::ExportStruct tetwild_with_export(nlohmann::json json_params)
     if (const char* e = std::getenv("TW_STUCK_REFINE"))
         params.stuck_refine = (std::string(e) != "0");
 
+    // Skip good regions.
+    params.skip_good_regions = json_params.value("skip_good_regions", params.skip_good_regions);
```
daniel-zint: This can be written in a more compact way:
`params.skip_good_regions = json_params["skip_good_regions"];`

components/tetwild/wmtk/components/tetwild/tetwild.cpp
```diff
@@ -161,6 +161,13 @@ TetWildMesh::ExportStruct tetwild_with_export(nlohmann::json json_params)
     if (const char* e = std::getenv("TW_STUCK_REFINE"))
```
daniel-zint: This is probably from another commit but I'd like to remove this environment variable stuff again.

components/tetwild/wmtk/components/tetwild/TetWildMesh.cpp
```diff
@@ -1046,6 +1046,24 @@ std::tuple<double, double> TetWildMesh::get_max_avg_energy()
     return std::make_tuple(std::cbrt(max_energy), avg_energy);
 }
 
+std::vector<size_t> TetWildMesh::active_vertices() const
+{
+    const double thr = active_quality_threshold();
+    std::vector<char> seen(vert_capacity(), 0);
+    std::vector<size_t> out;
+    for (size_t i = 0; i < tet_capacity(); ++i) {
```
daniel-zint, line 1173: Could be done with a `parallel_for`. Might not be worth it, though. In that case, it would be better to iterate over vertices and check the quality of incident tets. However, this is probably the wrong place to optimize.

daniel-zint (commented):
I'll deal with these things myself. I just wanted to write them down somewhere

PR #943, opened 2026-07-17
tetwild: exact-rational split fallback for worst-tet edges

danielepanozzo:
When an edge split's rounded (double) midpoint would invert an incident tet, `split_edge_after` currently rejects the whole split (`"Exact-rational fallback is forbidden"`). Around a stuck worst element that can block refinement.

## Change
For edges of the **current worst-tet set** (recorded by `refine_sizing_around_worst` in `m_worst_region_vids`), fall back to the **exact rational midpoint** of the two endpoints when the double midpoint inverts. That point lies on the shared edge, so it can never invert a previously-valid incident tet — the split always succeeds. The new vertex stays un-rounded (`m_pos` exact, `m_is_rounded=false`) until a later `round()` reclaims it. A pre-existing inverted incident tet is still rejected (re-checked in exact arithmetic).

Gated by the new `stuck_refine_rational_split` parameter (default true).

## Note
This is a conservative safety net: in practice the double midpoint essentially never inverts a worst-region split on the tested meshes, so it rarely fires — but it removes a hard failure mode and costs nothing when it doesn't fire.

Third of a 4-PR stack (VR-bump -> parallelize -> **rational-fallback** -> force-split).


daniel-zint:
I think we should just bring back the rational split without the check for the worst region. I don't see any benefit in that check, and it requires additional bookkeeping, which makes the code harder to read and maintain.

PR #952, opened 2026-07-21
Concurrency cleanup

danielepanozzo:
## Concurrency cleanup

Three independent commits over the concurrency shim left behind by the TBB removal. Each builds and passes the suite on its own.

### 1. Collect parallel results without a lock
`tbb::concurrent_vector` was only ever used to collect results out of a parallel loop, and most of those sites need no lock at all.

The four edge collectors (isotropic remeshing's collapse/split/swap, qslim's collapse) already know a unique slot per result: `for_each_edge` invokes the callback only for the triangle that *owns* an edge, and tests `tup.eid(m)` to decide that. Writing into a preallocated slot indexed by `eid` means no two threads ever touch the same element, so `indexed_collector` needs no synchronisation whatsoever — and the output order becomes slot order rather than whichever thread won the mutex, removing a source of multithreaded irreproducibility.

- `is_matched` needed no wrapper at all — `match_tet_faces_to_triangles` fills it from a plain serial loop — so it is a `std::vector<bool>`.
- `to_revert` becomes a flag per vertex. A vertex is reachable from several tets, so the old code took the lock once per push and then pushed the same vid repeatedly; every write now stores the same value to a slot indexed by vid, so relaxed ordering suffices and the duplicates collapse for free.

What remains of `concurrent_vector` (the failure list in `LocalizedRetry`, the sub-edge list in `split_remeshing`) keeps its mutex, over a `std::vector` rather than a deque now that no caller holds a reference across an append. Per-thread buckets would drop that lock too, but this shim's `enumerable_thread_specific` cannot enumerate its slots — they live in `thread_local` storage, so the owner cannot reach other threads' copies — and both sites append at most once per *failed mesh operation*, where the lock is nowhere near the cost of the operation that preceded it.

**Measured** on crown (tetwild, 10 threads): insertion **8.497s** against a baseline of 8.850 / 8.748 / 9.007s over three runs — below all of them. Only insertion is worth quoting: total runtime varies by **32%** run to run at this thread count (406–535s over three identical runs), so it cannot resolve a change this size.

### 2. Pass the thread count to `parallel_for` instead of `task_arena`
The `task_arena` shim held no threads. `execute()` wrote its concurrency into a `thread_local`, ran the lambda, and restored it; `parallel_for` read that `thread_local` to size its pool. A `parallel_for`'s width therefore depended on invisible caller state, and silently fell back to `hardware_concurrency()` anywhere the arena had not been entered — including inside threads spawned by `task_group`, whose `thread_local` copy is its own zero-initialised one, so the two sites that wrapped a `task_group` in an arena were getting nothing from it.

The count is now an explicit trailing argument of `parallel_for`, and the shim is gone. 55 call sites say what width they want where they ask for it. No behaviour change: the resolution rule is unchanged (`> 0` uses that many threads, otherwise `hardware_concurrency`).

### 3. Drop `parallel_sort`; give the morton sort a total order
`wmtk::parallel_sort` was a two-line forward to `std::sort`, so the name claimed something the code did not do, and none of its eight callers is inside a `parallel_for` — each sorts a vertex list once to assign morton-order partitions.

The comparator was the real problem:

```cpp
return (a.morton < b.morton);
```

Morton codes are quantised, so distinct vertices routinely share one. Tied elements compare equal, `std::sort` is not stable, and the order it leaves them in **differs between libc++, libstdc++ and MSVC** — and that order decides which partition, and therefore which thread, each vertex belongs to. It now breaks the tie on the original index.

Same class of bug as the non-stable sorts previously fixed in `TetMesh::get_edges` and `sort_edges_by_length`. It stayed hidden because the reproducibility tests run single-threaded, where there is one partition and the order cannot matter.

### Why not the C++17 parallel algorithms?
Worth stating explicitly, since `std::sort(std::execution::par, ...)` and `std::for_each(std::execution::par, ...)` look like they would make the whole shim redundant. Three reasons they do not:

1. **On Linux they resurrect TBB.** libstdc++ implements the parallel execution policies on top of Intel TBB — `<execution>` with a parallel policy requires linking it. This repository has just removed TBB (#951) because nothing needed it; adopting `par` would bring it straight back on the platform where it is most standard.
2. **On macOS they may not exist.** libc++'s PSTL support is partial and recent; on many Apple clang versions the parallel policies are simply absent. MSVC is the one platform with a solid implementation.
3. **They cannot honour `num_threads` — the disqualifying one.** `std::execution::par` offers no way to request a specific width; the thread count is implementation-defined and unspecified. This codebase depends on it everywhere: `num_threads: 1` is what makes the golden-hash tests reproducible. Under `par` you cannot ask for exactly one thread, so single-threaded determinism would be unachievable.

There is also a shape mismatch: `parallel_for` hands its callback a **chunk** (`blocked_range`), and several sites exploit that to accumulate into a per-chunk local vector and merge once. `std::for_each(par, ...)` is per-element, so each would need restructuring for a weaker guarantee.

So the shim is not redundant with C++17 — it exists to do the one thing the standard algorithms explicitly will not, namely let the caller fix the thread count. `parallel_sort` was the exception: it did no work the standard did not already do, which is why it goes.

### Testing
51/51 unit tests and `Integration_Tests` pass, at each commit.

Stacked on #951.

### DIFF
--- a/components/isotropic_remeshing/wmtk/components/isotropic_remeshing/IsotropicRemeshing.cpp
+++ b/components/isotropic_remeshing/wmtk/components/isotropic_remeshing/IsotropicRemeshing.cpp
@@ -430,11 +430,16 @@ bool IsotropicRemeshing::collapse_remeshing(double L)
     igl::Timer timer;
     timer.start();
     auto collect_all_ops = std::vector<std::pair<std::string, Tuple>>();
-    auto collect_tuples = wmtk::concurrent_vector<Tuple>();
+    // One slot per (triangle, local edge): for_each_edge only invokes the callback for
+    // the triangle that owns the edge, so eid() is unique and no two threads write the
+    // same slot -- no lock, and the order is the slot order rather than whichever thread
+    // won a mutex.
+    auto collect_tuples = wmtk::indexed_collector<Tuple>(3 * tri_capacity());
 
-    for_each_edge([&](auto& tup) { collect_tuples.emplace_back(tup); });
-    collect_all_ops.reserve(collect_tuples.size());
-    for (auto& t : collect_tuples) collect_all_ops.emplace_back("edge_collapse", t);
+    for_each_edge([&](auto& tup) { collect_tuples.set(tup.eid(*this), tup); });
+    const auto collected = collect_tuples.compact();
+    collect_all_ops.reserve(collected.size());
+    for (const auto& t : collected) collect_all_ops.emplace_back("edge_collapse", t);
     wmtk::logger().info(
         "***** collapse get edges time *****: {} ms",
         timer.getElapsedTimeInMilliSec());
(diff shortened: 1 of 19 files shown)


daniel-zint:
This was integrated in #934

PR #958, opened 2026-07-30
Port the recent tetwild work to triwild

danielepanozzo:
Over the last three weeks tetwild received a batch of performance, robustness and sizing-field work. triwild only got the parts that lived in `src/wmtk` (the TBB removal and the VolumeRemesher 2D arrangement), so it was still running the pre-July optimizer: serial op-list construction, brute-force retry loops, the old global KNN sizing field, no force-split, no skip-good-regions, and an unguarded rational split fallback.

Meanwhile simwild had received a **second copy** of the same tetwild machinery. Porting to triwild naively would have made three, so the shared algorithms move into `src/wmtk` first and all three applications are rewired onto them.

## Deduplication first (`20e1264`)

New `src/wmtk/utils/SizingField.hpp` holds what was identical between tetwild and simwild: `select_worst_cells`, `longest_edge`, `grow_vertex_region`, `apply_sizing_refinement`, `gradation_smooth_sizing`, `active_vertices`. Each application keeps its own `refine_sizing_around_worst` driver, because simwild additionally caps the sizing by its per-tag sizing field and treats surface vertices as always active.

`ParallelCollect` becomes dimension-generic at the same time: `TetMesh` and `TriMesh` both gain `cell_capacity()` / `tuple_from_cell()` / `EDGES_PER_CELL`, so `parallel_collect_edge_ops` works on a `TriMesh` too with no change at the existing call sites.

Net −292/+139 lines across the two existing apps, and a dead `max_sizing_scalars` pass dropped from simwild.

## What triwild gains

- **Scheduling** (`34523bb`): parallel op-list prepare and `run_localized_to_convergence` for split/collapse/swap, replacing the loop that rebuilt the whole op list from `get_edges()` after every collapse pass. Smoothing gains `skip_good_regions`.
- **Sizing field** (`00f74d2`): `refine_sizing_around_worst` with the relative stall test and cooldown, replacing the `M = 2` counter and its hardcoded 5e-1 / 10% thresholds. Force-split of each worst triangle's longest edge. `adjust_sizing_field_serial` stays compiled for comparison, as in tetwild and simwild.
- **Options** (`8169f69`): `filter`, `skip_winding_number`, `remove_duplicate_eps`, `DEBUG_hausdorff` and the 11 optimizer knobs, all with tetwild's names and defaults. The finalize phase now reuses the input matrices read when the initial mesh was built instead of parsing every file twice, through a new parallel `wmtk::utils::winding_number_2d`.
- **Tests** (`3da5e95`): the `tests/` subdirectory was commented out in triwild's CMakeLists; it now exists and mirrors tetwild's.

## Bugs fixed on the way

- The rational split fallback fired unconditionally and never re-checked its result, so a split could leave an inverted triangle behind. It is now gated and re-runs the inversion check in exact arithmetic.
- `swap_all_edges()` returned `true` from a `size_t` function, so the `if (cnt_success == 0) break;` in `local_operations` could never fire.
- `preallocation_factor` was read by eight components via `set_preallocation_factor_from_json` but declared in no spec, so jse rejected any JSON that set it. It was dead everywhere; both specs now declare it.
- `flood_fill` kept its visited set in a `std::map` keyed by fid, and `local_operations` started its timer once so the logged per-operation times were cumulative.

## JSON alignment

The option sets now differ by exactly six keys, all genuinely 3D-only: `skip_simplify`, `use_sample_envelope`, `use_legacy_code`, `allow_surface_swap`, `check_surface_topology`, `DEBUG_euler`. 31 shared. tetwild gains `input_names` and `write_vtu`; triwild gains everything else. simwild already used these names, so all three now line up.

## Measured

Same input, same config, serial, puzzle.obj at `length_rel` 4e-3 / `stop_energy` 4:

| | before | after |
|---|---:|---:|
| total | 0.99s | **0.51s** |
| smoothing | 0.505s | 0.061s |
| collapse | 0.147s | 0.070s |
| split | 0.186s | 0.229s |
| final max energy | 3.81 | 3.58 |

Split does more work because it now retries to convergence, and pays for it in quality.

## Two things to know

**`stuck_refine_force_split` is sharp in 2D.** The AMIPS2D energy of an equilateral triangle is 2, so with the default `stuck_refine_num_worst: 0` a `stop_energy` near that floor makes "every triangle above the filter energy" mean nearly the whole mesh — 2382 gate-bypassing splits in one pass in one measurement, turning a 4.95s run into minutes. The 3D defaults are safe because `stop_energy` 10 sits far above the AMIPS3D floor of 3. tetwild's defaults are kept as-is and the hazard is documented in the spec; triwild may want `stuck_refine_num_worst > 0`.

**triwild's multithreaded path is pathological, and was before this branch**: 7.75s at 8 threads versus 0.99s serial on the same input. Worth a separate look.

Vertex-order caching (tetwild's cached `m_order`) is deliberately not ported: sampling puts `get_order_of_vertex` at 0.0% of the profile against 95% in `smooth_after`'s polysolve solve.

## Verification

- Full integration suite: 33/33 inputs pass.
- `tetwild_sphere` and `tetwild_double_sphere` (serial, hence deterministic) are **byte-identical** to the branch point — `out_final.msh`, `out_surface.obj`, `out.vtu`, `out_surf.vtu`.
- New `wmtk_test_triwild`: 55 assertions in 7 cases. `wmtk_test_simwild`: 744 assertions. `wmtk_test_tetwild`: same 733/4 result as the branch point.
- Two pre-existing failures confirmed against the base commit and untouched: `test_surface_swap` (4 assertions) and `test_sec` (3 assertions).


components/triwild/wmtk/components/triwild/TriWildMesh.h
```diff
@@ -17,6 +17,8 @@
 
 #include "Parameters.h"
 
+#include <set>
```
daniel-zint, line 20: Why is `set` included here?

daniel-zint (approved):
lgtm

PR #959, opened 2026-07-30
wmtk: share the Morton partitioner and the surface-topology signature

danielepanozzo:
Stacked on #958. Three deduplications across the wild-meshing components, none of which changes any output.

## Morton partitioning — 6 copies → 1

tetwild, simwild (tet and tri), triwild, qslim and isotropic_remeshing each carried their own ~85-line copy of the same routine, every one of them under a `TODO: this should not be here, but inside wmtk`.

The shared implementation **already existed** as `wmtk::partition_vertex_morton`, but had exactly one caller (`tetwild.cpp`, for the input partition) — the mesh classes never adopted it. They do now.

The four 3D copies were line-for-line the shared function. The two 2D copies differ only in dropping the z terms, which is exactly what feeding the shared 3D function a zero z does: z contributes nothing to the bounding box, nothing to the scale (a `fabs` term under a `max` with two non-negative terms) and a constant zero to the Morton code. So all six produce the same partition ids as before — which the byte-compare confirms.

`partition_utils.hpp` also stops including `TetMesh.h`, which it never needed and which the TriMesh-based callers should not be pulling in.

## Surface topology signature — 2 copies → 1

tetwild's and simwild's `SurfaceTopology.cpp` were **byte-identical after renaming the class** — 161 lines each, and both live (called from their respective `EdgeSwapping.cpp` under `check_surface_topology`). Now one header-only implementation in `wmtk/utils/SurfaceTopology.hpp`, parameterised on the mesh and a `bool(fid)` surface predicate. Each application keeps its `surface_topology_signature()` / `warn_if_surface_topology_changed()` members as two-line forwarders, so call sites and `test_surface_swap.cpp` are untouched. Both `.cpp` files go away.

## Dead duplicated code

- simwild carried a verbatim copy of the free function `bfs_orient` that nothing in simwild ever called — tetwild's copy is the live one.
- `check_attributes` was duplicated in both and had **no caller at all**; the only references are commented out in two tetwild tests.

Both removed. git has them if they are wanted back.

**Not removed:** `adjust_sizing_field_serial` — four copies, 387 lines, one live call site. Three of those call sites are commented out *by design*: it was deliberately re-added in bb5a1d6 as an A/B baseline for the new worst-element refinement. Retiring it is a call for whoever is still comparing against it, not a mechanical cleanup.

## Verification

**−1014 / +182 lines**, and byte-identical output.

- **25 output files across 17 deterministic configs** — 2 tetwild, 2 triwild, 13 simwild, all `num_threads: 0` and hence reproducible — compared file by file against the same runs before the change. Zero differences, checked after each of the three steps and again after clang-format.
- Full integration suite: 33/33.
- All component unit tests match their previous results exactly, including the two pre-existing failures (`shortest_edge_collapse` 18/3, `tetwild` 733/4).


daniel-zint (approved):
lgtm

PR #961, opened 2026-07-31
sec: collapse onto a frozen vertex instead of rejecting the collapse

danielepanozzo:
Stacked on #960. Follow-up to a question about the freeze rule in `ShortestEdgeCollapse` — the check was real, and stronger than it needed to be.

## What was there

`collapse_edge_before` rejected a collapse whenever **either** endpoint was frozen:

```cpp
if (vertex_attrs[t.vid(*this)].freeze || vertex_attrs[t.switch_vertex(*this).vid(*this)].freeze)
    return false;
```

It was not gratuitous. `collapse_edge_after` always places the survivor at the midpoint, which moves *both* endpoints, so any collapse touching a frozen vertex would have moved it.

But rejecting is stronger than necessary. When exactly one endpoint is frozen, placing the survivor at **that endpoint's** position preserves it exactly and the collapse is fine. Only two frozen endpoints genuinely cannot be merged — and that case must stay rejected, because it is what stops the outline of an open surface retracting along itself. The envelope cannot catch that: it is a containment test, not a coverage test, so a boundary sliding inward along the surface stays "inside" it.

## What it is now

Cache both freeze flags; reject only when both are set; otherwise place the survivor on the frozen endpoint when there is one, midpoint otherwise. The survivor inherits the freeze, since it now stands where the frozen vertex stood.

## Measured

Two open patches, collapsed to exhaustion (new `test_sec_boundary.cpp`):

| patch | | vertices left | successes | failed attempts |
|---|---|---:|---:|---:|
| 21×21 square (441 V, 80 outline) | before | 81 | 360 | 1226 |
| | **after** | **80** | 361 | **454** |
| 3×60 strip (180 V, 122 outline) | before | 123 | 57 | 953 |
| | **after** | **122** | 58 | **780** |

Both now reach the floor — the frozen outline and nothing else.

**A correction to my own earlier claim.** When I ported this rule to 2D in #960 I wrote that rejecting outright "leaves a ring of short segments around every junction". That is wrong, and the measurement above shows why: interior vertices can always collapse into *each other*, so only the last one is stranded. The gain is one vertex per frozen-bounded region, not a ring. The larger effect is wasted work — the executor kept retrying collapses that could never succeed, 63% of the failed attempts on the square patch. The 2D comments are corrected in this PR too, and the two implementations now follow the same rule.

## Verification

- **No change to any existing result**: all 23 output files across the 15 deterministic tetwild and simwild integration configs are byte-identical. Every one of those inputs is closed, so `freeze_boundary` marks nothing and the new branch is never taken — which is also why this needed a dedicated open-surface test to measure at all.
- Full integration suite 33/33; all component test binaries unchanged.
- The three failures in `test_sec.cpp` are pre-existing and untouched: that test asserts `out_v.size() == 9` on a hand-written list of 5 vertices.


daniel-zint (approved):
lgtm

PR #962, opened 2026-08-01
Apply preserve_topology consistently to input de-duplication

danielepanozzo:
Stacked on #961. `preserve_topology` was consulted by the optimizer's collapse and by the final Euler-characteristic check, but not by the de-duplication that runs before them — and it turned out not to be enforced at all by default.

## 1. Input vertex de-duplication ignored it

`remove_duplicate_eps` was passed to `read_triangle_mesh` / `read_edge_mesh` unconditionally, and `igl::remove_duplicate_vertices` merges everything within that radius. That is a topology change, not a de-duplication: it welds sheets of the surface that pass close to each other, so handles and tunnels disappear before any other stage sees the mesh.

The effect is not subtle. `Octocat.obj` is a closed genus-0 surface, and at the default tolerance the mesh that reaches the rest of the pipeline has Euler characteristic **156** instead of 2:

| preserve_topology | dedup tolerance | input euler characteristic |
| --- | --- | --- |
| false | 2e-3 | **[156]** |
| **true** | **0** | **[2]** |

Under `preserve_topology` the tolerance is now 0 — only exactly coincident vertices are merged, which is a genuine de-duplication of the file and leaves the topology alone.

simwild already made this distinction ([`read_image_msh.cpp`](components/simwild/wmtk/components/simwild/read_image_msh.cpp)); tetwild and triwild did not. That inconsistency is the bug.

## 2. The `preserve_topology` check itself was vacuous

The throw compared `ecs_input` against `ecs_output`, but both were only computed under `DEBUG_euler`. With `DEBUG_euler` off — the default — it compared two empty vectors and passed whatever had happened to the topology. They are now computed whenever `preserve_topology` is set, which is what the comment already claimed.

## 3. The simplification's topology check stays on regardless — and now says so

`TriMesh::collapse_edge_before` is the classical link condition, and gating it on `preserve_topology` looks right at first: it is exactly what stops the simplification closing a thin tunnel or welding two sheets.

But it is not only a topology guard. It is the only check between `collapse_edge_conn` and a connectivity that `wmtk::TriMesh` cannot represent, since nothing downstream re-checks it — `invariants()` only tests the envelope. Relaxing it does not produce a valid mesh with different topology; it produces a corrupt one.

So it is applied unconditionally, and tetwild logs a warning when the caller asked for topology changes and will not get them:

```
TODO the simplification still preserves topology as the toolkit does not support
non-manifold meshes, to fix
```

Lifting this needs a mesh data structure that can represent non-manifold connectivity. A `check_mesh_connectivity_validity()` call now follows the simplification, so any future relaxation fails loudly instead of carrying on with a corrupt mesh.

triwild's 2D counterpart in `simplify_segments` — merging two frozen vertices, which fuses two junctions or closes an open curve — has no such constraint, since a segment network with junctions of any valence is representable. It is gated on `preserve_topology` as intended, and the frozen flags are recomputed after a collapse because merging two of them no longer leaves the surviving vertex's valence unchanged.

## Testing

`tetwild_sphere`, `tetwild_octocat`, `tetwild_double_sphere`, `triwild_puzzle` and `triwild_collection` all pass with `throw_on_fail`.

Default behaviour changes for anyone relying on `remove_duplicate_eps` while also setting `preserve_topology`; the simplification behaves as before in all cases.

### DIFF
--- a/components/shortest_edge_collapse/wmtk/components/shortest_edge_collapse/ShortestEdgeCollapse.cpp
+++ b/components/shortest_edge_collapse/wmtk/components/shortest_edge_collapse/ShortestEdgeCollapse.cpp
@@ -137,6 +137,19 @@ void ShortestEdgeCollapse::write_vtu(const std::string& path)
 
 bool ShortestEdgeCollapse::collapse_edge_before(const Tuple& t)
 {
+    // TriMesh::collapse_edge_before is the classical link condition, and it is applied
+    // unconditionally -- it is not gated on preserve_topology.
+    //
+    // It does stop the simplification from changing the surface topology, so on the face of
+    // it it belongs under that flag. But it is not only a topology guard: it is the only
+    // check standing between collapse_edge_conn and a connectivity that wmtk::TriMesh
+    // cannot represent, since nothing downstream re-checks it (invariants() only tests the
+    // envelope). Relaxing it does not produce a valid mesh with different topology, it
+    // produces a corrupt one.
+    //
+    // TODO: the simplification therefore still preserves topology even when
+    // preserve_topology is off, because the toolkit does not support non-manifold meshes.
+    // Lifting this needs a mesh data structure that can represent them.
     if (!TriMesh::collapse_edge_before(t)) return false;
 
     // v1 is removed by the collapse, v2 survives (TriMesh::collapse_edge_conn keeps vid2).
--- a/components/shortest_edge_collapse/wmtk/components/shortest_edge_collapse/ShortestEdgeCollapse.cpp
+++ b/components/shortest_edge_collapse/wmtk/components/shortest_edge_collapse/ShortestEdgeCollapse.cpp
@@ -39,7 +39,8 @@ void ShortestEdgeCollapse::create_mesh(
     size_t n_vertices,
     const std::vector<std::array<size_t, 3>>& tris,
     const std::vector<size_t>& frozen_verts,
-    double eps)
+    double eps,
+    bool freeze_bnd)
 {
     wmtk::TriMesh::init(n_vertices, tris);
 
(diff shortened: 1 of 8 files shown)


components/shortest_edge_collapse/wmtk/components/shortest_edge_collapse/ShortestEdgeCollapse.cpp
```diff
@@ -137,6 +137,19 @@ void ShortestEdgeCollapse::write_vtu(const std::string& path)
...
+    // It does stop the simplification from changing the surface topology, so on the face of
+    // it it belongs under that flag. But it is not only a topology guard: it is the only
+    // check standing between collapse_edge_conn and a connectivity that wmtk::TriMesh
+    // cannot represent, since nothing downstream re-checks it (invariants() only tests the
+    // envelope). Relaxing it does not produce a valid mesh with different topology, it
+    // produces a corrupt one.
+    //
+    // TODO: the simplification therefore still preserves topology even when
+    // preserve_topology is off, because the toolkit does not support non-manifold meshes.
+    // Lifting this needs a mesh data structure that can represent them.
```
daniel-zint, line 219: This comment is weird. Our data structure does support non-manifold meshes.

components/tetwild/wmtk/components/tetwild/tetwild.cpp
```diff
@@ -172,8 +178,12 @@ TetWildMesh::ExportStruct tetwild_with_export(nlohmann::json json_params)
 
     // Informational input-topology report; gated behind DEBUG_euler because the
     // Euler-characteristic computation is expensive on meshes with many components.
+    // It is also needed by the preserve_topology check at the end, which compares it
+    // against the output -- without this the comparison is between two empty vectors and
+    // silently passes whatever happened to the topology.
+    const bool compute_euler = json_params["DEBUG_euler"] || params.preserve_topology;
```
daniel-zint, line 184: I don't see why this is needed all the time.

components/tetwild/wmtk/components/tetwild/tetwild_spec.json
```diff
@@ -162,13 +162,13 @@
     "pointer": "/preserve_topology",
     "type": "bool",
     "default": false,
-    "doc": "Preserve the topology of the input surface."
+    "doc": "Preserve the topology of the input surface. This forces remove_duplicate_eps to merge only exactly coincident vertices, keeps the link condition enabled during the input simplification, enables the substructure link condition during optimization, and makes the run fail if the output Euler characteristic differs from the input. When unset, all four are relaxed and the topology may change."
```
daniel-zint, line 172: These comments are really dubious. 
We don't need the standard link condition if we have the substructure link condition.

daniel-zint (changes requested):
There are some strange comments in this code, and I don't think checking the Euler number is necessary all the time. We need to discuss these changes.

danielepanozzo:
Superseded by #1030, which is the same change rebased onto current main.

Two parts of this branch are dropped there as superseded by later work:

- **`simplify_segments`** has since gained its own `use_link_condition` knob plus a `surviving_edges` out-parameter — a more developed form of the `preserve_topology` rename here. Main's version is kept.
- **`remove_duplicate_eps` now defaults to `0`** in both specs, so the Octocat demonstration above (input Euler characteristic 156 instead of 2) no longer reproduces out of the box: at tolerance 0 the de-dup is already a no-op. The bug is still real for anyone setting a nonzero tolerance, but it is a footgun rather than default-path corruption.

The two substantive fixes carry over unchanged: `preserve_topology` now gates the de-duplication tolerance in tetwild and triwild, and the Euler-characteristic vectors are computed whenever `preserve_topology` is set so the throw is no longer comparing two empty vectors.

PR #963, opened 2026-08-01
sec: let the simplification move the boundary, bounded by its own envelope

danielepanozzo:
Stacked on #962.

`ShortestEdgeCollapse` froze every boundary vertex unconditionally, so the outline of an open surface came through simplification exactly as it went in. That was a sensible default only because the surface envelope cannot police the boundary: it is a *containment* test, so a boundary sliding inwards along the surface never leaves it and goes unnoticed.

## What this does

Gives the boundary its own envelope, built the way tetwild builds `m_open_boundary_envelope`: collect the open-boundary edges of the input, `SampleEnvelope::init` on them at the same thickness as the surface envelope, and reject any collapse that takes a boundary edge out of it. The check mirrors `is_open_boundary_edge`.

A new **`freeze_boundary` flag, default `false`**, selects between the two behaviours. The envelope and its BVH are only built when the boundary is actually free to move.

It is always a *sampled* envelope regardless of `use_sample_envelope`, because `SampleEnvelope::is_outside` for edges calls `log_and_throw_error("Cannot use an exact envelope for edges")` — again as in tetwild.

## Measurements — circle.obj, 160 boundary edges

| freeze_boundary | simplified #v | boundary segs | max deviation from input outline |
| --- | ---: | ---: | ---: |
| true | 160 | 160 | 0 |
| **false** | **111** | **110** | **3.1e-4** (envelope 8.0e-4) |

Raising `eps_rel` from 1e-3 to 1e-2 takes it to 41 vertices, so the envelope is what binds rather than the check never firing.

Closed surfaces have no boundary edges, so no envelope is built and nothing changes — `tetwild_sphere`, `tetwild_octocat`, `tetwild_double_sphere` and `triwild_puzzle` all pass unchanged.

## Two review notes

**Boundary ends and junctions are pinned even when the boundary is free.** The envelope stops the boundary leaving the input outline but not sliding *along* it, so an open chain could otherwise retract from its ends. The rule is boundary valence != 2 — the same endpoint/junction test `simplify_segments` uses in 2D.

Being straight about its reach: this pins **nothing on manifold input**, where the boundary is a union of closed loops and every boundary vertex has valence exactly 2 (verified on `circle`, `bunny`, `fan`, `crown` — all `{2: n}`). And a closed loop cannot retract along itself; it can only shrink, which the envelope already catches. It is there for input that is not manifold along its boundary, tested on a synthetic two-strips-sharing-a-vertex case (boundary valence 4) where it pins the junction.

**`freeze_bnd` is the last parameter of `create_mesh`**, rather than sitting before `eps` as in the otherwise equivalent `IsotropicRemeshing::create_mesh`. Every existing caller passes `eps` positionally as the fourth argument, and a `bool` inserted ahead of it would silently swallow that value (`double` -> `bool`) and disable the envelope. The inconsistency with the sibling class is deliberate and commented.

### DIFF
--- a/components/shortest_edge_collapse/wmtk/components/shortest_edge_collapse/ShortestEdgeCollapse.cpp
+++ b/components/shortest_edge_collapse/wmtk/components/shortest_edge_collapse/ShortestEdgeCollapse.cpp
@@ -39,7 +39,8 @@ void ShortestEdgeCollapse::create_mesh(
     size_t n_vertices,
     const std::vector<std::array<size_t, 3>>& tris,
     const std::vector<size_t>& frozen_verts,
-    double eps)
+    double eps,
+    bool freeze_bnd)
 {
     wmtk::TriMesh::init(n_vertices, tris);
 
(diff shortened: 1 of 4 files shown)


daniel-zint (approved):
lgtm

PR #964, opened 2026-08-01
Non-manifold edges and vertices in TriMesh, and an optional link condition

danielepanozzo:
Makes non-manifold edges and vertices first-class in `TriMesh`, and makes the link
condition optional so simplification can go further on real-world input.

Draft: the `data2` pin points at a branch, not `main` there — see "Before merging".

## Why

`TriMesh` stored no edge adjacency at all. `Tuple::switch_face` rebuilt the opposite face on
every call by copying both endpoint fans, sorting them, intersecting, and then
`assert(fids.size() == 1 || fids.size() == 2)`. On an edge with three or more faces that
aborted in Debug and silently returned an arbitrary fan member in Release.

That is not a theoretical case. `tetwild_octocat` welds a clean `Octocat.obj` at
`remove_duplicate_eps` 2e-3 before simplifying, and the welded mesh has **278 non-manifold
edges** and 4 duplicate face pairs. That integration test has been running the simplifier on
a non-manifold mesh all along.

## What

**1. Two cyclic linked lists per triangle corner** (`TriangleConnectivity`):

```
m_edge_next[e]            next face around local edge e
m_vert_next_component[v]  representative of the next edge-connected component of v's fan
```

Both visit their set in increasing fid order, wrapping exactly once. A boundary edge is a
self-loop; a manifold edge is a two-cycle, which is the classical opposite face; a manifold
vertex is a self-loop on its own component.

The order is combinatorial, not geometric — `TriMesh` holds no vertex positions. That makes
each cycle a canonical function of the connectivity, which is what lets
`check_mesh_connectivity_validity()` verify them by recomputing from scratch and comparing.
`compute_edge_cycles`/`compute_vertex_cycles` are shared between the rebuild and the check,
so the check is not a second transcription of the same rules.

Operations recompute over the region they touched rather than splicing. Splicing would be a
constant factor faster on stars six faces wide and would buy a real class of aliasing bugs.

New API: `edge_valence`, `is_manifold_edge`, `is_boundary_edge` (now O(1)),
`vertex_component_count`, `is_manifold_vertex`, and `switch_component`, which jumps to a face
at the same vertex that no `switch_edge`/`switch_face` walk can reach.

**2. Operations.** Swap already refused non-manifold edges; it now says so via
`is_manifold_edge`. Split needed no connectivity change — it already looped over the whole
fan. Collapse merges the duplicate triangles it creates, retires vertices left with no face,
and refuses a collapse that would leave the surviving vertex faceless.

**3. `TriMesh::set_use_link_condition(bool)`**, default `true`. On `TriMesh` rather than on
one component so qslim and tetwild's simplification can opt in later without more plumbing.

**4. `WMTK_DEBUG_BRUTE_FORCE_OPS`**, a CMake option, **off by default**. When on, every split
and collapse rebuilds the mesh as a plain list of vertex triples, applies the operation there
the obvious way, and byte-compares. `#ifdef`-gated rather than a runtime flag, so a normal
build carries no branch, no member and no code for it. No CI job — it is O(#F) per operation
and earns its keep only when an operation is actually suspected.

**5. `use_link_condition` and `remove_duplicate_eps`** on both `shortest_edge_collapse` and
`isotropic_remeshing`. Each keeps its existing behaviour as the default (`remove_duplicate_eps`
-1 for sec, 1e-5 for remeshing, both previously unreachable), so no existing run changes.

Also re-enables `tests/2d_tests.cpp`, which was never listed in `tests/CMakeLists.txt` and so
had never been compiled. It builds unchanged and its 26 cases pass; it contained the only
existing non-manifold coverage.

## Result

Shortest-edge-collapse to 10% of input vertices, `eps_rel` 0.1, on the five new test meshes:

| mesh | #V in | target | link ON: #V / nm edges | link OFF: #V / nm edges |
| --- | ---: | ---: | ---: | ---: |
| 49883 | 20 | 2 | 12 / 4 | **8** / 8 |
| 91143 | 74 | 7 | 7 / 2 | 7 / 2 |
| 188018 | 128 | 12 | 96 / 192 | **24** / **0** |
| 229953 | 135 | 13 | 69 / 18 | **51** / 1 |
| 314748 | 293 | 29 | 65 / 60 | **28** / 32 |

Four of five simplify strictly further, two reach the target instead of one, and the
non-manifold edge count drops — 188018 goes from 224 in the input to 0 — because merging
duplicate triangles is what removes those configurations.

At the tighter `eps_rel` 0.01 the envelope binds long before the link condition does and the
two settings are within a few vertices on three of the five. The gain is real but only
visible once the envelope is not the limiting constraint.

## Verification

- **33 of 34 integration configs byte-identical** to a `main` build, run serially, each in
  its own directory. The one that differs is `tetwild_octocat`, for the reason above: it
  simplifies further (2839 vs 2912 vertices), Hausdorff 0.078 vs 0.077 against a 0.272
  envelope, final mesh 13,185 vs 13,180 vertices, `throw_on_fail` passing.
- Unit tests green in Release, in Debug with asserts live, and in Debug with
  `WMTK_DEBUG_BRUTE_FORCE_OPS=ON` — the last meaning every split and collapse the suite
  performs, including every edge of all five real meshes under both link-condition settings,
  agrees with the brute-force reference byte for byte (82 cases, 39,162 assertions).
- Multi-threaded on Octocat at 4 and 10 threads, with the cycles re-validated from scratch
  after each run.

## Performance

Measured by alternating the two binaries on an otherwise idle machine, since earlier numbers
I quoted turned out to be contention artefacts rather than real differences.

**tetwild input simplification (Octocat), no measurable difference:**

| | run 1 | run 2 | run 3 |
| --- | ---: | ---: | ---: |
| `main` | 20.89s | 19.86s | 19.75s |
| this PR | 19.91s | 19.86s | 19.87s |

The spread within `main` alone is larger than any gap between them.

**A bare collapse loop with no envelope (Octocat, `target_rel` 0.05), 1.3x:**

| | | | | | |
| --- | ---: | ---: | ---: | ---: | ---: |
| `main` | 0.151s | 0.169s | 0.161s | 0.151s | 0.155s |
| this PR | 0.202s | 0.206s | 0.202s | 0.205s | 0.200s |

Output identical (947 vertices, 1890 faces) in every run. That gap is the cycle maintenance
in the collapse itself. It is visible only when nothing else is in the loop; as soon as a
geometric predicate is involved -- which is every real use, including tetwild above -- it
disappears.

## Before merging

The `data2` pin (`cmake/wmtk_data.cmake`) points at branch `nonmanifold-models`
(`03c28509`), which is the previously pinned commit plus `models/nonmanifold/`. If pins are
meant to track `data2`'s `main`, that branch needs merging there first and the tag updating.
This is the only reason the PR is a draft.


daniel-zint:
#966 is the better implementation. Storing vertex components explicitly offers no advantage, since they are rarely required.

PR #965, opened 2026-08-02
Share simwild's vertex smoothing with tetwild and triwild

danielepanozzo:
Gives tetwild simwild's vertex smoothing, extracted so all four applications run one of two
shared drivers. Rebased onto `main`, which now carries #966 (the lean rewrite that replaced
#964), #967, #968 and #969.

## Why

tetwild froze on `crown.obj`: max energy stuck at ~10¹² against a target of 100, with **94%
of smoothing attempts rejected** (117 accepted, 1,694 refused per pass) and 85 surface
vertices pinned at >90% of the envelope thickness, unchanged to five decimal places across
ten consecutive passes.

Comparing the two smoothers explains it. tetwild optimized AMIPS alone, then **teleported**
the vertex onto the envelope with `nearest_point`, then accepted or rejected the whole thing.
A vertex at 99% of eps is moved ~0.18 in one jump; through a bad neighbourhood that almost
always worsens an incident tet, so the operation was vetoed, the rollback discarded the
projection, and the vertex could never come back. tetwild also applied the quality veto to
surface vertices, which simwild does not.

simwild puts the envelope *in the objective* as a squared-distance penalty, so the pull is a
gradient rather than a teleport and the line search refuses steps that leave the envelope.

## Result on the crown

Same configuration that never converged:

| | old | new |
| --- | ---: | ---: |
| max energy, iterations 0→2 | 7.78e11 → 7.78e11 → **1.13e12 rising** | 7.08e7 → 4232.91 → **86.73** ✓ |
| smoothing | accepted 117, rejected 1694 | accepted 56, rejected 30 |
| rejections by cause | — | envelope 26, quality 4, **inverted 0** |
| Hausdorff vs eps 0.364 | — | **0.212** ✓ |

Zero inversion rejections is the mechanism. It converges *inside* the envelope — disabling
the envelope also "converged" but left the mesh 12.11 away, i.e. by abandoning the constraint
rather than satisfying it.

## What's shared

`src/wmtk/optimization/SmoothVertex.hpp` holds two functions:

- `smooth_vertex_3d` — simwild 3D, tetwild
- `smooth_vertex_2d` — triwild, simwild 2D

Separate rather than dimension-templated: element arity, the containment test (envelope edges
vs triangles) and the position representation all differ.

Each mesh supplies two envelope accessors. **They are two and not one on purpose** — simwild
pulls toward `m_envelope_orig` (or `m_order_2_edge_envelope` for order-2 vertices) while
checking containment against `m_envelope`, and collapsing them changed one config out of
fourteen. The byte comparison is what caught it.

Also adds `SmoothRejectCounters`, logged once per pass. A pass that refuses most of what it
tries is indistinguishable from a converging one without it, and it is the measurement that
would have identified this problem immediately.

## Verification

- **simwild: all 14 configs byte-identical** after the 3D extraction. It is the reference, so
  any difference means the extraction is wrong rather than that something improved.
- **triwild: both configs byte-identical** after the 2D extraction, same role, both with
  `throw_on_fail`.
- **simwild 2D moves and improves** — it gains the float-inversion pre-check and the two-stage
  solve. Of its eight 2D configs, the two that run smoothing passes moved and the six that
  run zero passes are byte-identical:

  | | max energy | #V | #F |
  | --- | --- | --- | --- |
  | energies_2d | 4.907 → 4.782 | 1383 → 1217 | 2686 → 2351 |
  | remeshing_2d | 4.607 → 4.063 | 204 → 198 | 347 → 335 |

- **All four tetwild configs pass**, all with `throw_on_fail`, all under target energy, all
  rounded, all inside the envelope. Re-measured on the rebase (`num_smoothing_passes` 10):

  | config | max energy / stop | containment d(output→input) / eps |
  | --- | ---: | --- |
  | crown | 97.79 / 100 | 0.1182 / 0.3644 |
  | double_sphere | 8.54 / 10 | 0.001341 / 0.005195 |
  | octocat | 9.91 / 10 | 0.08487 / 0.272 |
  | sphere | 5.95 / 10 | 0.001167 / 0.00485 |

  The containment column is the invariant the envelope actually promises, per #967. crown
  sits higher on energy than the pre-rebase 88.72 because the passes went from 50 to 10;
  it still lands under the 100 target.

## Read this before merging — three default changes

The smoothing rewrite is the point of the PR, but the branch also carries three tetwild
defaults that were set while debugging and are **not obviously right for everyone**:

1. **`num_smoothing_passes` defaults to 10** (was effectively 1, hard-coded as `ops[3]`).
   That is 10x the smoothing work per optimization iteration for every tetwild user. The
   integration configs all still pass, but the default deserves a decision rather than
   inheritance.
2. **`simplify_envelope_ratio` defaults to 0.5.** The simplification and the tetrahedralisation
   previously shared one `SampleEnvelope` object at the same eps, leaving the optimizer no
   headroom. They are now two envelopes with the simplification at half. Reasonable, but it
   changes tetwild's simplification for everyone; 1.0 restores the old shared behaviour.
3. **`w_amips` defaults to 1e-4**, hence `w_envelope` ≈ 0.9999. This is simwild's convention
   and is what makes the new smoothing work, but it means smoothing is now primarily about
   staying on the surface with quality as a secondary preference.

`DEBUG_disable_envelope` is also included; it defaults to false and is diagnostic only.

## Not fixed by this

The crown's simplified input has 75 non-manifold edges, 10 components where the original had
7, and Euler characteristic +4 against −9,634. That comes from simplifying with the link
condition off and no smoothing change repairs it. This removes one specific deadlock; whether
crown then produces a *good* mesh is a separate question.

## Rebase notes

Four conflicts, all where `main` moved under the branch:

- **`tetwild.cpp` / `tetwild_spec.json`** — `main` gained `simplify_use_link_condition` and
  `simplify_use_sample_envelope` (#966) next to this branch's `simplify_envelope_ratio`.
  All three kept; they are independent knobs.
- **`TetWildMesh.cpp`** — `main` deleted `check_attributes`, this branch had edited it for the
  `shared_ptr` envelope. Took the deletion.
- **`TetWildMesh.h`** — kept `main`'s `m_input_names` alongside this branch's per-thread solver.
- **`test_surface_swap.cpp`** — kept #969's corrected comment (the "stale flag" state does not
  occur), with only the `shared_ptr` conversion on top.

`src/wmtk/envelope/Envelope.cpp` merged cleanly, so the branch's `disabled` flag now sits on
top of #968's sampling shrink. The data2 pin is `main`'s, unchanged.

One commit on top of the rebase fixes two comments that the rebase made false: the
`SampleEnvelope::disabled` doc still described the `nearest_point` snap this branch removes,
and `tetwild.cpp` still claimed the simplification envelope is handed to `TetWildMesh` (it
gets its own now).

## Verification after the rebase

- `ctest` **91/91 green**, which now includes the component tests (#969 registered them).
- All 33 integration configs pass, 133 s. Six of them run with `throw_on_fail`.
- `clang-format` 21.1.8 clean on every changed file.

### DIFF
--- a/components/tetwild/wmtk/components/tetwild/Parameters.h
+++ b/components/tetwild/wmtk/components/tetwild/Parameters.h
@@ -98,7 +98,7 @@ struct Parameters
      * changing connectivity, so on meshes where split/collapse/swap have run out of useful
      * moves it is the only thing left that can lower the energy.
      */
-    int num_smoothing_passes = 50;
+    int num_smoothing_passes = 10;
 
     bool debug_output = false;
     bool perform_sanity_checks = false;
--- a/components/tetwild/wmtk/components/tetwild/Parameters.h
+++ b/components/tetwild/wmtk/components/tetwild/Parameters.h
@@ -82,6 +82,15 @@ struct Parameters
     double stop_energy = 10;
 
     /**
+     * Relative weight of the AMIPS (quality) term against the envelope (stay-on-surface)
+     * term during smoothing. w_envelope is derived as 1 - w_amips in the TetWildMesh
+     * constructor, so the small default means the envelope dominates and AMIPS acts as a
+     * light quality preference. Matches simwild.
+     */
+    double w_amips = 1e-4;
+    double w_envelope = 1. - 1e-4; // derived; not read from json
+
+    /**
      * How many smoothing passes each optimization iteration runs.
      *
      * This is ops[3] in local_operations({{split, collapse, swap, smooth}}), which was
--- a/components/tetwild/wmtk/components/tetwild/Parameters.h
+++ b/components/tetwild/wmtk/components/tetwild/Parameters.h
@@ -81,6 +81,16 @@ struct Parameters
 
     double stop_energy = 10;
 
+    /**
+     * How many smoothing passes each optimization iteration runs.
+     *
+     * This is ops[3] in local_operations({{split, collapse, swap, smooth}}), which was
+     * hard-coded to 1. Smoothing is the only phase that improves element quality without
+     * changing connectivity, so on meshes where split/collapse/swap have run out of useful
+     * moves it is the only thing left that can lower the energy.
+     */
+    int num_smoothing_passes = 50;
+
     bool debug_output = false;
     bool perform_sanity_checks = false;
 
--- a/src/wmtk/envelope/Envelope.hpp
+++ b/src/wmtk/envelope/Envelope.hpp
@@ -64,9 +64,10 @@ public:
      * optimization is blocked by the envelope or by the mesh itself. It removes the
      * containment guarantee entirely -- the output is not usable, only diagnostic.
      *
-     * Note nearest_point() is unaffected: smoothing still projects surface vertices onto
-     * the envelope surface, because that is a constraint on where a vertex goes rather
-     * than a veto on the operation.
+     * Note nearest_point() is unaffected, so EnvelopeEnergy still pulls surface vertices
+     * toward the input while smoothing. Only the veto goes away, which is the point: it
+     * separates "the optimizer cannot move because the envelope forbids it" from "the
+     * optimizer cannot move at all".
      */
     bool disabled = false;
     void init(
--- a/src/wmtk/envelope/Envelope.hpp
+++ b/src/wmtk/envelope/Envelope.hpp
@@ -55,6 +55,20 @@ public:
     double eps2 = 1e-6;
     double sampling_dist = 1e-3;
     bool use_exact = false;
+
+    /**
+     * Debug escape hatch: when set, every is_outside() answers "inside".
+     *
+     * The envelope is the only thing that can reject an operation for geometric rather
+     * than combinatorial reasons, so turning it off tells you whether a stalled
+     * optimization is blocked by the envelope or by the mesh itself. It removes the
+     * containment guarantee entirely -- the output is not usable, only diagnostic.
+     *
+     * Note nearest_point() is unaffected: smoothing still projects surface vertices onto
+     * the envelope surface, because that is a constraint on where a vertex goes rather
+     * than a veto on the operation.
+     */
+    bool disabled = false;
     void init(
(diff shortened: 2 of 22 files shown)


src/wmtk/envelope/Envelope.hpp
```diff
@@ -55,6 +55,20 @@ class SampleEnvelope : public wmtk::Envelope
...
+     * The envelope is the only thing that can reject an operation for geometric rather
+     * than combinatorial reasons, so turning it off tells you whether a stalled
+     * optimization is blocked by the envelope or by the mesh itself. It removes the
+     * containment guarantee entirely -- the output is not usable, only diagnostic.
+     *
+     * Note nearest_point() is unaffected: smoothing still projects surface vertices onto
+     * the envelope surface, because that is a constraint on where a vertex goes rather
+     * than a veto on the operation.
+     */
+    bool disabled = false;
```
daniel-zint, line 72: Sounds like this should be removed.

components/tetwild/wmtk/components/tetwild/Parameters.h
```diff
@@ -81,6 +81,25 @@ struct Parameters
...
+
+    /**
+     * How many smoothing passes each optimization iteration runs.
+     *
+     * This is ops[3] in local_operations({{split, collapse, swap, smooth}}), which was
+     * hard-coded to 1. Smoothing is the only phase that improves element quality without
+     * changing connectivity, so on meshes where split/collapse/swap have run out of useful
+     * moves it is the only thing left that can lower the energy.
+     */
+    int num_smoothing_passes = 10;
```
daniel-zint, line 101: 10 iterations sound like a lot. Is it really necessary/helpful to do that many?

daniel-zint (approved):
seems fine in general

components/tetwild/wmtk/components/tetwild/Parameters.h
danielepanozzo: It makes a big difference to lower the energy below 1000, let's leave it for now, we can try lowering it later if it is a performance bottleneck

PR #966, opened 2026-08-02
Non-manifold edges and vertices in TriMesh, without the extra connectivity

danielepanozzo:
Same functionality as #964 — non-manifold edges and vertices in `TriMesh`, an optional link
condition, the brute-force cross-check, the same tests — but without adding any connectivity
data. It is an alternative to #964, not a follow-up: only one of the two should land.

## The idea

#964 stored two cyclic linked lists per triangle:

```
m_edge_next[e]            next face around local edge e, in increasing fid order
m_vert_next_component[v]  representative of the next edge-connected component of v's fan
```

Both are a cache. `VertexConnectivity::m_conn_tris` is already kept sorted — the codebase
asserts it in `set_intersection` and relies on it in `split_edge` — so the faces common to an
edge's two endpoint fans **are** that edge's fan in increasing fid order, which is exactly what
`m_edge_next` encoded. Likewise the fan's edge-connected components, and so
`m_vert_next_component`.

So both arrays come out, together with `compute_edge_cycles`, `compute_vertex_cycles`,
`vertex_cycle_successors`, `rebuild_edge_cycle`, `rebuild_vertex_cycle`,
`rebuild_cycles_around`, `rebuild_cycles_for`, `rebuild_all_cycles`, and the twelve
maintenance call sites in `split_edge`, `collapse_edge`, `swap_edge`, `split_face`,
`consolidate_mesh` and `init`.

The public API is unchanged and answers the same questions directly:

| | how |
| --- | --- |
| `switch_face` | one lockstep pass over the two endpoint fans, tracking the next fid above this one and the smallest one for the wrap |
| `is_boundary_edge` | the same pass, stopping at the second face |
| `edge_valence` | the same pass, counting |
| `is_manifold_edge` | the same pass, stopping at the third face |
| `vertex_component_count`, `switch_component` | union-find over the fan, then the next representative |

The scan allocates nothing and keeps no scratch, so it needs no `thread_local` and is safe on
as many threads as you like — #964 needed `thread_local` buffers inside `const` methods.

`switch_faces` reverts to what `main` had. It already scanned the smaller sorted fan, which
yields increasing fid order, so #964's rewrite of it was buying nothing — and, as the numbers
below show, was costing something.

## Byte-identity

By construction rather than by luck: both versions encode the same total order (by fid) on the
same set (the fan), so `switch_face` steps to the same face and `switch_component` picks the
same representative.

Verified against a **#964 build**, not against `main`:

- **32 of 34 integration configs byte-identical**, run serially, each in its own directory.
- The other two are `tetwild_crown` and `tetwild_octocat`, the only two configs with
  `num_threads: 10`. Those are nondeterministic: **#964 does not reproduce itself** on them —
  three runs of the same binary gave three different hashes. Pinned to one thread, both become
  **byte-identical** too.

So: identical everywhere a byte comparison is meaningful.

## Verification

- Unit tests green in Release and in Debug with asserts live: **48592 assertions in 81 cases**,
  the same counts as a #964 build.
- Green with `WMTK_DEBUG_BRUTE_FORCE_OPS=ON`: **48812 assertions in 82 cases**. Every split,
  collapse, swap and split_face the suite performs — including every edge of all five real
  meshes under both link-condition settings — agrees with the brute-force reference byte for
  byte.
- Multi-threaded Octocat at 4 and 10 threads, in Release and in Debug with asserts (so
  `assert(check_mesh_connectivity_validity())` inside `consolidate_mesh` is live). Clean, with
  `throw_on_fail` passing and 0 non-manifold edges in the output.
- `test_nonmanifold.cpp`'s `brute_force_edge_valence` no longer calls
  `get_incident_fids_for_edge`: that helper intersects the two endpoint fans, which is now how
  `edge_valence` itself answers, so it would have compared the implementation against itself.
  It scans every live face instead. That is the only test change.

## Size

| | #964 | this PR |
| --- | ---: | ---: |
| `TriMesh.cpp`, real code | +583 | **+351** |
| `TriMesh.cpp`, inside `#ifdef WMTK_DEBUG_BRUTE_FORCE_OPS` | +340 | +340 (unchanged) |
| `TriMesh.h` | +216 / -6 | **+149 / -6** |

Removing the caching cuts the code that actually compiles by **40%**. Everything else in the
PR — the components' options, the tests, the CMake option, the data pin — is identical to #964.

## Performance

Measured by alternating builds on an otherwise idle machine, first round discarded as warm-up.
All three builds produce identical output on every benchmark below, so they are doing the same
work.

**Pure navigation** (Octocat, 2.27M calls each, mean of 3) — the case that most favours storing
the cycles:

| | `main` | #964 | this PR |
| --- | ---: | ---: | ---: |
| `switch_face` | 388 ms | **11.8 ms** | 40.9 ms |
| `is_boundary_edge` | 105 ms | **2.3 ms** | 29.9 ms |
| `switch_faces` | 112 ms | 151 ms | **115 ms** |

#964 wins the first two by a wide margin, as it should — O(1) against O(d₀+d₁). Both are far
ahead of `main`, which copied and sorted both fans on every `switch_face`. The third row is the
one to notice: **#964's `switch_faces` rewrite is ~33% slower than the code it replaced**, and
reverting it recovers that.

**Real workloads** (single-threaded, so the comparison is deterministic):

| | `main` | #964 | this PR |
| --- | ---: | ---: | ---: |
| bare collapse loop (Octocat, `target_rel` 0.05, no envelope) | 0.257 s | 0.323 s | **0.263 s** |
| isotropic remeshing (bunny) | 3.97 s | 3.57 s | **3.52 s** |
| tetwild simplify phase (Octocat) | 19.75 s | 19.91 s | **19.58 s** |

The collapse loop reproduces #964's own reported result — it is **1.26× slower than `main`**
there, which is the cycle maintenance. This PR has no maintenance to pay for and lands back on
`main`. On isotropic remeshing both are ~11% faster than `main`, this PR marginally ahead. The
tetwild simplify phase — the only phase that touches `TriMesh` at all — is the same in all
three.

One number worth heading off, since it looks alarming in isolation: the *whole* tetwild run on
Octocat is 84.6 s on `main` against 99.8 s (#964) and 98.2 s (this PR). None of that is the
data structure. Both #964 and this PR simplify the input further — 2839 vertices / 5706 faces
against `main`'s 2912 / 6025, because merging duplicate triangles unlocks collapses the link
condition used to forbid — so the arrangement and optimization downstream get a different
surface to work on. That is #964's intended behaviour change, inherited here unchanged, and it
shows up entirely *after* the simplify phase.

So the summary is that the linked lists were a large win on isolated navigation and a net loss
on anything that mutates the mesh, which is what the operations actually do. Removing them
costs nothing measurable on real work and returns the `switch_faces` regression.

## Before merging

Same caveat as #964: the `data2` pin (`cmake/wmtk_data.cmake`) points at branch
`nonmanifold-models` (`03c28509`), which is the previously pinned commit plus
`models/nonmanifold/`. That branch needs merging into `data2`'s `main` and the tag updating.


src/wmtk/TriMesh.cpp
```diff
@@ -635,13 +1100,38 @@ bool TriMesh::collapse_edge(const Tuple& loc0, std::vector<Tuple>& new_tris)
...
+    // at. The link condition used to rule this out on the way in; refuse it explicitly so
+    // it stays ruled out when the link condition is off.
+    {
+        const size_t vid1 = loc0.vid(*this);
+        const size_t vid2 = loc0.switch_vertex(*this).vid(*this);
+        const auto& n1 = m_vertex_connectivity[vid1].m_conn_tris;
+        const auto& n2 = m_vertex_connectivity[vid2].m_conn_tris;
+        const size_t shared = set_intersection(n1, n2).size();
+        if (shared == n1.size() && shared == n2.size()) {
+            return false;
```
daniel-zint, line 1140: That essentially means a single triangle can't collapse to nothing, but I think we would like to allow that?

src/wmtk/TriMesh.cpp
```diff
@@ -1707,13 +2338,16 @@ size_t TriMesh::get_next_empty_slot_v()
 
 bool TriMesh::swap_edge_before(const Tuple& t)
 {
-    const auto opps = t.switch_faces(*this);
-    if (opps.size() != 1) {
+    // A swap replaces the edge by the one joining the two opposite vertices, which needs
+    // exactly two of them: a boundary edge has none and a non-manifold edge has no
+    // canonical pair. Rejecting is the only sensible answer for both, and it is what keeps
+    // swap from having to reason about fans at all.
+    if (!is_manifold_edge(t)) {
```
daniel-zint: I'll revert that change. `is_manifold_edge` is also `true` if the edge is on the boundary. A swap is only allowed if it is an interior manifold edge, so the old check was better.

daniel-zint (commented):
Some remarks. In general, I think this PR is fine. I **did not** carefully check the vertex components code.

PR #975, opened 2026-08-05
tetwild: Thingi10K sweep defaults, plus the two fixes the sweep needed

danielepanozzo:
Two things, both driven by running tetwild over Thingi10K: the retuned defaults the sweep runs with, and the two defects the sweep exposed. The fixes are here rather than in separate PRs because the sweep cannot complete without them.

## 1. Retuned defaults

Biased toward getting through a large heterogeneous input set rather than toward the best mesh for any one model.

| parameter | old | new | why |
| --- | --- | --- | --- |
| `eps_rel` | `2e-3` | `1e-3` | the value the sweep runs at; no reason for every config to restate it |
| `remove_duplicate_eps` | `2e-3` | `0` | merge only exactly coincident input vertices. A tolerance relative to the bbox diagonal silently welds real features on models whose detail sits below it. `0` still runs the dedup pass (needed for valid connectivity); negative would skip it |
| `stop_energy` | `10` | `100` | the tail of the optimization is where a sweep spends its time, and the interesting question is whether a model completes at all |
| `simplify_use_sample_envelope` | `false` | `true` | the exact envelope dominates the cost of simplifying a large input. The doc's caveat stands — this is a throughput tradeoff, not a claim it is safe everywhere |
| `simplify_envelope_ratio` | `0.5` | `0.5` (unchanged) | **was raised to 0.9, then reverted** — 0.9 left the optimizer no envelope headroom and two sweep models stopped converging. See below. |
| `simplify_use_link_condition` | `true` | `false` | let the input simplify past topology-preserving collapses. The surface handed to the arrangement may then differ topologically from the input |

`Parameters.h` keeps `stop_energy = 10` as the C++ struct default — that is what the unit tests construct against, and `tetwild.cpp` overwrites it from the JSON on every run through the component entry point.

### Why `simplify_envelope_ratio` went back to 0.5

Raising it to 0.9 is what the parameter's own doc warns against — "if the simplification is allowed to place vertices at the full limit, the optimizer starts with a mesh where almost every move already lies outside and most operations are vetoed". Two sweep models stalled on exactly that: 240280 plateaued at max energy 4891 with smoothing at **17 accepted / 825 envelope-rejected**, and 562343 ran 60 iterations without reaching the target.

Running 240280 with `DEBUG_disable_envelope` confirmed the envelope was the binding constraint — energy **1.69e12 → 10.05 in two iterations**, against a plateau at ~4900 with it on. At ratio 0.5 both converge: 240280 in 18 iterations (rejections 825 → 60), 562343 in 13.

## 2. Fix: the high-valence split gate never applied to new vertices

`split_all_edges` sized its claim array with `vert_capacity()`, which returns `current_vert_size` — the **live vertex count**, not the reservation. Every vertex created during the pass therefore lands at an id at or above that size, trips

```cpp
if (vid >= m_high_valence_claim_size) continue;
```

and is silently exempt from the gate. The comment there asserts the opposite invariant.

Those exempt vertices are the ones that run away: on 509315, `v15601` — created mid-pass — absorbed **1792 valence-increasing splits in a single pass**, where the gate permits one, reaching valence ~700.

Sized to the preallocated capacity instead. Note this does not by itself stop the divergence (see below); it is a real defect regardless, since the gate is inoperative for exactly the vertices it most needs to cover.

## 3. Fix: a split that would invert now always takes the exact-rational midpoint

**This is the one that fixes the sweep timeouts.**

When a split's rounded midpoint inverts an incident tet, the split fell back to the exact rational midpoint — which lies on the shared edge and so cannot invert a previously-valid tet — but **only when an endpoint was already rational**. Between two rounded endpoints the split was rejected outright.

That restriction refuses the split *exactly where the mesh is degenerate*, which is where refinement is needed to break the configuration apart. The region cannot be repaired, stuck-refine then hammers it from outside, and the max energy climbs by orders of magnitude per split pass.

On Thingi10K 509315, 8 runs of the sweep configuration:

| | converged | diverged |
| --- | --- | --- |
| before | 3 | 5, to 1e16–1e20, hitting the 1 h timeout |
| after | **8** | **0** — 2–5 iterations, 85–98 s |

Healthy models are unchanged or better: 100026 4 s → 3 s (max energy 51.2 → 18.2), 213888 38 s → 32 s (91.573 both), 39925 and 229958 unchanged.

### Why the rounding guard is not needed

It existed to stop a fully-rounded mesh reacquiring exact coordinates, with model 73435 as the counter-example. The iteration already prevents that, and more reliably:

- a split is the **only** operation that can un-round a vertex — collapse, all four swaps and smoothing never do (`m_is_rounded = false` appears once in `EdgeSplitting.cpp`);
- the post-optimization pass is `local_operations({{0,1,0,0}})`, collapse-only, so no rational vertex can appear after the loop;
- the loop does not stop until every vertex is rounded **and** the energy target is met.

73435 now finishes `all_rounded` in 3 of 3 runs (max energy 11.7 / 13.7 / 15.2, no "Not all vertices rounded").

`stuck_refine_rational_split` is removed with it, from tetwild, triwild and simwild. In triwild and simwild it only ever gated this same fallback and defaulted to `true`, so their behaviour is unchanged — it just removes the off switch.

**Breaking:** a config still setting `stuck_refine_rational_split` will now fail JSE validation.

## Verification

`ctest` **97/97 including `Integration_Tests`**, Release, macOS/arm64 — which covers triwild and simwild as well.

The sweep itself is at ~9,300 of 10,000 with 8 failures on the pre-fix binary; the 4 remaining non-input failures are the divergence class this addresses.

### DIFF
--- a/components/simwild/wmtk/components/simwild/EdgeSplitting.cpp
+++ b/components/simwild/wmtk/components/simwild/EdgeSplitting.cpp
@@ -75,7 +75,7 @@ void SimWildMesh::split_all_edges()
             "[force-split] {} worst-tet longest edges force-split",
             m_force_split_count);
     }
-    // Consumed: the queued force-split edges no longer exist after this pass.
+    // Consumed: the queued force-split edges no longer exist after this pass
     m_force_split_edges.clear();
 }
 
--- a/components/simwild/wmtk/components/simwild/EdgeSplitting.cpp
+++ b/components/simwild/wmtk/components/simwild/EdgeSplitting.cpp
@@ -172,10 +172,6 @@ bool SimWildMesh::split_edge_after(const Tuple& loc)
         std::atomic_ref<size_t>(m_force_split_count).fetch_add(1, std::memory_order_relaxed);
     }
     if (!m_vertex_attribute[v_id].m_is_rounded) {
-        if (!m_params.stuck_refine_rational_split) {
-            return false;
-        }
-
         m_vertex_attribute[v_id].m_pos =
             (m_vertex_attribute[v1_id].m_pos + m_vertex_attribute[v2_id].m_pos) / 2;
         p = to_double(m_vertex_attribute[v_id].m_pos);
--- a/components/simwild/wmtk/components/simwild/Parameters.h
+++ b/components/simwild/wmtk/components/simwild/Parameters.h
@@ -58,7 +58,6 @@ struct Parameters
     // the current worst-tet set (the seeds used by refine_sizing_around_worst)
     // instead fall back to the EXACT rational midpoint (never inverts) and keep
     // the new vertex un-rounded, so the worst region can still be refined.
-    bool stuck_refine_rational_split = true;
 
     // ---- Skip good regions ----------------------------------------------
     // Only smooth vertices incident to a tet whose energy is >=
@@ -144,7 +143,6 @@ struct Parameters
         stuck_refine_force_split = json_params["stuck_refine_force_split"];
         stuck_refine_min_scalar = json_params["stuck_refine_min_scalar"];
         stuck_refine_gradation = json_params["stuck_refine_gradation"];
-        stuck_refine_rational_split = json_params["stuck_refine_rational_split"];
 
         // Skip good regions.
         skip_good_regions = json_params["skip_good_regions"];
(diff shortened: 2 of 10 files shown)


daniel-zint (approved):
I dislike that the C++ struct parameters don't match the JSON spec. `stop_energy` should be the same in both. Otherwise, these changes look good to me.

PR #981, opened 2026-08-06
tetwild: optional interleaved smoothing between the topology passes

danielepanozzo:
**Draft — parked for later.** The idea works on one model and not on another; the measurements are below so this can be picked up without redoing them.

## What it does

An iteration currently runs split, collapse and swaps back to back, then smooths `num_smoothing_passes` times. With `interleaved_smoothing` set it instead runs

```
split    + interleaved_smoothing_passes smoothing passes
collapse + ...
swaps    + ...
```

so each topology pass hands the next one a relaxed mesh rather than its raw output. Smoothing is the only phase that improves quality without changing connectivity, so the hypothesis is that it keeps the optimizer off the plateaus where split, collapse and swap simply undo each other — a pattern visible in the stalled sweep models, where `after swap == start` exactly, iteration after iteration.

**Off by default.** The scheduler already runs op types in order, so this is three `local_operations` calls rather than a change to the pass machinery.

## Measurements

| model | baseline | interleaved | |
| --- | --- | --- | --- |
| 240280 (stalled) | 18 iterations, 219 s | **13 iterations, 178 s** | clear gain |
| 562343 (stalled) | 13 iterations | **15 iterations**, mesh identical | mild loss |
| octocat | max E 87.4 | **81.3** | gain |
| crown | max E 89.3, avg 7.061 | max E 97.3, avg **6.841** | mixed |
| sphere | max E 6.7505 | 6.7505 | identical |
| double_sphere | max E 7.5491 | 7.5491 | identical |

One clear win, one mild loss, one gain, one mixed, two neutral.

## Why it is not on by default

- **The 240280 result does not generalise.** That single model was the case for the idea; 562343, the other stalled model tested, got slightly worse.
- **sphere's Hausdorff distance moves 4.2e-07 → 6.5e-04.** Small in absolute terms — both are well inside an `eps_rel` of 1e-3 — but three orders of magnitude, and the only number that moved that far. Worth understanding before this becomes a default.
- The two stalled models it was aimed at turned out to be stalled for a different reason: `simplify_envelope_ratio` leaving the optimizer no envelope headroom. Both converge once that is addressed, with or without this change.

## If picked up again

The useful experiment is a larger sample of *stalled* models rather than more single cases, plus an explanation for the sphere Hausdorff shift. `interleaved_smoothing_passes` (default 2) is the other knob worth sweeping.

97/97 including `Integration_Tests`, Release, macOS/arm64.

### DIFF
--- a/components/tetwild/wmtk/components/tetwild/Parameters.h
+++ b/components/tetwild/wmtk/components/tetwild/Parameters.h
@@ -117,6 +117,18 @@ struct Parameters
      */
     int num_smoothing_passes = 10;
 
+    // Interleave smoothing between the topology passes instead of running it all at the end
+    // of the iteration. With this on, one iteration is
+    //     split   + interleaved_smoothing_passes smoothing passes
+    //     collapse + ...
+    //     swaps    + ...
+    // rather than split, collapse, swaps, then num_smoothing_passes passes. Smoothing is the
+    // only phase that improves quality without changing connectivity, so giving each topology
+    // pass a chance to be relaxed before the next one runs may keep the optimizer off the
+    // plateaus where split, collapse and swap simply undo each other.
+    bool interleaved_smoothing = false;
+    int interleaved_smoothing_passes = 2;
+
     bool debug_output = false;
     bool perform_sanity_checks = false;
 
--- a/components/tetwild/wmtk/components/tetwild/TetWildMesh.cpp
+++ b/components/tetwild/wmtk/components/tetwild/TetWildMesh.cpp
@@ -76,8 +76,18 @@ void TetWildMesh::mesh_improvement(int max_its)
     for (int it = 0; it < max_its; it++) {
         ///ops
         logger().info("\n========it {}========", it);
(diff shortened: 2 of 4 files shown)


danielepanozzo:
Closing: this change is now carried by #982.

The commit was cherry-picked into `triwild-sweep` as part of bringing triwild to feature parity with tetwild — `interleaved_smoothing` and `interleaved_smoothing_passes` needed to exist in both applications, and having the tetwild half sitting in a separate branch would have let the two drift. #982 has the tetwild commit verbatim plus the matching triwild implementation.

Nothing is lost and nothing is decided by this: the option is still **off by default** in both, for the reasons measured here — the 240280 win did not generalise, and sphere's Hausdorff moved 4.2e-07 to 6.5e-04. The measurement table in this PR body is the record of that and is worth keeping; the branch `interleaved-smoothing` is left in place rather than deleted.

If #982 is ever reworked to drop the tetwild half, this should be reopened.

PR #994, opened 2026-08-09
tetwild sweep: TETWILD_SANITY_CHECKS to turn on the exact orientation checks

danielepanozzo:
The checks that name an inverted tet directly — `After embed_tri_in_poly_mesh: Tet [...] is inverted!` — sit behind `DEBUG_sanity_checks` and are **off by default**. So a run that produces inverted tets says nothing about them. The failure only surfaces later, if two of them happen to share a face, as

```
Face [3863, 3880, 1267799] appears more than once in the tet list
```

from a downstream combinatorial check that points at the wrong cause — it reads as duplicated geometry when the actual condition is a wrong winding.

Chasing that on Thingi10K 100727 cost most of a day, and the answer turned out to be six tets with negative exact signed volume coming straight out of the arrangement. Filed upstream as [VolumeRemesher#23](https://github.com/wildmeshing/VolumeRemesher/pull/23).

Setting `TETWILD_SANITY_CHECKS=1` puts `DEBUG_sanity_checks` into the per-model config, so a sweep can hunt for the real condition instead of its symptom. Off by default: the per-operation checks run for the whole optimization, not just insertion, and are expensive.

An env var rather than hand-editing the deployed script, because doing it by hand is how `SAMPLE=name` got silently reverted earlier today — a stale working copy was written over the committed one, and the sweep would have run in the wrong order without anyone noticing.

Scripts only; no library or component code touched.

### DIFF
--- a/components/tetwild/wmtk/components/tetwild/scripts/README.md
+++ b/components/tetwild/wmtk/components/tetwild/scripts/README.md
@@ -67,6 +67,7 @@ TETWILD_JOB_TIMEOUT=21600 python3 run_tetwild_sweep.py
 | `TETWILD_SEED` | 0 | for `TETWILD_SAMPLE=random` |
 | `TETWILD_REPORT_ONLY` | unset | regenerate the report and exit |
 | `TETWILD_REPORT_BANNER` | unset | note shown at the top of the HTML report |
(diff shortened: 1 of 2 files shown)


danielepanozzo:
Closing at the author's request. The env var is easy to reintroduce if the exact orientation checks are needed for another sweep.

PR #1026, opened 2026-08-12
Pull surface vertices with a fixed-target spring, in 2D and 3D, on by default

danielepanozzo:
Adds a fixed-target spring as an alternative to the envelope pull during smoothing, in 2D and 3D, plus a switch for the smoothing quality gate.

> **⚠️ This PR currently defaults `spring_pull` to `true`, and in that configuration it FAILS 2 of the 53 integration configs.** The full measurements are below and they do not support that default. Recommendation is to ship it `false` (opt-in); it is left `true` pending the author's decision.

## What it does

`EnvelopeEnergy` penalises the distance to the input **as a set** — it recomputes the nearest point at every evaluation, so its gradient `2w(x − n(x))` is always normal to the input and a vertex slides along it at no cost. That is deliberate, and it is what the rank-one Gauss-Newton Hessian encodes.

`SpringEnergy2D/3D` captures the nearest point once, before the solve, and springs the vertex to that single point. The energy is then an exact quadratic — value, gradient (`2w(x − t)`) and Hessian (`2w·I`) all exact, no Gauss-Newton approximation — and tangential motion is resisted as strongly as normal motion.

## The one thing the spring is better at

The envelope pull leaves a floor on surface deviation that no weight removes: pushed tangentially with no restoring force, a vertex slides until geometry rather than the force balance stops it. Sweeping the weight over six decades on `triwild20k_202090`, mean tracked-vertex deviation:

| `w_amips` | envelope | spring |
|---|---|---|
| 1e-1 | 2.872e-04 | 8.560e-04 |
| 1e-3 | 4.211e-05 | 8.419e-06 |
| 1e-5 | 3.981e-05 | 8.419e-08 |
| 1e-7 | **3.978e-05** (floor) | **8.419e-10** |

The envelope flattens at ~4e-5; the spring is a clean one decade per decade, no floor. Reproduce with `tools/compare_pull_energy.py` (see below).

Worth keeping in proportion: that floor is ~4e-5 against an envelope of 2.2e-2, i.e. 0.2% of eps.

## What it costs — 53 integration configs, `THREADS=0`, sequential

| `w_amips` | pull | gate | terminates | wall time | iterations |
|---|---|---|---|---|---|
| 1e-4 | envelope | on | **53/53** | 376.0s | 83 |
| 1e-4 | envelope | off | **53/53** | 377.6s (+0.4%) | 83 (+0.0%) |
| 1e-4 | **spring** | on | **51/53** | 476.7s (+26.8%) | 122 (+47%) |
| 1e-4 | spring | off | 53/53 | 461.5s (+22.7%) | 112 (+35%) |
| 1e-9 | envelope | on | 53/53 | 380.1s (+1.1%) | 85 (+2.4%) |
| 1e-9 | envelope | off | 53/53 | 373.7s (−0.6%) | 86 (+3.6%) |
| 1e-9 | spring | on | 52/53 | 531.7s (+41.4%) | 125 (+51%) |
| 1e-9 | spring | off | 51/53 | **5822.9s (+1448%)** | 155 (+87%) |

The failures are tetwild stall configs blowing their iteration assertions — `tetwild_thingi_111388_stall` (max 20) traces the whole story across the matrix: **11 → 25 → 20 → 11 → 8 → 27 → 40 (cap) → 9**.

Readings:

- **The envelope pull is insensitive to both knobs** — all four cells 53/53 within ±1.1% time.
- **Every spring cell costs 23–41%**, except `1e-9 + no gate`, which took 97 minutes against 6.3 and failed twice.
- **Removing the quality gate helps the spring at 1e-4** (fixes both failures) **and catastrophically hurts it at 1e-9**. The gate is what contains the damage once the quality term is too weak to drive convergence.
- The best spring cell still terminates on a knife edge: `111388` at exactly its limit of 20.

## Quality

Not the bar for this PR, but recorded because the regressions are large and 3D-specific:

| | max_energy worse | better | ~same |
|---|---|---|---|
| 2D | 4 | 2 | 16 |
| 3D | **7** | 1 | 9 |

`tetwild_sphere` 5.41 → 98.71, `tetwild_thingi_100827` 18.68 → 38.69, `tetwild_crown` 44.32 → 86.13, `tetwild_octocat` 52.98 → 79.89. This is the failure mode the `gauss_newton_hessian` comment predicted — surface vertices that cannot slide cannot relax a sliver — so that comment's rationale is **confirmed**, not superseded.

An earlier revision of this description claimed the spring "reaches a lower max energy, 14.33 against 16.25". That was true of the single model swept, which turned out to be one of only 2 improvements out of 22 2D configs against 4 regressions. The claim was over-read from one model and is withdrawn.

## Reproducing

```bash
# deviation sweep (the first table)
python3 tools/compare_pull_energy.py \
    --app triwild --input data/models/triwild20k_202090.obj --extra '{"stop_energy": 20}'
python3 tools/compare_pull_energy.py --app tetwild --input data/models/sphere.obj
```

It prints the deviation table and then the line carrying the claim:

```
===== shape of each curve (hausdorff_vertex_mean) =====
A quadratic penalty gives ~10x per decade of weight; ~1x means a floor.
   envelope:  4.82x   1.41x   1.05x   1.00x   1.00x   1.00x   (total 7.22x)
     spring: 10.16x  10.00x  10.00x  10.00x  10.00x  10.00x   (total 1.02e+06x)
```

Flags: `--weights`, `--outdir`, `--wmtk-app`, `--extra`, `--svg`. Only triwild reports the vertex-level figures; tetwild falls back to `hausdorff`.

The cost matrix was produced by running the 53 integration configs under each combination of `spring_pull` and `smooth_quality_gate` at `num_threads: 0`, sequentially.

## Changes

- `SpringEnergy2D` / `SpringEnergy3D`, and `SmoothVertexOptions::spring_pull`, used by `smooth_vertex_2d` and `smooth_vertex_3d`.
- `SmoothVertexOptions::smooth_quality_gate` — drops the "don't make the worst incident element worse" check for every vertex. Default `true` (unchanged behaviour).
- Both exposed as JSON in the tetwild, triwild, simwild and topological_offset specs.
- `tools/compare_pull_energy.py`.
- triwild now also reports `hausdorff_mean`, `hausdorff_vertex` and `hausdorff_vertex_mean` — the max alone hides what a change did to the bulk of the curve.

## Still not done

- The hidden `[challenging]` set has not been run.
- 3D deviation has not been swept; the 3D spring is implemented symmetrically and runs, but the accuracy claim above is 2D-only evidence.

### DIFF
--- a/components/simwild/wmtk/components/simwild/Parameters.h
+++ b/components/simwild/wmtk/components/simwild/Parameters.h
@@ -87,7 +87,7 @@ struct Parameters : public wmtk::OptimizerParameters
         eps_simplify = json_params["eps_simplify"];
 
         w_amips = json_params["w_amips"];
-        spring_pull = json_params["spring_pull"];
+        pull_mode = json_params["pull_mode"];
         smooth_quality_gate = json_params["smooth_quality_gate"];
         two_stage = json_params["two_stage"];
         num_smoothing_passes = json_params["num_smoothing_passes"];
--- a/components/simwild/wmtk/components/simwild/Parameters.h
+++ b/components/simwild/wmtk/components/simwild/Parameters.h
@@ -87,6 +87,7 @@ struct Parameters : public wmtk::OptimizerParameters
         eps_simplify = json_params["eps_simplify"];
 
         w_amips = json_params["w_amips"];
+        spring_pull = json_params["spring_pull"];
         num_smoothing_passes = json_params["num_smoothing_passes"];
         interleaved_smoothing = json_params["interleaved_smoothing"];
         interleaved_smoothing_passes = json_params["interleaved_smoothing_passes"];
--- a/components/simwild/wmtk/components/simwild/Parameters.h
+++ b/components/simwild/wmtk/components/simwild/Parameters.h
@@ -89,6 +89,7 @@ struct Parameters : public wmtk::OptimizerParameters
         w_amips = json_params["w_amips"];
         spring_pull = json_params["spring_pull"];
         smooth_quality_gate = json_params["smooth_quality_gate"];
+        two_stage = json_params["two_stage"];
         num_smoothing_passes = json_params["num_smoothing_passes"];
         interleaved_smoothing = json_params["interleaved_smoothing"];
         interleaved_smoothing_passes = json_params["interleaved_smoothing_passes"];
--- a/components/simwild/wmtk/components/simwild/Parameters.h
+++ b/components/simwild/wmtk/components/simwild/Parameters.h
@@ -88,6 +88,7 @@ struct Parameters : public wmtk::OptimizerParameters
 
         w_amips = json_params["w_amips"];
         spring_pull = json_params["spring_pull"];
+        smooth_quality_gate = json_params["smooth_quality_gate"];
         num_smoothing_passes = json_params["num_smoothing_passes"];
         interleaved_smoothing = json_params["interleaved_smoothing"];
         interleaved_smoothing_passes = json_params["interleaved_smoothing_passes"];
--- a/components/simwild/wmtk/components/simwild/SimWildMesh.h
+++ b/components/simwild/wmtk/components/simwild/SimWildMesh.h
@@ -99,7 +99,7 @@ public:
 
         optimization::deactivate_opt_logger();
 
-        m_s_envelope = 1. / (m_params.diag_l * m_params.eps * m_params.eps);
+        m_s_envelope = 1. / (m_params.eps * m_params.eps);
 
         double& wa = m_params.w_amips;
         double& we = m_params.w_envelope;
--- a/components/simwild/wmtk/components/simwild/simwild_spec.json
+++ b/components/simwild/wmtk/components/simwild/simwild_spec.json
@@ -33,6 +33,7 @@
       "w_amips",
       "spring_pull",
       "smooth_quality_gate",
+      "two_stage",
       "num_smoothing_passes",
       "interleaved_smoothing",
       "interleaved_smoothing_passes",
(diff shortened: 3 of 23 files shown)


udaykusupati:
Pushed nine commits with the measurements behind them. The short version of two weeks of sweeps:

**The fidelity floor this PR measured is real, but it is not the envelope pull's fault — it is the two-stage warm-up's.** The warm-up solves a *balanced* (1:1) blend of AMIPS and the pull before the weighted solve. That objective contains no small weight, so its line search accepts full tangential slides at any configured `w_amips`: weight-independent drift, injected every smoothing call, that the 10-iteration weighted solve cannot fully undo. The knob stopped controlling the outcome because the effective trade was set by iteration budgets, not by `w_amips`.

Evidence (triwild20k_202090, mean vertex deviation, `pull_mode=envelope`):

| w_amips | warm-up on | warm-up off |
|---|---|---|
| 1e-3 | 4.2e-05 | 2.4e-06 |
| 1e-4 | 4.0e-05 | 1.7e-07 |
| 1e-7 | 4.0e-05 (floor) | 1.7e-10 |
| 1e-12 | 1.6e-05 (floor) | 2.2e-14 |
| 0 (control) | 3e-16 | 4e-16 |

With the warm-up off, the envelope pull scales decade-per-decade with no floor and beats both spring variants at most weights. Max energy is unchanged to four digits; the 53-config corpus shows the warm-up buys nothing there either (165 vs 168 iterations, wall time within noise). The springs never showed the floor because their warm-up stage has full tangential stiffness and cannot slide.

**Corpus (53 registered configs, one binary, sequential):**

| arm | terminates | iterations vs envelope |
|---|---|---|
| envelope | 53/53 | baseline |
| envelope, no warm-up | 53/53 | +1.8% |
| spring_refresh | 53/53 | +8% |
| spring (frozen) | 52/53 (36075 stalls) | +21% |
| spring (frozen), no warm-up | 53/53 | +16% |

This reproduces your gate-commit finding from the other side: the frozen spring's stall is pinning, and per-iterate refresh (now `pull_mode=spring_refresh`, enabled by the EnergySum forwarding fix) recovers termination and most of the iteration cost while keeping the clean `1/w` fidelity scaling (within 3% of the frozen spring at every weight).

**What the commits add:** dimensionless envelope scale (2D/3D reconciled); envelope veto removed from the line search (measured no-op; feasibility stays with the accept layer); `two_stage` exposed as a parameter (default still true — flipping it is the decision this evidence points to); EnergySum hook forwarding; `pull_mode` enum superseding `spring_pull` (envelope | spring | spring_refresh | exact); the `exact` 2D mode (true PL distance with its true region-wise Hessian — FD-consistent, free sliding on segment interiors, isotropic hold at corners) plus a feature-boundary step clamp; FD unit tests; `compare_pull_energy --modes` so every table here is one command.

**Suggested resolution:** default `pull_mode=envelope` + `two_stage=false` (best fidelity, best corpus numbers, no floor), with `spring_refresh` as the fidelity-first alternative where hard vertex anchoring is wanted. Both beat the current frozen-spring default on the corpus. Happy to run the `[challenging]` suite on any pair of finalists before flipping defaults.

udaykusupati:
check #1027

PR #1027, opened 2026-08-13
Place surface vertices by a projected line search, not an envelope pull

danielepanozzo:
Replaces the envelope pull for surface vertices with an unconstrained solve followed by a projected line search. **On by default.**

## The problem

`EnvelopeEnergy` trades the surface against quality inside one objective, so a surface vertex ends up at a compromise governed by a weight — neither on the input nor where quality wanted it.

The trade also has a floor. The term penalises distance to the input **as a set**, recomputing the nearest point at every evaluation, so its gradient `2w(x − n(x))` is always normal to the input and a vertex slides along it for free. Pushed tangentially by AMIPS with no restoring force, it slides until geometry rather than the force balance stops it, and the leftover distance stops responding to the weight. Sweeping the weight over six decades on `triwild20k_202090` moves the mean vertex deviation 7× and then flattens at ~4e-5 however hard it is weighted.

## The change

Smooth the vertex as if it were interior — AMIPS alone, no pull — then walk the result back toward where it started along `t = 1, 1/2, 1/4, …`, **projecting each candidate onto the input** before testing it. Take the first projected candidate that does not invert and **strictly lowers** the worst incident element; if none does, the vertex does not move.

The vertex then lands exactly on the input by projection rather than by penalty, and no smoothing move can make the worst element worse.

## Results — 30 hidden `[challenging]` models (14 tetwild, 16 triwild)

| | main | projected | |
|---|---|---|---|
| terminates | 30/30 | **30/30** | |
| runtime | 142 min | **73 min** | **−48%** |
| iterations | 313 | 345 | +10% |
| final energy (geo. mean) | 9.957 | 9.934 | unchanged |
| average energy | 4.154 | 4.109 | −1% |
| haus / vertex (mean) | 1.13e-02 | **2.30e-03** | **−80%, better on all 30** |
| hausdorff (sampled) | 0.333 | 0.309 | −7% |

Also on the 53 registered configs: **53/53**, 328.7s against 376.0s (**−12.6%**), 94 iterations against 83.

**Why it is faster.** The solve no longer carries the envelope: one AMIPS-only solve with no BVH query inside the Newton loop, against a two-stage warm-up plus a weighted sum that queries the BVH in every value, gradient and Hessian evaluation. The projections and quality evaluations the line search adds are far cheaper than what it removes. Iterations rise 10% while wall time halves — each iteration is much cheaper.

**Why the sampled Hausdorff moves so much less than the per-vertex figure.** Once vertices sit on the input, what remains is triangles spanning whatever the input does between their corners, which no vertex placement can fix. The honest claim is "vertices are on the input", not "the output is 5× closer to the input".

### Per model

| model | iters | runtime | final E | avg E | haus/vtx | haus |
|---|---|---|---|---|---|---|
| `tetwild_101168` | 15&rarr;15 | 776&rarr;318s | 9.99&rarr;9.99 | 4.08&rarr;4.10 | 1.32e-02&rarr;7.56e-04 | 0.117&rarr;0.092 |
| `tetwild_1017012` | 22&rarr;19 | 1169&rarr;418s | 9.99&rarr;9.99 | 3.98&rarr;3.94 | 4.11e-03&rarr;4.39e-04 | 0.071&rarr;0.070 |
| `tetwild_1017020` | 15&rarr;21 | 988&rarr;621s | 9.98&rarr;9.97 | 3.88&rarr;3.89 | 5.22e-03&rarr;8.09e-04 | 0.073&rarr;0.079 |
| `tetwild_101881` | 12&rarr;14 | 837&rarr;494s | 9.99&rarr;9.99 | 3.90&rarr;3.88 | 2.68e-02&rarr;3.74e-03 | 0.192&rarr;0.176 |
| `tetwild_101954` | 7&rarr;9 | 22&rarr;17s | 9.96&rarr;9.94 | 3.79&rarr;3.75 | 1.38e-02&rarr;5.23e-04 | 0.143&rarr;0.116 |
| `tetwild_101955` | 9&rarr;7 | 18&rarr;9s | 9.91&rarr;9.57 | 3.76&rarr;3.76 | 1.03e-02&rarr;5.61e-04 | 0.120&rarr;0.115 |
| `tetwild_103197` | 11&rarr;12 | 592&rarr;243s | 9.99&rarr;9.93 | 3.96&rarr;3.88 | 8.12e-03&rarr;8.06e-04 | 0.115&rarr;0.115 |
| `tetwild_1036467` | 9&rarr;13 | 117&rarr;86s | 9.73&rarr;9.77 | 3.79&rarr;3.75 | 1.46e-02&rarr;1.26e-03 | 0.156&rarr;0.144 |
| `tetwild_104187` | 11&rarr;15 | 655&rarr;353s | 9.99&rarr;9.99 | 4.09&rarr;4.11 | 1.29e-02&rarr;6.54e-04 | 0.113&rarr;0.090 |
| `tetwild_104188` | 8&rarr;13 | 487&rarr;293s | 9.99&rarr;9.99 | 4.09&rarr;4.11 | 1.25e-02&rarr;7.00e-04 | 0.114&rarr;0.093 |
| `tetwild_104509` | 8&rarr;11 | 213&rarr;137s | 9.96&rarr;9.97 | 4.13&rarr;4.03 | 1.27e-02&rarr;1.11e-03 | 0.119&rarr;0.095 |
| `tetwild_1047980` | 11&rarr;12 | 90&rarr;48s | 9.82&rarr;9.89 | 3.66&rarr;3.67 | 5.47e-03&rarr;7.58e-04 | 0.065&rarr;0.065 |
| `tetwild_1053397` | 8&rarr;12 | 458&rarr;282s | 9.99&rarr;9.88 | 4.03&rarr;3.95 | 6.67e-03&rarr;6.81e-04 | 0.103&rarr;0.084 |
| `tetwild_108010` | 10&rarr;10 | 157&rarr;79s | 9.94&rarr;9.96 | 3.91&rarr;3.91 | 1.40e-02&rarr;3.73e-04 | 0.141&rarr;0.137 |
| `triwild_162463` | 7&rarr;9 | 129&rarr;69s | 9.98&rarr;9.98 | 2.76&rarr;2.75 | 1.56e-02&rarr;6.42e-03 | 0.562&rarr;0.562 |
| `triwild_166331` | 11&rarr;11 | 212&rarr;110s | 10.00&rarr;10.00 | 6.14&rarr;6.18 | 1.52e-03&rarr;6.82e-04 | 0.126&rarr;0.126 |
| `triwild_169900` | 13&rarr;20 | 217&rarr;116s | 10.00&rarr;10.00 | 4.37&rarr;4.30 | 1.15e-02&rarr;4.39e-03 | 1.025&rarr;0.946 |
| `triwild_177469` | 8&rarr;8 | 100&rarr;60s | 9.99&rarr;9.97 | 3.88&rarr;3.85 | 8.59e-03&rarr;4.94e-03 | 0.758&rarr;0.695 |
| `triwild_184836` | 4&rarr;6 | 6&rarr;3s | 9.99&rarr;9.69 | 3.97&rarr;3.88 | 6.38e-03&rarr;3.91e-03 | 0.765&rarr;0.762 |
| `triwild_189017` | 11&rarr;12 | 43&rarr;20s | 9.99&rarr;9.99 | 5.12&rarr;5.02 | 1.12e-02&rarr;6.98e-03 | 0.759&rarr;0.684 |
| `triwild_190586` | 9&rarr;8 | 160&rarr;85s | 9.99&rarr;10.00 | 4.02&rarr;4.04 | 4.14e-02&rarr;2.77e-02 | 1.767&rarr;1.792 |
| `triwild_191015` | 5&rarr;5 | 4&rarr;2s | 9.85&rarr;9.72 | 3.68&rarr;3.59 | 7.72e-03&rarr;2.94e-03 | 0.412&rarr;0.425 |
| `triwild_191265` | 8&rarr;8 | 191&rarr;94s | 9.98&rarr;9.99 | 4.44&rarr;4.40 | 8.45e-03&rarr;6.52e-03 | 0.546&rarr;0.573 |
| `triwild_191874` | 12&rarr;10 | 225&rarr;122s | 9.99&rarr;10.00 | 5.10&rarr;5.00 | 4.36e-03&rarr;3.31e-03 | 0.611&rarr;0.610 |
| `triwild_193105` | 11&rarr;18 | 41&rarr;23s | 9.97&rarr;9.97 | 4.09&rarr;3.98 | 8.54e-03&rarr;4.43e-03 | 0.555&rarr;0.472 |
| `triwild_193153` | 12&rarr;9 | 198&rarr;112s | 9.99&rarr;9.99 | 5.12&rarr;5.11 | 4.15e-03&rarr;2.74e-03 | 0.611&rarr;0.611 |
| `triwild_194286` | 11&rarr;9 | 37&rarr;14s | 10.00&rarr;9.99 | 5.90&rarr;5.88 | 5.55e-03&rarr;5.39e-04 | 1.253&rarr;0.932 |
| `triwild_196407` | 6&rarr;5 | 42&rarr;31s | 9.79&rarr;9.93 | 3.57&rarr;3.49 | 2.80e-01&rarr;2.41e-01 | 10.346&rarr;10.210 |
| `triwild_202260` | 19&rarr;13 | 204&rarr;76s | 9.99&rarr;9.98 | 3.89&rarr;3.60 | 1.63e-01&rarr;3.23e-02 | 3.167&rarr;3.398 |
| `triwild_202302` | 10&rarr;11 | 120&rarr;54s | 10.00&rarr;10.00 | 5.08&rarr;5.09 | 1.59e-02&rarr;1.07e-02 | 0.864&rarr;0.726 |

## Optional second pass: partial projections

`project_line_search_nested_steps`, **off by default (0)**.

The search above only ever tests points **on** the input, so a vertex whose one-ring cannot tolerate being pulled all the way there is refused at every step, does not move, and is refused again next pass from the same place.

This adds a second pass, run **only when the first pass found nothing anywhere**: for each candidate, bisect the segment from the interpolated point (`s = 0`) to its projection (`s = 1`) looking for the **largest acceptable** `s`. Since `s = 1` is already known to fail — that is exactly what the first pass established — the search brackets the boundary and converges up to it from below, rather than retreating from it. The vertex ends up off the input but closer to it, so the next pass starts from somewhere better and it can walk onto the surface over several passes instead of being stuck indefinitely.

It is **off by default** because, measured over the same 30 `[challenging]` models, it costs runtime without saving iterations:

| | nested 0 | nested 8 | |
|---|---|---|---|
| terminates | 30/30 | 30/30 | |
| iterations | 345 | 365 | +5.8% |
| runtime | 77 min | 93 min | **+20.8%** |
| max energy (geo. mean) | 9.934 | 9.946 | unchanged |

**25 of 30 models are slower and one is faster.** The cost is the search itself rather than extra passes — `tetwild_101168` takes 30% longer at an *identical* 15 iterations — because a failed vertex now runs the nested bisection over its whole one-ring, and failure is common on these meshes.

It is kept as an option because it does work where it fires (`tetwild_1036467`: 13 iterations → 6) and because it is a useful knob when tuning a difficult model. Enabling it degrades deviation (haus/vertex 7.8e-04 → 1.6e-02 over the tetwild challenging models), which is the expected trade: accepted points are deliberately *not* on the input.

`project_line_search_steps` (default 12) is exposed at the same time. Exhausting it is the definition of the projected search failing, so it also governs when the second pass runs at all.

## Risk

**This changes tetwild and triwild output**, so it deliberately breaks the byte-identity policy. `"project_line_search": false` restores the envelope pull exactly.

What has been verified: `ctest` 107/107 locally (`Integration_Tests` 193s against 225s on main), the 53 registered configs 53/53, and the 30 `[challenging]` models 30/30.

Two things a reviewer should know:

- The `[challenging]` baseline arm was **this branch with the flag off**, not a separate build of `main`. Every addition is behind the flag, so the two should be identical, but that is reasoning rather than a measurement.
- The per-vertex deviation figures needed instrumentation (`hausdorff_vertex` / `hausdorff_vertex_mean` in the tetwild and triwild reports) that is **not** included here, to keep this diff to the change itself. Happy to add it if you want the numbers reproducible in-tree.


udaykusupati:
Pushed eight commits onto this PR. They merge the measured conclusions of the #1026 investigation into the projected-line-search design: correct weight normalization, no envelope veto in the line search, warm-up off, and a second smoothing mode — the true-Hessian pull — behind a `smoothing_mode = projected | exact` switch (default unchanged: `projected`). The Gauss–Newton pull and the spring variants are retired by measurement. Receipts below.

## Why the warm-up is off (and why the old fidelity floor existed)

The two-stage warm-up solves a balanced (1:1) blend of AMIPS and the pull. That objective contains no small weight, so its line search accepts full tangential slides at any configured `w_amips` — weight-independent drift injected every smoothing call. Sweep on `triwild20k_202090` (mean vertex deviation, envelope pull):

| w_amips | warm-up on | warm-up off |
|---|---|---|
| 1e-3 | 4.2e-05 | 2.4e-06 |
| 1e-4 | 4.0e-05 | 1.7e-07 |
| 1e-7 | 4.0e-05 (floor) | 1.7e-10 |
| 1e-12 | 1.6e-05 (floor) | 2.2e-14 |
| 0 (control) | 3e-16 | 4e-16 |

Max energy unchanged to four digits; 53-config corpus: +1.8% iterations, wall within noise. The mobility rationale is structurally void: tangential gradient and Hessian both scale with `w_amips`, so Newton cancels the weight — vertices slide identically at any fitting strength.

## The exact mode

Within each closest-feature region, distance to a PL input is exactly quadratic, so the true Hessian is computable: rank-1 over segment/face interiors (free sliding), rank-2 across 3D edges and curve segments, isotropic at corners/vertices. Unlike the GN model it passes finite-difference checks (unit-tested, 2D + 3D, per region). At default `w_amips = 1e-4` its median vertex sits ~1e-4·eps-scale off the input — no tuning needed.

**On the weight.** `w_amips = 1e-4` is the simwild paper's ω, by convention, not derivation; with both terms dimensionless its one meaning is the price of shape relative to fit, and its one visible effect is the resting offset, `∝ w` (measured coefficient ≈ 8.4e-3·w on the sweep model). Sensitivity is a flat plateau roughly 1e-2 to 1e-12 wide: fidelity responds exactly linearly (one decade per decade, verified over eleven decades, no floor), while quality and convergence barely notice — max energy is unchanged across the sweep, and tangential mobility is provably weight-independent (both tangential gradient and Hessian carry the factor w, so Newton cancels it). Cliffs only at the absurd ends: above ~1e-1 the fit is genuinely weak; below ~1e-12 floating point starts eating the AMIPS term. Nobody needs to touch it, and touching it does exactly one predictable thing.

## projected vs exact, measured

53 registered configs: both 53/53; comparable wall; exact fewest iterations (76 vs 95 in our runs).

30 challenging models (geo-means, our hardware; our projected run reproduces this PR's published iterations 345 and final energy 9.934 exactly):

| | projected | exact |
|---|---|---|
| terminates | 30/30 | 30/30 |
| iterations | 345 | 322 |
| final energy | 9.934 | 9.951 |
| vertex deviation | 1.07e-02 | 1.23e-02 |
| hausdorff | 0.309 | 0.305 |

Mechanism note: projected wins the tracked-target tail (vertices land exactly on the input), exact wins the median (its pull moves every vertex every pass; the strict-improvement gate leaves split midpoints at chord distance until their projection happens to improve quality).

Thin features (two spheres, thin wall between them): both modes converge at every gap we tested, down to the harshest honest pair (gap 1e-2, eps_rel 1e-3): exact 65 min / 19 iterations, projected 82 min / 20 iterations — exact ~20% faster, same final quality. Gap thinness alone does not break the strict-improvement gate; what does is a tight envelope (next section).

## The simwild paper's Fig. 4 experiment, revisited

The simwild paper's Fig. 4 crescent (spheres nearly coincident, subtract one, remesh with topology preservation) reproduces cleanly under BOTH modes at pure defaults — smooth walls, intact rim, input polygons preserved. The locked "TetWild's smoothing" panel is not reproducible on any modern binary: we ran the legacy snap-projection proposal under today's accept layer and it converges — the historical lock needed the era's eps/2 containment and pass structure, all since fixed.

But the Fig. 4 *argument* — projection plus a tight envelope locks — is not historical. Two frontiers on the same crescent, three smoothers each (this PR's projected, this PR's exact, and the legacy snap for reference).

Quality-target descent (default eps_rel 6.7e-3, lowering stop_energy):

| stop_energy | projected | exact | legacy snap |
|---|---|---|---|
| 8 | 24 it / 145 s | 20 it / 53 s | 24 it / 121 s |
| 6 | 28 it / 728 s | 25 it / 497 s | 25 it / 604 s |
| 5 | (running) | 38 it / 2558 s | (running) |

Envelope squeeze (stop_energy 8, shrinking eps_rel):

| eps_rel | projected | exact | legacy snap |
|---|---|---|---|
| 6.7e-3 (default) | reached | reached | reached |
| 5e-4 | **stalled: E=8.21 after all 80 iterations, 2786 s** | reached, 26 it / 482 s | reached, 32 it / 860 s |
| 2e-4 | (running) | reached, 49 it / 3076 s | reached, 46 it / 3233 s |

The mechanism is the strict worst-element-improvement gate on projected candidates: with little envelope slack, most projections fail the gate and the pass burns its budget rejecting. Exact never projects during the solve — the pull term keeps every step admissible — so it is unaffected. Legacy's plain non-worsening veto also survives. (A 200-iteration budget run is separating whether the projected stall is a hard lock or asymptotic slowness; either way it is >5× exact's cost at this rung.)

This bears on the default. On the easy regimes the two modes are interchangeable; on every hard axis measured so far — tight envelopes, low quality targets, thin gaps — exact is faster or the only finisher, with no weight tuning required. We're leaving the default at `projected` in this PR and will propose the flip separately once the remaining rungs land.

## Also in the commits

`EnergySum` now forwards `solution_changed` / `after_line_search_custom_operation` / `max_step_size` to its children (they were silently swallowed — latent bug, nothing depends on it today). Triwild reports vertex and mean deviation next to the sampled Hausdorff (ported from #1026), so every table above is reproducible from `report.json` without external tooling.

Cells marked (running) — projected/legacy at stop 5, exact at stop 4, projected at eps_rel 2e-4, and the 200-iteration lock-vs-slow run — will be posted as a follow-up comment; none of them can change the code in this PR, only the future default recommendation.

PR #1035, opened 2026-08-14
Coarsen the mesh without letting the max energy rise (2D and 3D)

danielepanozzo:
## What

A **coarsening pass** for both optimizers, 2D and 3D: a collapse taken optimistically, its neighbourhood re-smoothed, and the whole block undone unless the worst element in the region it touched came out no worse than it went in. This is the toolkit's first composite operation.

Plus the generic n-ring lock it needs — added alongside the existing lockers, **not** replacing them, for measured reasons.

## Why

`collapse_edge_before` refuses any collapse whose resulting cells are worse than the ring they replace, and it decides that on the raw post-collapse geometry — the worst moment in the operation's life, before smoothing has had any chance to absorb the damage. A collapse that would be perfectly fine once its neighbourhood relaxes is therefore never taken, and the converged mesh carries elements it does not need.

The new pass judges the same collapse on a *re-smoothed* neighbourhood instead. It runs once, at the end of `mesh_improvement`, and its contract is narrow: remove as many elements as it can **without letting the max energy rise**.

## Results

Three arms, run serially. `off` / `bounded` (`coarsen_unbounded: false`) / `unbounded` (the default).

| | off | bounded | **unbounded (default)** |
|---|---|---|---|
| **11 registered tetwild** — cells | 348,366 | 197,837 (−43.2%) | **60,247 (−82.7%)** |
| wall clock | 371s | 1116s (+201%) | 3183s (+758%) |
| **16 challenging triwild** — faces | 180,425 | 168,830 (−6.4%) | **164,861 (−8.6%)** |
| wall clock | 768s | 1151s (+50%) | 1240s (+61%) |

**Max energy is raised on no model in any arm**, and the pass-internal invariant (`coarsen_max_energy_after ≤ before`) holds everywhere. On several models it *improves*: `thingi_100827` goes 24,505 → 591 cells with max energy 43.2 → 10.7.

The asymmetry between the dimensions is the interesting part. In 2D the target-length bound costs most of the effect (−6.4% vs −8.6%), because triwild's converged meshes already sit near target length and the slack was mostly above it. In 3D the bound barely touches the first −43%: those are short edges the quality gate refused, entirely below target. Unbounded then goes much further, and that second tranche is exactly the part that changes element size.

Hausdorff moves by more than 5% on two models per 3D arm, worst `octocat` 0.0774 → 0.0848 unbounded. The envelope check still gates every collapse, so this is redistribution *inside* the tolerance rather than anything escaping it — but it is not zero.

## The composite, and why it needs no new machinery

`collapse_edge` already calls `collapse_edge_after` **inside** `start_protect_attributes()`, and on `false` runs the rollback path → connectivity restored → `rollback_protected_attributes()`. Three existing properties make the composite fall out of that for free:

- `AttributeCollection::operator[]` journals with `emplace` — **first write wins** — so however many times the smoothing rewrites a position, the recorded value is still the pre-collapse one.
- The destructive `begin_protect()` (which *clears* the journal) lives only in `smooth_vertex`. `smooth_after` forwards straight to `optimization::smooth_vertex_{2,3}d` and touches no protection state, so calling it from inside the hook nests correctly.
- Rollback restores the old cell `hash` with the connectivity record, so tuples minted before the attempt are valid again afterwards — not merely invalidated.

**No change to `AttributeCollection`, no scope stack, no journal.** The one thing that does not come for free is per-vertex granularity: `smooth_vertex_{2,3}d` writes the position and the one-ring qualities and only *then* decides whether to keep them, deliberately — its contract puts the undo on the caller. One rejected vertex would otherwise poison the whole composite, so each individual smooth gets a narrow save/restore, which nests correctly inside the outer journal precisely because of first-write-wins.

In 3D the composite reaches cell quality through the `cell_quality` / `set_cell_quality` virtuals, so it serves tetwild and simwild-3D without touching either app's cell attributes.

## Why the local test implies the global one

Every cell outside the smoothed vertices' incident set is untouched. The before-max is taken over the pre-collapse r-ring of `{v1, v2}`, which is a superset of the after-set's pre-images — a collapse only shortens paths, so post-collapse distance-d from the merged vertex is contained in pre-collapse distance-d from the edge. So `after_local ≤ before_local` gives `after_global ≤ before_global`. `≤` rather than `<` is what lets a region coarsen at all when its worst element is unchanged and the global max lives elsewhere.

## A pre-existing crash the port surfaced

`TetMesh::collapse_edge_conn` probes for a neighbouring tet to name its return vertex with, falls back to a second probe on the boundary, and then calls `.value()` unconditionally:

```cpp
std::optional<SmartTuple> v_C_tuple = tt.switch_edge().switch_face().switch_tetrahedron();
if (!v_C_tuple.has_value()) {
    boundary_flag = true;
    v_C_tuple = tt.switch_vertex().switch_edge().switch_face().switch_tetrahedron();
}
const size_t v_C = v_C_tuple.value()...   // the fallback can be empty too
```

When *both* probes run into a boundary face this threw `bad_optional_access` and took the process down. Nothing has mutated the mesh at that point, so it now refuses the collapse — inert on any input that does not currently crash. Normally unreachable, because `collapse_edge_before` will not move a vertex off the bbox or the surface and a tetwild mesh tags every boundary vertex; the coarsening pass drops the quality gate, which lets far more candidates reach it.

## Locking

`try_set_vertex_mutex_n_ring(v, tid, n)` / `try_set_edge_mutex_n_ring` / `try_set_face_mutex_n_ring`: one breadth-first ball per mesh, expanded only through vertices the thread already holds. `release_vertex_mutex_to(mark)` lets a failed acquisition unwind to its own watermark. The coarsening pass locks `coarsen_smooth_ring + 1`, because smoothing a vertex at distance r reads its one-ring and writes the quality of its incident cells.

**The existing `*_two_ring` / `*_one_ring` helpers are deliberately left alone.** The first cut routed every pass through the ball, on the argument that they under-lock. Two things came out of measuring that:

- **It costs.** At 16 threads on the challenging set: **+80% wall clock on the 14 tetwild models (844s → 1522s)**, −0.4% on the 16 triwild ones. A tet vertex has ~30 neighbours, so the honest 2-ring ball is enormous next to what those helpers claim, and the whole difference is contention and retries. After reverting, `challenging_tetwild_101168` retimes at 70.0s / 84.1s against main's 75.9s / 85.7s — inside the run-to-run spread.
- **The argument was weaker than I first stated.** They do claim less than their names promise: each `continue`s past a vertex the thread already owns and so never expands through it. But that is a coverage-vs-name mismatch, not a demonstrated race. What the operations rely on is holding the complete **one**-ring of the seed simplex, and they do — the outer loops lock every unowned neighbour directly and only skip *expanding* through owned ones. That set is the operation's entire write set, and two operations interfere only if their one-rings meet, in which case they contend on the shared vertex.

The non-equivalence is documented where someone would trip over it: a "Ring lockers — NOT balls" note on the declarations in `TriMesh.h` / `TetMesh.h` with the mechanism, reasoning and measurement; a `@warning` on `try_set_vertex_mutex_n_ring`; a `@warning` on `default_ring` that `make_locker` routes to the named helper at that radius and only there; and a test, `named_two_ring_helpers_are_not_balls`, pinning both facts.

One genuine bug fixed on the way: `try_set_vertex_mutex_one_ring` nested the seed lock and the ring lock, so a caller that already held the seed got `true` back having claimed **nothing**. Unreachable through the executor, so the fix is inert until something composes lock acquisitions.

## Parameters

In `triwild_spec.json`, `simwild_spec.json` and `tetwild_spec.json`.

| key | default | meaning |
|---|---|---|
| `coarsen_pass` | `true` | run the pass at all |
| `coarsen_unbounded` | `true` | coarsen as far as the quality guarantee allows; `false` stops at the target edge length |
| `coarsen_local_smoothing_passes` | `2` | smoothing sweeps inside each candidate collapse |
| `coarsen_smooth_ring` | `2` | radius re-smoothed; the lock claims this + 1 |
| `coarsen_global_smoothing_passes` | `1` | smoothing between coarsening rounds |
| `coarsen_max_rounds` | `2` | cap on the alternation; also stops when a round accepts nothing |

Two of those defaults were themselves measured:

- **`coarsen_max_rounds: 2`.** A round is not there to finish what the previous one started — within a round the collapse pass already runs to a fixed point. What a round adds is the global smoothing in between, which *moves* that fixed point, and which the dirty-epoch retry cannot see (it only re-offers failures next to a successful collapse). Accepted collapses by round over the 16 challenging triwild models were 68.8% / 21.7% / 7.2% / 1.8% / 0.5%; five rounds gave −9.6% faces for +134% wall clock, two give −8.6% for +61%.
- **`coarsen_unbounded: true`.** The sizing FIELD is deliberately never applied — it is the optimizer's own refinement scratch work, and honouring it would leave the pass unable to undo refinement that turned out to be unnecessary. The bounded arm gates on the target length instead, at the same `collapsing_l2` threshold the ordinary collapse uses.

The report gains `coarsen_accepted`, `coarsen_{f,t}_before`, `coarsen_{f,t}_after`, `coarsen_max_energy_before`, `coarsen_max_energy_after` — all zero when the pass is off, which is also how an A/B tells the arms apart.

`Integration_Tests` no longer fits ctest's 1500s default (it timed out), so it is raised to 7200s on discovery.

## Validation

**`ctest`: 114/114 Release.** `Integration_Tests` takes 2443s against the 7200s ceiling — up from 209s before the pass, which is the coarsening running at the end of every tetwild and triwild config.

**Inertness gate.** With `coarsen_pass: false`, all 56 registered integration configs at `num_threads: 0` are **byte-identical to main** — every produced mesh matches by md5, and the only report differences are the new keys, on exactly the 24 configs that emit them. Re-run after the locking revert and again after the 3D port; same result each time. (`simwild_energies_2d` fails on both arms; pre-existing.)

**New tests.**
- `tests/test_ring_lock.cpp` — the ball equals an independently computed BFS ball for n = 0..3, seeded from a vertex, an edge and a tet; a failed acquisition unwinds to its entry watermark and leaves the caller's locks alone; acquiring incrementally gives the same set as acquiring at once; a radius past the mesh diameter terminates; and the named helpers are a strict subset of the ball but a superset of the one-ring.
- `components/{triwild,tetwild}/.../test_coarsen_pass.cpp` — a refused coarsening collapse leaves connectivity, positions, qualities **and cell hashes** exactly as it found them; the pass removes elements without the max energy rising; `coarsen_unbounded` gates on the target length in both directions; the pass is a no-op when off.

One note on those tests: the "stored quality still describes the geometry" check compares to a **relative tolerance, not exactly**. `collapse_edge_after` stores what `collapse_edge_before` computed from the *predicted* post-collapse vids, and recomputing reads the vids the connectivity ended up with — often a permutation of the same tuple. AMIPS is permutation-invariant mathematically but not bit-for-bit, and an exact comparison there was flaky.

**Not done:** Debug-build ctest.


udaykusupati:
## Follow-ups from the `quality_rel` fix (735f2b040a)

That commit fixed one thing: the coarsening composite maxed **raw** quality across a region and compared it against the same before. Correct for tetwild and triwild, which have one global target; wrong for simwild, whose target is per tag, because the maximum is then taken across cells whose numbers do not mean the same thing.

Getting there surfaced five things that are real but did not belong in a correctness fix. Listing them here rather than in the commit so they can be argued separately.

> **The naming work (item 1) is handled separately in #1041**, branch `quality-naming`. It is opened empty on purpose — the scope wants agreeing before anything is renamed. Items 2–5 have no home yet.

### 1. "quality" vs "energy" — one quantity, two words, no rule

*Tracked in #1041.*

`m_quality` holds an energy. `stop_energy` is a quality target. `optimization_quality_stats` returns things the callers name `max_energy`. `filter_energy` is a quality threshold.

The scope is smaller than it looks, and the obvious sweep is wrong. The word is doing two legitimate jobs:

| genuinely an energy — leave alone | actually a quality — misnamed |
| --- | --- |
| `slim_energy`, `compute_energy`, `total_energy`, `energy_from_point`, `amips_energy`, `envelope_energy`, `add_energy`, `new_energy` | `stop_energy` (105 uses), `max_energy` (125), `avg_energy` (46), `filter_energy` (25) |

The left column is the solver objective being minimised; renaming those to "quality" would be actively wrong. So this is **four identifiers**, not a sweep.

Worth being honest about what it buys: readability, not correctness. The bug fixed in 735f2b040a was raw-vs-normalised, and *both* of those were already called "quality" — this rename would not have prevented it.

Mechanics, if it goes ahead. `stop_energy` is a user-facing spec key. jse has no alias or deprecation facility (the whole API is `verify_json`, `inject_defaults`, and a `strict` flag), and simwild runs `strict = true`, so an old key is a hard validation failure rather than a silent ignore — loud, which is the good failure mode, but it does mean every existing config breaks on the day it lands. Suggested shape: a ~10-line pre-pass in `DriverPrologue` that moves the old key to the new name and logs a deprecation warning, so the spec only ever knows the new one.

### 2. `report["max_energy"]` means two different things

`report["max_energy"]` calls the **non-virtual** `get_max_avg_energy()`, which is raw AMIPS. simwild's stop criterion and its logs use `optimization_quality_stats()`, which it overrides to be relative. Same mesh, same run, two conventions.

This is not theoretical. A/B-ing 735f2b040a on `simwild_double_sphere_3d`:

|  | baseline | fixed |
| --- | --- | --- |
| tets after the pass | 4036 | **3280** (−18.7%) |
| max **relative** quality (the stop metric) | 0.565 | **0.549** |
| report `max_energy` (raw AMIPS) | 5.74 | **26.22** |

Read off the report key alone, the fix looks 4.6× worse. Read off the metric simwild actually stops on, it is better on both axes — 19% fewer cells at a lower max. The raw number rises because ambient finally spends the slack its target of 50 grants it (0.549 × 50 = 27.5).

### 3. The new virtual has four call sites waiting for it

`utils::active_vertices` and `utils::select_worst_cells` both take a `(quality lambda, threshold)` pair, and each app passes it in its own units:

| call site | quality lambda | threshold |
| --- | --- | --- |
| `TriOptimizerMeshSmooth.cpp:85` | `m_face_attribute[fid].m_quality` | `margin * stop_energy` |
| `SimWildMeshTri.cpp:474` | `quality_rel(fid)` | `margin` |
| `TriWildMesh.cpp:345` | `m_face_attribute[fid].m_quality` | `filter_energy` |
| `SimWildMeshTri.cpp:524` | `m_quality / target_quality(fid)` | `1.0` |

That is the same concept written out by hand, plus `quality_rel` as a named member in both simwild meshes, plus another divide inside `refine_sizing_around_worst`. Six instances. The coarsening pass was simply the first shared site written *without* an injection point, which is why it started out wrong and stayed wrong.

Caution against over-unifying: triwild's `filter_energy` is a dynamic clamp recomputed per call, not a per-cell requirement. "Target" is not one concept even inside one app, so only the two `active_vertices` sites are candidates, and each needs its own before/after.

### 4. simwild's 3D collapse gate ignores its own per-region target

`SimWildMeshTri::collapse_quality_allowed` uses `target_quality(fid)`. `SimWildMesh::collapse_quality_allowed` (3D, `EdgeCollapsing.cpp:14`) just forwards to the base with an on/off switch. So a non-uniform `quality_field` steers 3D's stats and worst-cell selection but not its collapse gate. Pre-existing, one line to change, but it is a behaviour change for existing 3D pipelines and the 3D base gate is shape-different from 2D anyway (no `stop_energy` term at all). Wants its own measurement.

### 5. Bounded coarsening drops user intent along with optimizer scratch

Bounded, the pass rejects `edge_length² > collapsing_l2`. The ordinary collapse tests `edge_length² > collapsing_l2 * sizing_ratio²`. The stated reason for dropping the sizing field — it is the optimizer's own refinement scratch, and honouring it would leave the pass unable to undo refinement that turned out to be unnecessary — is sound for the adaptive part.

It does not cover the other writer. simwild's user-facing `sizing_field` json min-clamps into the *same* `m_sizing_scalar` array that `refine_sizing_around_worst` writes (`SimWildMeshTri.cpp:582`). User intent and optimizer scratch are conflated in one scalar, so ignoring the field discards both.

### One note on the fix itself

It is **not** bit-identical for tetwild/triwild, despite the divisor being constant. The transform is monotone but not injective, so it can collapse a strict `>` into `==`. Measured over 4M pairs at realistic magnitudes: that happens when the two maxima are within ~1 ULP (2D) or ~3 ULP (3D — the cube root compresses roughly three adjacent doubles into one), never beyond 10 ULP, and never in the other direction. So the guarantee is "the max cannot rise by more than ~1e-15 relative", not "cannot rise". Worth expecting hairline diffs rather than assuming a clean pass.

On coverage, so nobody leans on the wrong thing. The change ships with a unit test — `triwild-coarsen-judges-on-the-per-face-target`, on a `TriWildMesh` subclass carrying a per-face target, since triwild's own is uniform and cannot tell raw from relative apart. Reverting `region_max_quality_rel` to the raw read fails it and nothing else, so it is known to discriminate rather than merely pass.

What does **not** cover it is the 54-config integration gate, in two separate ways. The inertness run quoted in this PR was made with `coarsen_pass` off, and with the pass off `region_max_quality_rel` never executes at all. And with it on — the spec default, so every registered config runs it — exactly one config uses a non-uniform `quality_field` (`simwild_double_sphere_3d`, ambient 50 vs bodies 10), and that is the only one where raw and relative can possibly disagree. Everywhere else the divisor is constant and cancels. So a green suite says the other 53 were not perturbed; it says nothing about whether the fix works.

PR #1037, opened 2026-08-14
Stop paying for a lock and two counters the hot path does not need

danielepanozzo:
Stacked on #1036 — **review that one first**; this PR's base is `concurrency-vertex-mutex-race` and it will retarget to `main` once #1036 lands.

Three pieces of per-operation overhead in the scheduler's inner loop, none of which buys anything.

## 1. The per-task queues were locking

Every task gets its own queue, seeded before any thread starts, and both pops from and pushes its renewed operations back into that same queue. The only queue that genuinely crosses threads is `final_queue` (retry overflow, drained after the barrier).

So the thread-private heaps were taking and releasing a `std::mutex` on every pop and every renewal for the benefit of nobody. They now use a new `serial_priority_queue` — same interface, no mutex — and only the shared queue stays concurrent. Both are `std::priority_queue` with the same comparator underneath, so pop order, and therefore output, cannot change.

## 2. `cnt_success` / `cnt_fail` were seq_cst atomics bumped per operation

One shared cache line ping-ponging between all cores, millions of times per pass. Now accumulated per task and folded in once, via an RAII guard so the loop's early returns can't skip the flush.

That wasn't possible for `cnt_success` while the stopping-criterion check read it live mid-pass, so that read goes through a separate counter maintained **only when a stopping criterion is actually configured** — never for tetwild/triwild/simwild, and exactly as before for qslim and shortest_edge_collapse.

## 3. `cnt_update` was dead

Incremented on every success, reset inside the stopping-criterion check, never read anywhere.

The reset made `stopping_criterion_checking_frequency` look like a period. It isn't. Both users set the criterion to `return true` and the threshold to the number of collapses needed to hit a target vertex count — it's a decimation counter that stops the pass on its first check. The member is gone and the field is now documented for what it actually is, including the trap for anyone who later writes a genuinely periodic criterion against it.

## Verification

The decimation path is the one behavior this touches, so I ran all six qslim / shortest_edge_collapse integration configs before and after. **Operation counts identical in every one:**

| config | executed | success / fail |
|---|---:|---|
| qslim_double_sphere | 266 | 259 / 7 |
| qslim_octocat | 17050 | 17050 / 0 |
| shortest_edge_collapse_101633 | 94108 | 78938 / 15170 |
| shortest_edge_collapse_double_sphere | 217 | 115 / 102 |
| shortest_edge_collapse_octocat | 17103 | 17050 / 53 |
| shortest_edge_collapse_sphere_with_env | 39 | 39 / 0 |

All six set `throw_on_fail`, so reaching the target is itself the assertion. Full build clean, all 99 unit test cases pass.

## What is *not* claimed

**No speedup.** These are strictly-less-work changes, but their magnitude against operations that do envelope BVH queries and exact predicates is unmeasured — the shared-cache-line counters are the more plausible of the two, the queue mutex is likely noise. The instrumentation to settle it is the next patch in this stack, and I'll post numbers there rather than assert them here.

Also drops a `concurrent_priority_queue` include from `DelaunayBoxMesh.cpp`, which never instantiated one.

### DIFF
--- a/components/tetwild/wmtk/components/tetwild/DelaunayBoxMesh.cpp
+++ b/components/tetwild/wmtk/components/tetwild/DelaunayBoxMesh.cpp
@@ -10,7 +10,6 @@
 #include "wmtk/utils/InsertTriangleUtils.hpp"
 #include "wmtk/utils/Logger.hpp"
 
-#include <wmtk/threading/concurrent_priority_queue.hpp>
 #include <wmtk/threading/enumerable_thread_specific.hpp>
 #include <wmtk/threading/parallel_for.hpp>
 #include <wmtk/threading/task_group.hpp>
--- a/src/wmtk/ExecutionScheduler.hpp
+++ b/src/wmtk/ExecutionScheduler.hpp
@@ -3,6 +3,7 @@
 #include <wmtk/TetMesh.h>
 #include <wmtk/TriMesh.h>
 #include <wmtk/threading/concurrent_priority_queue.hpp>
+#include <wmtk/threading/serial_priority_queue.hpp>
 #include <wmtk/threading/task_group.hpp>
 #include <wmtk/utils/Logger.hpp>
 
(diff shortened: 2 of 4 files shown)


danielepanozzo:
## Measured: no detectable performance effect. Merge this as cleanup, not optimization.

I said I'd post numbers here rather than assert a speedup. Here they are, and they do not show one.

Model: `challenging_tetwild_103197` (128k tets), idle machine, paired design — the two binaries alternate within each repetition so thermal drift hits both arms equally.

| bench | after − before | t | verdict |
|---|---:|---:|---|
| `nt=1`, uncontended (deterministic, identical 20,943,019 ops) | **−0.3% ± 0.8%** | −0.7 | indistinguishable from zero |
| `nt=16`, contended (µs/op, work-normalised, 5 pairs) | **−2.2% ± 4.0%** | −1.3 | indistinguishable from zero |

Directionally faster in both, resolvable in neither.

I had guessed the contended counters were "the more plausible of the two wins." That was wrong at any magnitude I can measure. The reason is visible in #1038's numbers: at 16 threads the pass spends ~3.35 µs/op on lock acquisition and ~6 µs/op on real work, so a couple of atomic RMWs and an uncontended mutex per operation are simply not where the time goes.

**This does not argue for reverting.** What it does argue is that the PR should be judged on what it actually is:

- deletes a `std::mutex` taken per pop and per renewal on queues only one thread ever touches
- deletes `cnt_update`, which was incremented on every success and never read
- documents `stopping_criterion_checking_frequency`, which is a decimation threshold and not the period its name and the old reset implied — the trap being that a future periodic criterion written against it would run on every operation

All three are strictly-less-work or clarity changes, verified behaviour-preserving on all six qslim/SEC configs. Merge it for that, not for the time it saves, which is none I can measure.

PR #1038, opened 2026-08-14
Instrument the scheduler pass: contention, overflow and the serial tail

danielepanozzo:
Stacked on #1037, which is stacked on #1036. Retargets to `main` as those land.

## Why

The pass could not be asked how much of its time went into contention rather than work. Acquisition failures, retries and the whole post-barrier drain were invisible — and `RunPass` bills the entire span, drain included, as `"... time parallel: N s"`. Any argument about whether the locking scheme needs replacing had nothing to stand on.

## What

A `PassStats` on `ExecutePass`, populated every `operator()` call:

- ring-acquisition failures (attempts, so one stubborn op can contribute up to `max_retry_limit`)
- operations that exhausted `max_retry_limit` into the serial queue, and that queue's size at the barrier
- wall time split between the parallel region and the serial tail
- busiest / idlest task time

All per-task locals folded in once, or a write to the task's own slot, so it adds nothing to the inner loop. Log lines are guarded on the debug level, and `WMTK_LOGGER_LEVEL=debug` already exists — no new plumbing:

```
WMTK_LOGGER_LEVEL=debug ./wmtk_app -j tetwild_octocat.json
```

## First numbers — and they are not subtle

tetwild Octocat, 10 threads, representative passes:

| failures/op | serial tail | busiest / idlest task |
|---:|---:|---|
| 0.29 | 11.0% | 0.5788s / 0.3975s |
| 3.23 | **70.3%** | 1.349s / 0.4034s |
| 3.58 | **76.9%** | 0.306s / 0.1036s |
| 4.22 | 64.8% | 0.1161s / 0.02967s |
| **17.22** | 38.0% | 0.3086s / 0.2447s |

Three things fall out:

1. **Ring acquisition fails 3–4× per executed operation** on a typical pass, 17× on the worst. Contention is not marginal.
2. **Two thirds to three quarters of a "parallel" pass is the serial drain.** The retry-then-serialize policy is converting contention into serial work at a rate nobody could see.
3. **The busiest task runs ~3× the idlest.** Expected, given the partition is by initial vertex count and tasks never steal — but now it is measured.

## Caveats

The failure counts are counts and hold regardless of machine load. **The timings were taken while another job had ~4 cores busy** and want re-running on an idle machine before anyone leans on the exact percentages. I'll post idle numbers as a follow-up comment rather than bake them into the commit message.

This PR only measures. It changes no policy — no redesign is proposed here.

### DIFF
--- a/src/wmtk/ExecutionScheduler.hpp
+++ b/src/wmtk/ExecutionScheduler.hpp
@@ -14,8 +14,10 @@
 #include <wmtk/utils/EnableWarnings.hpp>
 // clang-format on
 
+#include <algorithm>
 #include <atomic>
 #include <cassert>
+#include <chrono>
 #include <cstddef>
 #include <queue>
 #include <stdexcept>
@@ -335,6 +337,15 @@ public:
         std::vector<LocalQueue> queues(num_threads);
         SharedQueue final_queue;
 
+        // Contention accounting. Everything here is either a per-task local folded in once or a
+        // write to the task's own slot, so it adds nothing to the inner loop. It answers the
+        // two questions the pass could not previously be asked: how often ring acquisition
+        // loses a race, and how much of a "parallel" pass is really the serial drain.
+        m_stats = PassStats{};
+        std::atomic<size_t> lock_failures(0);
+        std::atomic<size_t> overflowed(0);
+        std::vector<double> task_seconds(queues.size(), 0.);
+
         auto run_single_queue = [&](auto& Q, int task_id) {
             // Per-task tallies folded into the shared counters once, on the way out. The guard
             // is RAII rather than a line at the bottom because the loop below has early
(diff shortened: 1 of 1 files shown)


danielepanozzo:
## Idle-machine measurements, on a challenging model

Re-run as promised, on `challenging_tetwild_103197` (128k tets, 23k vertices, converges to 9.987) rather than Octocat. Octocat turned out to badly overstate contention — see below.

### The headline: lock acquisition costs 56%, and it is *not* contention

`num_threads=0` disables locking entirely; `num_threads=1` runs the full lock machinery with one thread, so **zero contention**. Both are deterministic — all six runs executed **20,943,019 operations** and ended at **9.96271554299127**, identical to 16 digits. Same work, so the difference is purely what locking costs.

| | optimization time |
|---|---|
| `nt=0` — no locks | 125.26s (sd 0.22) |
| `nt=1` — locks, no contention | 195.46s (sd 2.01) |

**+70.19s ± 1.30 → +56.0%, t = 53.9.** Three alternating pairs, so thermal drift cancels.

The instrumentation confirms the isolation: at `nt=1`, **0 ring-acquisition failures and 0 serial tail** across 136 passes. Nothing is retried. That 70 seconds is purely walking two-rings and taking/releasing hundreds of vertex locks per operation — about 3.35 µs/op of lock overhead against 5.98 µs/op of actual work.

### Octocat overstated contention badly

| | Octocat | 103197 |
|---|---:|---:|
| ring-acquisition failures / op | 3–17 | **1.06** |
| ops overflowing to the serial queue | — | **7.2%** |
| serial tail | 70–77% | **21.9%** |

The numbers in the commit message came from Octocat and should be read as a small-mesh artifact. On a realistic model contention is real but secondary; **acquisition overhead dominates**.

### Scaling

| threads | optimization | vs serial | CPU used |
|---|---:|---:|---:|
| 0 | 125.3s | 1.00x | 125s |
| 1 | 195.5s | 0.64x | 195s |
| 4 | 84.6–105.4s | 1.2–1.5x | 273–329s |
| 16 | 48.3–71.8s | 1.7–2.6x | 386–582s |

16 threads burn 386–582s of CPU for a 1.7–2.6x speedup over serial.

### Methodology note

My first attempt at this read as a 25% regression and was entirely measurement error, from two compounding sources worth recording:

1. **Parallel runs are non-deterministic in the amount of work they do** — the same binary and model executed 19.2M to 30.3M operations across runs, a 35% spread, each ending at a different energy. Wall-clock comparison is meaningless; everything above is normalised to µs/op.
2. **Thermal drift** across a sustained-load sequence — the same binary measured 2.437 µs/op early in a session and 3.892 µs/op ten minutes in. Fixed by alternating the two binaries within each repetition so drift hits both arms.

Anyone repeating this should do both, or they will measure their machine rather than their change.

PR #1040, opened 2026-08-14
Offset 3d Optimization

udaykusupati:
Carries Spencer's 3D offset optimization work forward, rebased onto current `main`.

## What is in this PR right now

Two commits, authorship preserved, rebased cleanly onto `0086e56a52`:

- `3D offset: bring the component into lock-step with 2D`
- `3D offset: pool the quadric samples, reject zero-length smoothing moves`

The rebase is behaviour-neutral: on `127891` (prism, δ = 0.8368, `convergence_target` 0.0418, 5 iterations) every reported metric is bit-identical before and after.

| iter | max dist err | avg dist err | max normal dev | avg normal dev |
| ---- | ------------ | ------------ | -------------- | -------------- |
| 1    | 0.4046       | 0.0411       | 47.7°          | 17.6°          |
| 5    | 0.1647       | 0.0595       | 37.8°          | 19.7°          |

It does not converge, and `avg_dist_err` rises every pass — the "known gap" recorded in the first commit message.

## Why it does not converge

Not the smoother. **The offset surface is being decimated rather than refined**: 1172 → 490 faces over five iterations, ending 7.5× coarser than the sizing field asks for, with 724 of its 747 edges above the split gate and nothing splitting them.

Four separate defects, each measured on prism:

1. **Offset vertices are invisible to the shared engine.** `m_vertex_attribute[].m_is_on_surface` is never set anywhere in the 3D component. 2D sets it in `label_offset_boundary()`. `TetOptimizerMesh::is_edge_on_surface()` short-circuits on that flag, so the collapse's link condition, `preserve_topology`, and the split's propagation of the flag to the new vertex never apply to the offset. Instrumented: 0 of ~1700 offset-edge split attempts per iteration are seen as surface edges; 1727 once the flag is set.

2. **Most splits are silently discarded.** `TetMesh::split_edge` returns false at `get_next_empty_slot_v() < 0`, before `split_edge_after`, so no application hook runs and nothing is logged. 31,445 splits lost per iteration; 19 of the ~723 offset edges the sizing field asked to split actually got split. The per-iteration `consolidate_mesh()` already on this branch fixes the between-iteration case, not this one.

3. **The whole split budget goes to the ambient mesh.** `#V` 630 → 3270 per split pass, of which ~20 land on the offset. The ambient mesh is remeshed toward `length_rel × bbox_diagonal`, a target unrelated to the offset, and the collapse pass undoes it every iteration. The reference implementation has no global length target for the embedding at all: its target edge length is per-edge, initialised to each edge's own current length and updated by AMIPS, and embedding operations are confined by `RoiInvariant` to tets with AMIPS > 100 plus a 2-ring spread that does not cross substructures.

4. **Split and collapse fight each other.** Splitting an edge just over `(4/3)t` yields children at `(2/3)t`, below the `(4/5)t` free-collapse gate, and nothing rejects a collapse that recreates an over-long edge.

## Planned commits on this branch

1. Set `m_is_on_surface` at both labelling sites. *(defect 1)*
2. Add `TetMesh::slot_exhausted()`, make the split pass consolidate-and-repeat instead of losing splits silently. *(defect 2)*
3. Score distance per face over the four samples (centroid + three near-corner), as the reference does, instead of at vertices only — in `compute_distance_deviation` and the sizing `dist_bad` trigger. Vertex sampling is a lower bound and is satisfied by construction by any projection-based placement, so it cannot be tuned against.
4. Port the reference's embedding sizing rule and ROI restriction. *(defects 3 and 4; 4 may fall out of 3 for free)*
5. Distance adaptation (paper §5.3.1), or an explicit decision not to have it. With the remeshing fixed, the worst vertex is consistently *too close* — 0.64 against δ = 0.837 — and sits at the same interior pinch every iteration, with no domain-boundary warning. `max_dist_err` against a global δ is measuring a target that cannot be met there.
6. The disabled candidate-selection placement in `Smooth.cpp`, only if 3–5 leave a residual it can explain.

A prototype of 1, 2 and crude forms of 3 and 4 takes avg dist err 0.0595 → 0.0148 and avg normal dev 19.7° → 6.6°, with the offset surface growing 490 → 9050 faces. `max_dist_err` does not converge under it, which is what points at item 5.

### DIFF
--- a/components/topological_offset/wmtk/components/topological_offset/Collapse.cpp
+++ b/components/topological_offset/wmtk/components/topological_offset/Collapse.cpp
@@ -110,6 +110,55 @@ bool TopoOffsetTetMesh::collapse_before_vertex(
         return false;
     }
 
+    // BOTSCH-KOBBELT'S COLLAPSE GUARD: refuse a collapse that would CREATE an over-long
+    // edge. Bisection necessarily drops children into the collapse zone -- a just-over-4/3
+    // parent halves to just-over-2/3, below the 4/5 gate -- so without this rule,
+    // collapse-to-convergence unravels every freshly refined region: merging two children
+    // recreates the parent, and the cascade runs until the surface sits far past its own
+    // sizing target. Measured on specific_models/prism: without the guard the run pins at
+    // max_dist_err 0.162 (bit-stable over 30 iterations) with the offset at ~1100 faces
+    // and 96% of its edges over-gate; with it, refinement survives the collapse pass for
+    // the first time and the run converges once the sizing field is confined to the band
+    // (see refine_sizing_around_worst). B-K stabilises the same hysteresis the same way;
+    // an edge that was already over-long is exempt, so degenerate cleanup (v1 ~ v2) stays
+    // possible.
+    //
+    // VALENCE ESCAPE: the guard alone traps slivers -- splits keep pumping valence into
+    // vertices whose relieving collapses the guard refuses (measured: valence 240, 32
+    // vertices over the split threshold, 18x runtime from the retry grind). A collapse
+    // deletes the tets around edge (v1,v2) and so is the operation that RELIEVES valence;
+    // when any vertex it touches is past the engine's own high-valence threshold, that
+    // takes precedence and the guard stands down. Self-limiting: with the escape the
+    // global maximum sits exactly at the threshold and none over.
+    {
+        const auto thr = static_cast<size_t>(std::max(0, m_params.split_high_valence_threshold));
+        const auto valence = [&](size_t v) {
+            return get_one_ring_tids_for_vertex(tuple_from_vertex(v)).size();
+        };
+        bool relieve = thr > 0 && (valence(v1_id) > thr || valence(v2_id) > thr);
+        if (thr > 0 && !relieve) {
+            for (const size_t nb : get_one_ring_vids_for_vertex(v1_id)) {
+                if (valence(nb) > thr) {
+                    relieve = true;
+                    break;
+                }
+            }
+        }
+        if (!relieve) {
+            const Vector3d p1 = m_vertex_attribute[v1_id].m_posf;
+            const Vector3d p2 = m_vertex_attribute[v2_id].m_posf;
+            for (const size_t nb : get_one_ring_vids_for_vertex(v1_id)) {
+                if (nb == v1_id || nb == v2_id) continue;
+                const double after2 = (p2 - m_vertex_attribute[nb].m_posf).squaredNorm();
+                const double sbar = 0.5 * (m_vertex_attribute[v2_id].m_sizing_scalar +
+                                           m_vertex_attribute[nb].m_sizing_scalar);
+                if (after2 <= m_params.splitting_l2 * sbar * sbar) continue; // not over-long
+                const double before2 = (p1 - m_vertex_attribute[nb].m_posf).squaredNorm();
+                if (after2 > before2) return false; // creates or lengthens an over-long edge
+            }
+        }
+    }
+
     // The base only knows that both endpoints are on SOME tracked surface. A vertex may not
     // leave the particular surface it belongs to: an input-complex vertex carries input
     // geometry, and an offset-boundary vertex carries the offset.
--- a/components/topological_offset/wmtk/components/topological_offset/Collapse.cpp
+++ b/components/topological_offset/wmtk/components/topological_offset/Collapse.cpp
@@ -11,28 +11,36 @@ namespace wmtk::components::topological_offset {
 
 double TopoOffsetTetMesh::face_normal_deviation(const Tuple& f) const
 {
-    const std::array<Tuple, 3> fv = get_face_vertices(f);
-    const Vector3d p_a = m_vertex_attribute[fv[0].vid(*this)].m_posf;
-    const Vector3d p_b = m_vertex_attribute[fv[1].vid(*this)].m_posf;
-    const Vector3d p_c = m_vertex_attribute[fv[2].vid(*this)].m_posf;
+    // Paper Definition 5: the maximum angle between the OFFSET normal at the element's center
+    // and the OFFSET normal at other points within the element.
+    //
+    //     sigma(t) = max over p_i in t of angle( n(p_c), n(p_i) ),  p_i = (1-u)*p_v + u*p_c
+    //
+    // Both terms are FIELD normals; the face's own geometric normal does not appear. That is the
+    // whole point and it is what makes the quantity converge: n() is continuous away from the
+    // input complex's features, so shrinking a face brings its samples closer together and drives
+    // sigma to zero, which is something the sizing field can actually satisfy.
+    //
+    // This used to compare the face's own normal against the sample normals, i.e. misorientation.
+    // Refinement does not fix misorientation -- a smaller triangle in the same plane is just as
+    // misoriented -- so the sizing field could never drive it down, and update_sizing_field()
+    // would refine around such a face forever. The 2D side hit exactly that and was corrected to
+    // Definition 5; see the normal-deviation section of .claude/CLAUDE.md.
+    //
+    // offset_surface_samples() already produces the paper's sample set: index 0 is the centroid
+    // p_c, indices 1-3 are (1-u)*p_v + u*p_c with u = 0.1, and each carries the field normal.
+    const std::array<OffsetSurfaceSample, 4> samples = offset_surface_samples(f);
+    const Vector3d& n_c = samples[0].normal;
+    if (n_c.squaredNorm() < 1e-20) return 0.; // centroid sits on the complex: no direction
 
-    const Vector3d ab = p_b - p_a;
-    const Vector3d ac = p_c - p_a;
-    const Vector3d cross = ab.cross(ac);
-    const double cross_norm = cross.norm();
-    if (cross_norm < 1e-12) return 0.; // degenerate triangle, nothing to measure
-    const Vector3d face_normal = cross / cross_norm;
-
-    // max over all 4 samples, not just the centroid: a face straddling a feature has samples
-    // on both sides of it, and only the max catches the one that disagrees with the face's own
-    // (necessarily flat) normal -- averaging or using a single sample would miss it.
     double max_dev = 0.;
-    for (const OffsetSurfaceSample& s : offset_surface_samples(f)) {
-        if (s.normal.squaredNorm() < 1e-20) continue; // degenerate: no normal direction
-        // orientation independent: a triangle whose plane is parallel to the sample's implied
-        // plane counts as aligned regardless of which way get_face_vertices() winds it
-        const double c = std::clamp(face_normal.cross(s.normal).norm(), -1., 1.);
-        max_dev = std::max(max_dev, (180. / M_PI) * std::asin(c));
+    for (int i = 1; i < 4; ++i) {
+        const Vector3d& n_i = samples[i].normal;
+        if (n_i.squaredNorm() < 1e-20) continue; // degenerate sample: no normal direction
+        // A genuine angle between two field normals, both pointing outward from the complex, so
+        // there is no orientation ambiguity to fold away here.
+        const double c = std::clamp(n_c.dot(n_i), -1., 1.);
+        max_dev = std::max(max_dev, (180. / M_PI) * std::acos(c));
     }
     return max_dev;
 }
@@ -92,6 +100,16 @@ bool TopoOffsetTetMesh::collapse_before_vertex(
 {
     const auto& VE = m_vertex_extra;
 
+    // v1 is the vertex the collapse REMOVES (it merges into v2, which keeps its position). If
+    // v1 belongs to the input complex or the domain boundary, removing it deletes a simplex of
+    // a set that must not change at all, so the collapse is refused outright -- including the
+    // input-onto-input case the surface rule below would otherwise allow. That case is not
+    // hypothetical: it is what lets the shared engine decimate the input surface down to
+    // whatever m_envelope tolerates.
+    if (vertex_is_frozen(v1_id)) {
+        return false;
+    }
+
     // The base only knows that both endpoints are on SOME tracked surface. A vertex may not
     // leave the particular surface it belongs to: an input-complex vertex carries input
     // geometry, and an offset-boundary vertex carries the offset.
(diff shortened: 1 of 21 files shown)


udaykusupati:
Spencer's 3D offset optimization (`spenbert/offsets-3d`), rebased onto `main` at `0086e56a52` with authorship preserved, plus the work that takes it from "avg error rises every iteration" to a fast engine-driven loop with one measured blocker left. Every number below is from `127891` (prism, δ=0.837, convergence target 5% of δ) or the four `topological_offset_*` integration tests. 2D is bit-identical throughout.

## Why it didn't converge — the causal chain

1. **The offset was invisible to the shared engine.** `m_vertex_attribute[].m_is_on_surface` was never set in 3D (it is a different field from `m_vertex_extra[].m_is_on_offset/_on_input`, which only say *which* surface). `is_edge_on_surface()` short-circuits on it, so the collapse's surface link condition, `preserve_topology`, and the split's flag propagation were dead code for the offset. → fixed in `f8fc89b34c`.
2. **Most splits never happened, silently.** When the preallocated slot pool (6× live count at last consolidate) runs out, `TetMesh::split_edge` aborts *before any application hook* — no rejection recorded, nothing logged. The old single-split-pass-per-iteration loop lost ~31k splits/iteration; 19 of ~723 offset edges the sizing field asked to split were split, and the offset **coarsened** 1172→490 faces over a run. → counted and warned in `ebb8a6e8f3`/`6c1b6c2e1f` (TetMesh + TriMesh).
3. **With splits landing, the collapse undid them.** The softened normal-deviation bar `max(σ_max, nd_before)` takes the max over the patch, so one 35° face at a crease licenses coarsening of its whole neighbourhood; the guard fired 36–270×/iter against ~2,800 accepted offset-vertex removals, destroying 75% of offset faces per pass. The paper's rule is flat, and the reference implements it flat (`NormalDeviationAfterInvariant` with `compare_with_before=false`, `OffsetOptimization.cpp:1135`). → flat bar in `ce2de20696`; guard now fires 3,800–22,600×/iter, collapse keeps 52% of offset faces instead of 25%. 2D keeps worse-of (it converges with it; its recorded 429→80 counterexample was against the *ratchet* rule, not flat).
4. **Per-element sizing halving blew up the runtime.** Halving every failing vertex every iteration ratchets crease bands to the floor unconditionally (σ at a crease exceeds σ_max at *any* resolution), and gradation drags the fine sizing into the volume at ×8 tets per halving: 2.8M edges by iteration 4. → `9aa78ded29`: the optimization now runs on **`mesh_improvement()` itself**, simwild-style — `optimization_quality_stats()` = (max, avg) band-vertex distance error / `convergence_target`, `optimization_stop_metric()` = 1.0, `refine_sizing_around_worst()` = the engine's stall-driven ratchet driven by distance error instead of AMIPS (worst `stuck_refine_num_worst` band vertices, `stuck_refine_rings`, ×0.5 to the `l_min` floor, graded). 7 iterations now take 20 s.

Two recorded decisions in `9aa78ded29`, both to revisit: the engine loop is driven by **distance only** (the criterion's other half is an *average* normal deviation, which cannot be a max-based stop; it is tested after the loop, so a run can converge on distance and still be reported unconverged), and the paper's 1.5× sizing growth is kept, now firing on stall rather than every iteration.

Also on the branch: `6a7d072d48` + revert `5f15735843` — a sizing-seed experiment that measured better but silently voided `length_rel` and bypassed the engine's sizing strategy; the revert message preserves its measurements. `0aa92d5c1a`'s split-retry loop is superseded by the engine loop's consolidate-per-iteration; its offset-restricted diagnostics (`[offset …]` band histograms, valence, offset-vertex-removal counter) remain and are what most of the numbers here come from.

## The remaining blocker, measured

A fixed point at **max dist err 0.162 vs target 0.042**, bit-stable over 30 iterations. Mechanism: the split queue is longest-first in *absolute* length, so the offset's edges — the shortest, 96% above their own sizing gate — are exactly the tail the slot pool's exhaustion cuts off every pass, while the halo around the offset churns (split children land below 4/5 of target → freed by the collapse → merged survivors land above 4/3 → resplit). The pool budget is the one lever that measurably moves it:

| `preallocation_factor` | max dist err | note |
| --- | --- | --- |
| 6 (default, shipped) | 0.162 | fixed point |
| 40 | 0.140 | 3 min / 7 iters |
| 100 | 0.109 | iteration 13, still descending; cost ≈ linear |

2D is the precedent: its loop never consolidates, so it runs on the high-water pool from construction — effectively unbounded, which is why its splits all land (833→16,261 vertices in one pass, 19×) and why 2D converges.

## Pickup order

1. **Split priority under a non-uniform sizing field.** Longest-*relative-to-target* instead of longest-absolute would serve the offset first. That is `executor.priority` in `TetOptimizerMeshSplit.cpp` — engine semantics, not a virtual; needs its own discussion (uniform-field apps are order-unchanged).
2. **Pool budget policy.** `set_preallocation_factor()` is the engine's per-application knob; the branch ships the default. See the table above for what a value buys.
3. **Construction defect, pre-existing on main: both 3D integration tests can never converge.** Their pinned worst vertex is `on_offset && on_input` — the constructed offset *touches the input complex*; such a vertex is frozen by design (smoothing refuses on-input, collapse may not remove it), so its error is ≈ δ forever. Fingerprint: `max_dist_err` bit-identical through every commit here (0.08721737…, 0.15926708…). The paper's offset never touches S — this is a marching/growth-phase fix. Prism does not have it.
4. **Distance metric.** `compute_distance_deviation` samples at vertices; the reference scores per face over four samples (`sq_dist_max`, `OffsetOptimization.cpp:585`). Vertex sampling is satisfied by construction by any projection-based placement, so port this before tuning the smoother — the `#if 0` candidate-selection block in `Smooth.cpp` stays parked until then.

## Hypotheses tested and refuted (don't re-walk these)

- 4/5–4/3 gate overlap as a shared split/collapse defect — tetwild reaches a healthy fixed point from the same ~60%-collapsible post-split state (66% in band after collapse vs the offset's 26%); the shared gates are fine.
- Gradation-boundary leak in the collapse length gate (mean-of-endpoints sizing across a sizing step) — counted: 399 of 200,245 accepted collapses.
- High-valence gate blocking band refinement — band/global valence stay ≤ 50 at rest, none over threshold; the 20–35k per-pass refusals are mid-pass spikes from localized resplitting (a throttle, not the blocker).
- Prism needing distance adaptation (paper §5.3.1) — artefact of the starved runs; once refinement lands its worst vertex moves every iteration and is slightly too *far*, not pinned too close.

## Reproducing

Prism: `wmtk_app -j <config>` with `application: topological_offset`, `input: 127891.msh`, `target_distance_rel: 0.01`, `convergence_target_rel: 0.05`. Watch for `[offset after split/collapse]` (band histogram, valence), `[offset] collapses removing an offset vertex`, `[slots]` warnings, `[stuck-refine]` lines. Integration: the four `topological_offset_*` configs under `data/integration_tests/`.

udaykusupati:
Added components/topological_offset/wmtk/components/topological_offset/scripts/visualize_offset.py — a polyscope viewer for a result (.msh + optionally the config, or a run directory). Layers: input surface, the band's outer offset surface, inner interface, tag-region boundaries, and (2D) the envelope curves. In 3D the offset surface is coloured by |dist − δ|/δ (reds, range shown in the panel; switchable to raw distance or solid blue). The reference is the band's inner interface with exact point-to-triangle distances, cross-checked against the C++ logs: exact agreement on prism (avg 0.048, max 0.194), max exact / avg within 3% on double_sphere. Curve/vertex input complexes have no region to measure against, so the error layer is omitted with a message rather than faked. Deps: polyscope, meshio, numpy (the simwild conda env has all three).


# run either config with wmtk_app, then point the viewer at the output:
python .../scripts/visualize_offset.py <output>.msh data/integration_tests/topological_offset_3d.json
python .../scripts/visualize_offset.py <output>.msh data/integration_tests/topological_offset_2d.json
On topological_offset_3d the construction defect from this PR's description is directly visible: the dark-red hotspot (err → 0.87 δ) is where the offset surface is fused to the input complex — toggle the inner interface on to see them meet.

udaykusupati:
# 3D offset optimization: current design, issues found, fixes, and what carries over

Status: **the prism test case converges** — the first configuration to do so. Worst surface
distance 0.0404 against an allowance of 0.0418 at iteration 10 of 15, stable through 15,
in 7 minutes. Every configuration missing any one of the three main fixes stalls at 2.4–4×
the allowance indefinitely (each measured by ablation). The two 3D integration tests still
fail for a separate, pre-existing reason (last section).

## How the optimization works now

The offset is a shell of tetrahedra wrapped around the input; its outer surface must sit at
distance δ from the input (δ = `target_distance`, or `target_distance_rel` × bounding-box
diagonal). The optimizer runs on the toolkit's own improvement loop (`mesh_improvement`),
plugged in the way simwild is, through three overrides:

- **quality metric** = distance error of the shell's outer-surface vertices, normalized so
  1.0 means "at the allowance" (`optimization_quality_stats`);
- **stop metric** = 1.0 (`optimization_stop_metric`);
- **refinement on stall** = when the worst error stops improving, halve the target edge
  length around the 100 worst surface vertices and their neighbours only — the engine's
  stall-refinement pattern with distance error in place of AMIPS
  (`refine_sizing_around_worst`, built from the shared sizing utilities).

Everything else is stock engine: the split/collapse/swap/smooth passes, their scheduling
and gates, per-iteration storage recycling, the finishing collapse pass. Offset-specific
code is confined to (a) construction (marching), (b) placing surface vertices at distance δ
during smoothing, and (c) three documented behaviour divergences implemented through
existing override points — listed below with their reasons.

**Convergence criteria** (both required): the *worst* surface vertex within
`convergence_target` of δ (worst, because distance error can genuinely reach zero
everywhere), and the *average* deviation between the surface's facing and the ideal offset
direction under `convergence_normal_deviation` (average, because sharp input edges impose a
deviation floor no refinement can remove).

## Issues found and fixed

**1. The surface was invisible to the engine** (`f8fc89b34c`). The per-vertex
"tracked surface" flag was never set in 3D, so the engine's surface protections — the
collapse link condition, topology preservation, flag propagation on split — were dead code
for the offset. *Offset-only bug; the other tools set the flag.*

**2. Splits were silently dropped** (`ebb8a6e8f3`, `6c1b6c2e1f`, then `9aa78ded29`). New
vertices come from a storage pool sized at the last compaction; when it ran dry mid-pass,
remaining operations were dropped with no log — ~31k per iteration here, while the surface
*coarsened* 1172 → 490 faces. Now counted and warned (2D and 3D), and the engine loop's
per-iteration compaction retries the work. *Engine-wide and latent: tetwild/triwild only
defer the loss (their loop retries next iteration); any single-pass caller loses work
outright. The 2D offset measurably drops ~22k operations per pass today and converges
anyway.*

**3. Collapse undid every split** (`ce2de20696`, `c82383e3d7`). Halving an edge just over
the split threshold (4/3 of target) leaves children below the collapse threshold (4/5) —
arithmetic, not chance. The scheme these constants come from (Botsch–Kobbelt) is stable
only because its collapse also refuses to *create* an over-long edge; that rule was
missing, so collapse-to-convergence unraveled each freshly refined region. Two fixes: the
paper's flat normal-deviation limit on collapse (the reference implements it flat;
our softened variant let one bad face license coarsening of its whole neighbourhood), and
the don't-create-long-edges rule — with one exception: when a vertex has accumulated far
too many incident tetrahedra, the collapse relieving it proceeds anyway, else the mesh
jams (measured: valence 240 without the exception, capped exactly at the 200 threshold
with it). *Conditional for others: it needs deep refinement against a persistently
non-uniform target. Tetwild never does that (uniform target, one split generation,
measured healthy). Simwild with a user sizing field is the plausible other victim —
unmeasured.*

**4. Collapse discarded the target edge length** (`39984f88b7`). On merge the survivor
kept its own target and the removed vertex's finer one was thrown away — the survivor's
other attributes are merged in the same block (OR/max), sizing simply wasn't. So the very
operation that runs after refinement was forgetting it. Diagnosis and fix shape are
Daniele's: a default-inert engine hook (default = today's behaviour, other tools
bit-identical — verified by measurement, not just construction) overridden to min in the
offset. *Engine-wide fact. Consequence-free for default tetwild (stall machinery off by
default), but simwild feeds a user-provided sizing field into the same per-vertex value —
a collapse there discards user intent. Unmeasured; the hook is the ready-made opt-in.*

**5. The surface's fine sizing leaked into the volume** (`22273f67d7` — the fix that turns
"improves but stalls" into "converges"). The reference keeps its length-driven refinement
on the offset surface mesh and gives the tet embedding a separate, quality-driven target;
our port shared one per-vertex field and graded the surface's fine values into the
surrounding volume. That manufactured enormous volume refinement demand which the collapse
then deleted — a build-and-demolish cycle of ~4k vertices per iteration that exhausted the
storage pool and starved the surface's own (shortest, last-in-queue) edges. Fine targets
now stay on the surface; the background keeps the resolution `length_rel` asks for.
*Offset-only: for the volume-meshing tools, refining the volume is the goal, not a leak.*

Ablations: without fix 3, stall at 0.162 regardless of everything else; without fix 5,
plateau at ~2.4× the allowance; fix 4 alone is inert (until refinement survives, merges
only ever see equal targets) and contributes in composition.

## Deliberate divergences from the engine, all component-side

- Flat normal-deviation limit on collapse (paper's rule, reference's implementation; 2D
  keeps the softer variant it converges with).
- Don't-create-long-edges collapse rule + valence exception (Botsch–Kobbelt's missing leg).
- The paper's 1.5× target growth where the surface is flat and well-shaped (the engine's
  own refinement only ever shrinks; recorded for revisit).
- Distance drives the loop; the average-facing criterion is checked after it (an average
  cannot be a worst-based stop). A run can pass distance and still be reported unconverged.

## Still open

- **Construction defect, pre-existing, blocks both 3D integration tests:** their offsets
  are built already *touching* the input in places; such vertices are frozen on both
  surfaces, so their error is ≈ δ forever and no optimization change can converge these
  tests (their worst-error values have been bit-identical through every change here:
  0.0872…, 0.1593…). Needs a fix in the marching/growth phase.
- The distance metric samples at surface vertices; the reference scores per face over four
  samples. Worth porting before any smoother tuning.
- For Daniele: whether tetwild should measure the deferral cost of issue 2, and whether
  simwild wants the issue-4 hook for its user sizing field.
- 2D has the same missing pieces (issues 3–5) and converges today regardless; left
  untouched, bit-identical throughout.

Tools: `scripts/visualize_offset.py` (same directory as the component) renders any result
with the surface coloured by distance error; its numbers are cross-validated against the
optimizer's own logs.

PR #1041, opened 2026-08-14
Tracking: use "quality" consistently across triwild/tetwild/simwild

udaykusupati:
Empty on purpose. This branch exists to host the discussion; the PR body carries the report. Nothing is renamed here yet -- the scope needs agreeing first, and two of the five items below are behaviour changes rather than renames.

Scope is four identifiers -- stop_energy, max_energy, avg_energy, filter_energy -- and explicitly NOT the solver-objective energies (slim_energy, total_energy, compute_energy, energy_from_point, envelope_energy, add_energy, new_energy), which are correctly named and would be made worse by the rename.

Prompted by the coarsening fix in #1035 (735f2b040a), which found the shared optimizer comparing raw quality across cells whose targets differ. See the PR body for the five follow-ups and the measurements behind each.


udaykusupati:
Note to self: Add a python script to port old jsons to new

PR #1045, opened 2026-08-17
Test Linux on arm64, and stop shipping the build tree to Docker

danielepanozzo:
Two independent changes, both found while reproducing a tetwild test that fails **only** on Linux/arm64.

> [!IMPORTANT]
> **The new arm64 job will fail on this PR until [VolumeRemesher#25](https://github.com/wildmeshing/VolumeRemesher/pull/25) lands** and the pin in `cmake/recipes/volumeremesher.cmake` is bumped. That failure is the bug this job exists to catch. I'll push the pin bump here once the upstream PR merges.

## 1. The CI gap

| | x86-64 | arm64 |
|---|---|---|
| **Linux** | ✅ | ❌ **no job** |
| macOS | — | ✅ |
| Windows | ✅ | — |

No job ever compiled Linux on arm64 — and that is precisely where the failure lived.

`VolumeRemesher` cached `orient3D` results (`-1`, `0`, `+1`) in a `std::vector<char>` and read them back comparing against `-1`. **`char` is signed on x86-64 and under Apple's arm64 ABI, but unsigned under AAPCS64.** On Linux/arm64 a stored `-1` reads back as `255`, the comparison never matches, every constraint is judged not to split its cell, and the input surface is silently never embedded.

Symptom: `wmtk_test_tetwild` / `vertex_order` (shark-fin) finds **0 order-3 vertices where it expects 2**. 1 of 117 tests; everything else passes.

Reproduced natively on aarch64 Ubuntu 24.04 with GCC 13:

| | initial cells | final cells | tets |
|---|---:|---:|---:|
| macOS/arm64 | 12162 | 13884 | 20753 |
| Linux/arm64 before | 12162 | **12162** | 12162 |
| Linux/arm64 after | 12162 | 13884 | 20753 |

With the upstream fix: **117/117 wmtk tests pass** on linux/arm64, and VolumeRemesher's own suite goes from 22/28 to 28/28.

This adds `ubuntu-24.04-arm` in Debug and Release. These runners are free for public repos.

## 2. Docker build context

There was no `.dockerignore`, so `COPY . .` shipped the entire working tree — **5.2 GB**, of which 4.5 GB is local `build*` directories the image rebuilds from scratch anyway and 643 MB is `.git`. Now ~60 MB.

Dropping `.git` is safe: there is no `.gitmodules`, so the Dockerfile's `git submodule update --init --recursive` is a no-op on this repository.

## Note on the Dockerfile

The committed `Dockerfile` is fine — it uses `Release` and builds clean on arm64 (verified end to end). A modified copy circulating with `-DCMAKE_BUILD_TYPE=Developer` and appended `make test` steps is **not** what's in this repo. Worth knowing that `Developer` is not a build type this project defines: CMake accepts it silently and drops `-DNDEBUG`, compiling in all 413 assertions. That is not the cause of this bug, and needs no change here.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,33 @@
+# Without this, `COPY . .` ships the whole working tree to the daemon -- 5.2 GB here,
+# almost all of it local build directories that the image rebuilds from scratch anyway.
+
+# Local build trees (host-specific absolute paths in their CMake caches; useless in the image)
+build*/
+cmake-build-*/
+out/
+
+# Version control. There are no submodules, so the Dockerfile's
+# `git submodule update --init --recursive` step is a no-op and .git is dead weight (643 MB).
+.git/
+.gitignore
+.github/
+
+# Editor / agent scratch
+.claude/
+.vscode/
+.idea/
+.DS_Store
+
+# Test and run output that accumulates in the working tree
+Testing/
+*.log
+*.obj
+*.msh
+*.vtu
+*.stl
+
+# Python
+__pycache__/
+*.pyc
+.venv/
+venv/
--- a/.github/workflows/continuous.yml
+++ b/.github/workflows/continuous.yml
@@ -29,6 +29,19 @@ jobs:
             config: Release
             integration_tests_flag: &integration_tests_on "-DWMTK_BUILD_INTEGRATION_TESTS=ON"
             name: Linux
+          # arm64 runners. Until these existed the matrix covered Linux only on x86-64 and
+          # arm64 only on macOS, so no job compiled the combination -- which is exactly where
+          # a `char` signedness bug in VolumeRemesher hid: `char` is signed on x86-64 and on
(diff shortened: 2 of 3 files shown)


danielepanozzo:
[VolumeRemesher#25](https://github.com/wildmeshing/VolumeRemesher/pull/25) is merged, and the pin is bumped to `609e32c4` in `cmake/recipes/volumeremesher.cmake`.

That PR went green on all 8 jobs, **including the two `Linux arm64` jobs it added** — the platform that had never compiled this code. Its own integration suite goes 22/28 → 28/28 there.

So the caveat at the top of this PR no longer applies: the `Linux arm64 (Debug)` and `Linux arm64 (Release)` jobs here should now pass. They were failing on `wmtk_test_tetwild` / `vertex_order`, which is exactly the regression they were added to catch, and the bump is what fixes it.

PR #1053, opened 2026-08-20
Collapse limit at admission, stuck_refine_cooldown = 1, and no quality veto on splits

danielepanozzo:
Three changes to the collapse/refinement loop, each restoring behaviour that a guard or a gate had taken away. All measurements are over the 31 challenging models at `stop_energy 10`, `num_threads 1` (deterministic — no threading noise), 12 in parallel on 16 cores, `46024` separately at 16 threads.

## 1. The collapse length limit applies to the candidate list, not at acceptance

Reverts the placement introduced in 73ddbbae (#997). An edge at or above the collapse target is filtered out by `is_weight_up_to_date` and never offered to `collapse_edge_before`; the acceptance-time block — which admitted an over-length edge whenever it strictly improved the ring's worst element — is removed.

This is what both ancestors do: fTetWild filters at queue admission and re-checks on pop (`EdgeCollapsing.cpp:47`, `:88`), TetWild 2018 the same via `isCollapsable_cd3`. Constants and sizing scaling already matched theirs and are unchanged (`4/5` collapse, `4/3` split, squared, scaled by the endpoints' mean `m_sizing_scalar`). The pre-existing `collapse_quality_allowed` per-cell gate predates #997 (cb9b81a6) and is untouched.

| | main | this change | |
|---|---|---|---|
| iterations | 356 | 351 | −1.4% |
| elements | 2 151 046 | 2 192 768 | +1.9% |
| wall (sum) | 114.3 min | 104.5 min | **−8.5%** |
| converged | 31/31 | 31/31 | — |

Per-model differences are noise and go both ways. The wall-time gain is the expected one: a filtered candidate never reaches `collapse_edge_before`'s ring work. By application, **−10.2% wall on tetwild**, −6.1% on triwild.

## 2. `stuck_refine_cooldown` defaults to 1

Refining the sizing field every iteration can outrun the operations. On triwild20k `189017` at `eps_rel 1e-4` — the case change 1 regresses — the field ratchets down faster than split and collapse can act on it: the mesh holds at ~31k faces for a dozen iterations, then runs away `54k → 96k → 159k → 5.8M` and never converges, max energy pinned at the inverted sentinel `1e+50`. At cooldown 1 the same run climbs deliberately to 181k and **converges at 9.9994 in 35 iterations**.

Cooldown is the narrow instrument for this: it slows the refinement, where the acceptance-time gate instead licensed collapse to undo refinement anywhere the mesh already held a bad element.

| | iterations | elements | wall |
|---|---|---|---|
| tetwild (15) | 185 → 196 (+5.9%) | −1.8% | −5.3% |
| triwild (16) | 166 → 173 (+4.2%) | −0.1% | −6.2% |

All 31 converge at every value swept. **1 is the knee**: 2 and 3 buy a further 2–4% element reduction in tetwild for +34% and +76% iterations. Default changed in `OptimizerParameters` and in the tetwild, triwild and simwild specs so the three agree.

This supersedes an earlier reading of a 468-model triwild20k sweep (recorded in the comment change 2 replaces), where cooldown 1 cost ~13% wall for identical mesh sizes and was taken as evidence that a cooldown only delays the next escape. That holds where the escape hatch is not load-bearing; it does not hold on the models where it is.

## 3. A split is never refused on quality

Removes the guard added in a4ebcff2 (#997):

```cpp
if (max_quality_after >= MAX_ENERGY && cache.max_quality_before < MAX_ENERGY) return false;
```

That commit introduced it as insurance and said so — *"Neither guard is needed for convergence once the envelope is right, and neither changes any integration-test result."* The first half holds; the second is too generous, because where it does change a result the guard is what makes it worse.

| | guard | no guard | |
|---|---|---|---|
| identical (iterations **and** elements) | — | **29 of 31** | |
| iterations | 392 | 380 | −3.1% |
| elements | 2 274 988 | 2 269 955 | −0.22% |
| converged | 31/31 | 31/31 | — |

The two that move:

```
triwild 189017 @ 1e-4    35 -> 20 iterations,  180941 -> 173846 elements
tetwild 104187           11 -> 14 iterations,  189033 -> 191095 elements
```

`189017` at 1e-4 is the hardest case in the set and converges in 20 iterations instead of 35 for a 3.9% smaller mesh. Mechanism is the obvious one: refusing to split a stuck sliver leaves the sliver, and the run grinds around it.

**It was not preventing degeneracy either.** Ten of the triwild models report the `MAX_ENERGY` sentinel *with* the guard in place, and the per-model sentinel counts are identical without it — removing it created no additional degenerate element anywhere. `101954`, the model it was written against, is unchanged at 10 iterations / 28 925 tets and never reports the sentinel in either build; the failure it targeted does not reproduce, exactly as a4ebcff2 predicted (the envelope fix in that same commit is what fixed it).

Also removes the state that existed only to feed it: the `max_quality_before` scan at the top of `split_edge_before` and the field in both `SplitCache` structs.

## What was ruled out

- **Thresholds.** `sqrt(2)·l` / `l/sqrt(2)` in place of `4/3` / `4/5`, at both collapse placements: aggregate wash (tetwild elements −2.6%, triwild +4.7%), and `189017@1e-4` still diverges — the wider band only delays it ten iterations.
- **Force-split.** `stuck_refine_force_split: false` delays the same divergence ~4 iterations, then reproduces it. Not the mechanism.

## Testing

- 31 challenging models at `eps_rel 1e-3` for every combination above, plus the cooldown sweep 0/1/2/3.
- `189017` at `eps_rel 1e-4`, single-threaded, throughout.
- Golden hashes will move: this changes which collapses are performed, when refinement fires, and which splits are accepted, in tetwild, triwild and simwild.

### DIFF
--- a/components/simwild/wmtk/components/simwild/Parameters.h
+++ b/components/simwild/wmtk/components/simwild/Parameters.h
@@ -43,19 +43,6 @@ struct Parameters : public wmtk::OptimizerParameters
     VectorXd box_max;
     bool stop_at_float = false;
 
-    /**
-     * Verify at init that every surface edge starts inside the envelope (2D remeshing only).
-     *
-     * The same check triwild carries, and the same trade: the invariant is real, but it
-     * costs one sampled segment query per surface edge, serially, on every run. Measured in
-     * triwild's 2D sweep at 22s on a 3.5M-edge input and 5m07s on a 7.6M-edge one -- 8.5% of
-     * that model's entire budget, spent before the first iteration. Off by default.
-     *
-     * The 3D counterpart in VolumemesherInsertion is deliberately NOT gated by this: it is
-     * not a check but a decision, feeding the "rebuild the envelope from tet tags" branch.
-     */
-    bool check_envelope_at_init = false;
-
     std::string operation = "remeshing";
 
     /**
@@ -115,7 +102,6 @@ struct Parameters : public wmtk::OptimizerParameters
 
         debug_output = json_params["DEBUG_output"];
         perform_sanity_checks = json_params["DEBUG_sanity_checks"];
-        check_envelope_at_init = json_params["DEBUG_envelope_sanity_check"];
 
         verbose_quality_stats = json_params["verbose_quality_stats"];
 
--- a/components/simwild/wmtk/components/simwild/SimWildMeshTri.cpp
+++ b/components/simwild/wmtk/components/simwild/SimWildMeshTri.cpp
@@ -226,7 +226,7 @@ void SimWildMeshTri::init_surfaces_and_boundaries()
         tempE.emplace_back(v1, v2);
     }
 
-    if (!m_envelope) {
+    auto env_from_face_tags = [&]() {
         logger().info("Init envelope from face tags");
         // build envelopes
         std::vector<Vector2d> tempV(vert_capacity());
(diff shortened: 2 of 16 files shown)


daniel-zint (approved):
lgtm

PR #1055, opened 2026-08-20
3D offset, A/B optimization

Spenbert02:


cmake/recipes/geogram.cmake
```diff
@@ -0,0 +1,55 @@
...
+# default (DistanceTypeConfig::use_standard_ is false). high_order_contact -- the subtree the
+# offset potential actually evaluates -- calls distance_type from its collision builder and its
+# pair-distance templates, so that fallback would land squarely on our own code, in exactly the
+# near-degenerate configurations an offset surface produces. A version bump changes nothing;
+# dropping the predicates changes what the potential is computed from.
+#
+# Options are ipc's own, verbatim, so this differs from what ipc would have configured in the
+# version and in nothing else.
+
+if(TARGET geogram::geogram)
```
daniel-zint, line 34: Do we really need geogram here? We just got rid of it.

cmake/recipes/ipc_toolkit.cmake
```diff
@@ -0,0 +1,84 @@
...
+    return()
+endif()
+
+message(STATUS "Third-party: creating target 'ipc::toolkit'")
+
+# wmtk asks for C++17 per target (target_compile_features), leaving CMAKE_CXX_STANDARD unset.
+# Abseil, which ipc-toolkit pulls in for its hash functions, probes the compiler's DEFAULT
+# standard at configure time and hard-errors when that is below 17 -- so a per-target request is
+# invisible to it. Set the directory-scoped default for the subtree ipc-toolkit is configured in.
+set(CMAKE_CXX_STANDARD 17)
```
daniel-zint, line 22: It's super scary that there is a CMake include that enforces the C++ standard. We should move that to the main CMake file.
Or would it be an option to deactivate Abseil in the ipc-toolkit?

cmake/recipes/ipc_toolkit.cmake
```diff
@@ -0,0 +1,84 @@
...
+        "cmake/recipes/eigen.cmake. Without it ipc-toolkit's PUBLIC EIGEN_DONT_VECTORIZE gives "
+        "the targets that link ipc a different alignment for Vector2d/Vector4d/Vector3r than "
+        "the rest of the build, and Eigen's unaligned-array assert aborts "
+        "wmtk_test_manifold_extraction and wmtk_test_topological_offset.")
+endif()
+
+# ipc-toolkit links its hash-map backends PRIVATE, but leaks them through the PUBLIC header
+# ipc/utils/unordered_map_and_set.hpp, so any consumer including it fails to find
+# <tsl/robin_map.h> / <absl/hash/hash.h>. Upstream's own test target works around this by
+# linking them again; do it once here instead, so every wmtk consumer inherits the fix.
```
daniel-zint, line 79: That sounds like a fix that should be done on the ipc-toolkit side.

cmake/wmtk_data.cmake
```diff
@@ -16,13 +16,31 @@ ExternalProject_Add(
...
+    #
+    # The last two names added are FIXTURES FOR AN OPEN PROBLEM rather than cases awaiting a fix
+    # elsewhere: topological_offset_2d_two_circles and topological_offset_2d_annots_tag4_in both
+    # parse against the current strict spec and are parked because the case FAILS. They are the
+    # outward and inward forms of the same thing -- two offset fronts approaching one curve, where
+    # the summed potential never falls to the level c, so those vertices chase a level set that
+    # does not exist and the stuck-refine calls they generate run the ambient mesh away to 1e50.
+    # See "OPEN PROBLEMS" in .claude/CLAUDE.md.
+    #
+    GIT_TAG 5ec0988783795586ad58de404d152393bdabc96f
```
daniel-zint: We should just update the main branch of the data repo. I am not a huge fan of having different branches of the data repo. It might make sense in the short term, but it should be fixed before this PR is merged.

src/wmtk/envelope/Envelope.cpp
```diff
@@ -143,7 +143,19 @@ void SampleEnvelope::init(
...
-    exact_envelope.init(V, F, _eps);
+    // THE SAME ZERO-WIDTH TRAP init_exact_edges REFUSES, reached the other way. This overload
+    // builds the exact structure unconditionally, so it cannot lean on use_exact to skip a
+    // meaningless build the way the edge overloads do -- and a caller asking for a triangle
+    // envelope with eps <= 0 wants the BVH, not containment (isotropic_remeshing does exactly
+    // that, with use_exact true and m_has_envelope false). Building it anyway would leave a
+    // zero-width FastEnvelope that answers "outside" for every point not exactly on the input:
+    // not a crash, just an optimization that silently cannot move. So skip the build and let
+    // m_exact_built refuse the query instead.
+    m_exact_built = false;
```
daniel-zint, line 154: This code needs clean-up. The 3D and 2D versions look different, though I think they behave the same way.

Cleaning up the Envelope code should be done in a subsequent PR.

src/wmtk/optimization/SmoothVertex.hpp
```diff
@@ -357,7 +396,7 @@ bool smooth_vertex_3d(
         max_after_quality = std::max(max_after_quality, quality);
     }
 
-    if (!VA[vid].m_is_on_surface || opts.quality_veto_on_surface) {
+    if (opts.quality_veto && (!VA[vid].m_is_on_surface || opts.quality_veto_on_surface)) {
```
daniel-zint, line 399: There is a quality veto and a quality veto on the surface. Are both needed? Do we ever want to accept non-surface vertices that do not improve quality?

src/wmtk/OptimizerParameters.h
```diff
@@ -119,6 +119,14 @@ struct OptimizerParameters
...
     // still far above the position-rounding scale so it stays numerically safe.
     double stuck_refine_min_scalar = 1e-3;
+    // How a split midpoint inherits its sizing scalar. false (TetWild's rule): the MEAN of
+    // the two endpoints, which dilutes refinement across the frontier of a refined region --
+    // each split born on that frontier starts halfway back to coarse. true: the MIN of the
+    // two, so refinement propagates undiluted and a refined region reaches its target in one
+    // generation of splits instead of geometrically. (The collapse survivor keeps its own
+    // scalar in 2D either way; only the split reads this.) Wired in TriOptimizerMesh only --
+    // the 3D split keeps the mean unconditionally.
+    bool sizing_propagate_min = false;
```
daniel-zint: Do we need this? I cannot imagine that it is a good idea to propagate the `min`. If I see it correctly, it is also not used anywhere.

src/wmtk/OptimizerParameters.h
```diff
@@ -175,6 +183,11 @@ struct OptimizerParameters
     int project_line_search_steps = 12;
     /// Partial-projection bisections tried after it gives up; 0 disables that pass.
     int project_line_search_nested_steps = 0;
+
+    /// Whether smoothing refuses a move that raises the worst incident element's quality. True
+    /// is TriWild's and SimWild's behaviour; see optimization::SmoothVertexOptions::quality_veto
+    /// for the objective it stops being the right rule for.
+    bool smooth_quality_veto = true;
```
daniel-zint, line 183: ~Also, for this one, I am unsure if this should ever be `false`.~

Ok, seems necessary to push offset vertices to their positions. That should be documented somewhere.

src/wmtk/TetMesh.h
```diff
@@ -710,6 +710,19 @@ class TetMesh
...
+    /**
+     * @brief How many operations have aborted because the preallocated slot pool ran out.
+     *
+     * The abort happens inside the connectivity update, before any application hook, so nothing
+     * downstream can observe it: the operation simply does not happen, no rejection is recorded,
+     * and nothing is logged. mesh_improvement() is insulated from that because it consolidates
+     * every iteration and the operation is retried, but a caller that runs a single pass silently
+     * gets a fraction of the work it asked for. Reset it before a pass and check it after.
+     */
+    size_t slot_exhausted() const { return m_slot_exhausted.load(std::memory_order_relaxed); }
```
daniel-zint: **This must be removed!!** I added a function `bool slots_exhausted()` and this code here is just horribly confusing. The code was already adjusted to restart an operation if the slots are exhausted so there is no need for these three functions anymore:
`slot_exhausted()`, `reset_slot_exhausted()`, `note_slot_exhausted()`

src/wmtk/TetMesh.h
```diff
@@ -751,6 +764,7 @@ class TetMesh
     SlotPool<TetrahedronConnectivity> m_tet_connectivity;
 
     double m_preallocation_factor = 6.0;
+    std::atomic<size_t> m_slot_exhausted{0};
```
daniel-zint: This variable needs to go as well.

src/wmtk/TetOptimizerMesh.h
```diff
@@ -468,6 +517,22 @@ class TetOptimizerMesh : public wmtk::TetMesh, public wmtk::RationalPositions
...
+     *
+     * The default keeps the survivor's own value -- today's behaviour, byte-identical for
+     * every existing application. The offset overrides it to min: its stall-driven
+     * refinement lowers scalars on band vertices, and a collapse that discards the removed
+     * vertex's finer value onto a coarser survivor un-refines the FIELD by topology, so
+     * each round re-lowers from scratch and the band never actually gets finer. min is
+     * the field's established convention (refinement and gradation smoothing only ever
+     * lower a scalar).
+     */
+    virtual double collapse_merged_sizing(double /*removed*/, double survivor) const
```
daniel-zint: This looks like something I fixed once for SimWild. Collapse should always take the min of the two sizing scalars. I don't think any application wants a different behavior, so there is no need for this function here. It can be just hard-coded in the `collapse_after` function.

src/wmtk/TriMesh.h
```diff
@@ -408,6 +408,17 @@ class TriMesh
...
     size_t request_vert_slots(size_t n);
 
+    /**
+     * @brief How many operations have aborted because the preallocated slot pool ran out.
+     *
+     * Identical in meaning and in mechanism to TetMesh::slot_exhausted(): the abort happens
+     * inside the connectivity update, before any application hook, so the operation simply does
+     * not happen and nothing downstream can observe it. Reset before a pass and check after.
+     */
+    size_t slot_exhausted() const { return m_slot_exhausted.load(std::memory_order_relaxed); }
```
daniel-zint: Same as in `TetMesh`. These need to be removed, along with their variable `m_slot_exhausted`.

src/wmtk/TriOptimizerMesh.cpp
```diff
@@ -40,11 +40,14 @@ void TriOptimizerMesh::mesh_improvement(int max_its)
     partition_mesh_morton();
 
     if (m_params.debug_output) {
+        m_debug_pass_name = "improve-entry";
```
daniel-zint, line 43: I am fine with improving the debug output, but this seems like the wrong way. It is really hacky and hard to follow. I still vote for a standardized output name and a JSON that may contain additional information for the specific output file. Then the Polyscope script could also just read the JSON instead of a folder. I think that makes it cleaner.

src/wmtk/TriOptimizerMesh.cpp
```diff
@@ -53,6 +56,7 @@ void TriOptimizerMesh::mesh_improvement(int max_its)
     for (int it = 0; it < max_its; ++it) {
         m_iterations_used = it + 1;
         logger().info("\n========it {}========", it);
+        optimization_iteration_begin();
```
daniel-zint: This function does nothing anywhere. It should be cleaned up.

src/wmtk/TriOptimizerMesh.cpp
```diff
@@ -107,8 +111,7 @@ void TriOptimizerMesh::mesh_improvement(int max_its)
             --refine_cooldown;
         } else if (
             it > 0 && max_metric > optimization_stop_metric() &&
-            (pre_max_metric - max_metric) <=
-                m_params.stuck_refine_stall_eps * (max_metric - optimization_stop_metric())) {
+            optimization_stalled(pre_max_metric, max_metric)) {
```
daniel-zint, line 113: Similar here. This function has no override. We can keep it, but then the `TetMesh` version should be adapted as well.

src/wmtk/TriOptimizerMesh.cpp
```diff
@@ -165,9 +175,11 @@ std::tuple<double, double> TriOptimizerMesh::local_operations(
                 retry_count);
         }
         timer.start();
+        m_debug_pass_name = ops[i] > 0 ? std::string(names[i]) : std::string(names[i]) + "-skipped";
```
daniel-zint, line 177: As mentioned already, I dislike this.

src/wmtk/TriOptimizerMesh.cpp
```diff
@@ -165,9 +175,11 @@ std::tuple<double, double> TriOptimizerMesh::local_operations(
...
         if (i == 0) {
             for (int n = 0; n < ops[i]; ++n) {
                 logger().info("==splitting {}==", n);
+                ++m_op_epoch; // see m_op_epoch: one epoch per split pass
```
daniel-zint, line 181: Sounds like debug code. Let's clean that up.

daniel-zint (changes requested):
I am not comfortable with merging the code in its current state. Let's discuss these comments as soon as possible. Most should be easy to fix.

cmake/recipes/geogram.cmake
daniel-zint: It really is needed due to the higher-order IPC stuff. The main branch does not rely on geogram, so there is hope that it will be removed again at some point.

cmake/recipes/ipc_toolkit.cmake
daniel-zint: Abseil cannot be deactivated as it is not optional for the higher-order stuff. This is, for now, the cleanest solution.

daniel-zint (approved):
Let's clean up the debug output and instrumentation variables at some point, but I think the code is fine for now.

For the geogram/ipc-toolkit business, we need to wait until that repo is cleaned up, so nothing we can do for now on this side.

PR #1068, opened 2026-09-11
3D offset: front_refuse_converged_collapse, a guard against un-resolv…

Spenbert02:
…ing the placed front

New key (default false: the collapse and swap passes exactly as before). On, the passes refuse an operation that would turn a resolved patch of the front into an unresolved one, which is what the tail churn was made of: a placed front kept being re-coarsened by collapses inside the offset envelope, minting faces that sag 1-2 tubes and reopening the refinable set every turn.

- snapshot_front_convergence() at the start of the collapse and swap groups records every front vertex's Newton-step ratio; a face is resolved when its corners are within the bar there and its centroid sag is within the tube.
- Collapse: for a front-to-front edge whose endpoints have only resolved faces around them, the result ring is predicted before the collapse (survivor keeps its position) and refused if any face would sag over the tube; after the collapse the survivor's ratio is measured on its real ring and the base rolls the collapse back if it is over the bar.
- Surface flip: two resolved offset faces are not flipped into a pair of which one sags over the tube.
- Per-turn and run-total refusal counts are logged only when the key is on.

Tested on the five 3D test models at target_distance_rel 1e-2 with default parameters, guard off / on: sphere 20 / 10 turns, cube 13 / 8, prism 18 / 7, cylinder 22 / 8, triangular prism 15 / 9; run time 2-4x lower, meshes 15-18% smaller, max AMIPS within 0.5 of the plain run, all converged. No swap was ever refused.

testing results: https://claude.ai/code/artifact/251a0acb-dc97-4cc0-969f-c55b96ad7a44

### DIFF
--- a/components/topological_offset/wmtk/components/topological_offset/Optimize3d.cpp
+++ b/components/topological_offset/wmtk/components/topological_offset/Optimize3d.cpp
@@ -270,6 +270,18 @@ bool TopoOffsetTetMesh::swap_before_surface(
     if (face_mask({{a, b, c}}) != face_mask({{a, b, d}})) {
         return false;
     }
+    // front_refuse_converged_collapse: two resolved offset faces may not be flipped into a pair
+    // of which one sags over the tube. No vertex moves in a flip, so the new faces' sag is
+    // known here; the corners' ratios are taken as unchanged.
+    if (m_offset_params.front_refuse_converged_collapse &&
+        m_face_attribute[fid_abc].m_surface_class == OFFSET_SURFACE_CLASS &&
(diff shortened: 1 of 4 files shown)


Spenbert02:
unified placement criteria validated on topo_annots_groups model. description/results: https://claude.ai/code/artifact/e63b0fb8-d8c0-4dd5-b6ab-b661f386a102

Spenbert02:
Unified placement criteria now tested in 3D, it makes no difference. Convergence is still unpredictable with the tail. Validation: https://claude.ai/artifact/DtDSm2y3yXbpP7hPDY9Ym4

